Issue #55, opened 2025-11-06
Monitoring Vpc created even if I specify "use existing"


OleksandrRebenok:
Hi,
```
poetry run ccwb --version
Claude Code With Bedrock (version 1.0.0)
```
I run poetry run ccwb init.
Specified use existing vpc and selected 2 subnets.

I can confirm this by providing source/.ccwb-config/config.json 
```
      "monitoring_enabled": true,
      "monitoring_config": {
        "create_vpc": false,
        "vpc_id": "vpc-0e0e8e1c1b244d83",
        "subnet_ids": [
          "subnet-06aee8ba439b026d",
          "subnet-01d249885558749f"
        ]
      },
```

Now when I run `poetry run ccwb deploy` I see that networking stack is created(and vpc is created). 

Thanks in advance.

schuettc:
Have made some updates to `ccwb init` function in latest.  It should handle VPCs better now if you want to retest.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/releases/tag/v2.0.0

PR #66, opened 2025-11-19
Fix of using existing vpc

OleksandrRebenok:
*Issue #55*

*Description of changes:*
* Add dedicated stack s3bucket.
* Skip networking stack if user choose to use existing VPC.
* In case of using existing VPC, we pass subnets and vpc to monitoring stack from user input.


schuettc:
Thanks for the help!  Merged in.  Appreciate it!

PR #99, opened 2026-01-15
fix: quota stack deployment looks for S3 bucket in wrong stack

### LINKED ISSUE #95: `ccwb deploy quota` fails - looks for S3 bucket in wrong stack
## Description

The `ccwb deploy quota` command fails with the error:

```
Could not get S3 bucket from networking stack claude-code-auth-networking
The networking stack must be deployed first.
Run: ccwb deploy networking
```

This occurs even when the networking stack is fully deployed and complete.

## Root Cause

The quota stack deployment code in `deploy.py` is looking for `CfnArtifactsBucket` in the networking stack, but this output is actually defined in the s3bucket stack (`deployment/infrastructure/s3bucket.yaml`).

## Steps to Reproduce

1. Deploy all stacks using `ccwb deploy all`
2. Observe that quota stack deployment fails
3. Note that networking stack shows as complete in CloudFormation

## Expected Behavior

The quota stack should deploy successfully when the s3bucket stack is deployed.

## Environment

- ccwb version: latest
- OS: macOS

bluedoors:
## Summary

Fixes the quota stack deployment which was failing because it looked for `CfnArtifactsBucket` in the networking stack instead of the s3bucket stack.

## Problem

When running `ccwb deploy quota` (or `ccwb deploy all`), the deployment fails with:

```
Could not get S3 bucket from networking stack claude-code-auth-networking
The networking stack must be deployed first.
```

The networking stack doesn't have a `CfnArtifactsBucket` output - that's in `s3bucket.yaml`.

## Changes

- Updated `deploy.py` to look for `CfnArtifactsBucket` in the s3bucket stack instead of the networking stack
- Updated error messages to reference the correct stack name

## Testing

- Verified quota stack deploys successfully after this fix
- All other stacks continue to deploy correctly

Closes #95

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -711,17 +711,17 @@ class DeployCommand(Command):
                     console.print("Run: [cyan]ccwb deploy dashboard[/cyan]")
                     return 1
 
-                # Get S3 bucket from networking stack for packaging
-                networking_stack = profile.stack_names.get("networking", f"{profile.identity_pool_name}-networking")
-                networking_outputs = get_stack_outputs(networking_stack, profile.aws_region)
+                # Get S3 bucket from s3bucket stack for packaging
+                s3_stack = profile.stack_names.get("s3", f"{profile.identity_pool_name}-s3bucket")
+                s3_outputs = get_stack_outputs(s3_stack, profile.aws_region)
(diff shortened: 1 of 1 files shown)


the-smallest-coder:
tested, fix the problem clearly and totally.

RyanDsilva:
thanks @bluedoors and @the-smallest-coder!

PR #104, opened 2026-01-22
Fix port 8400 conflict by auto-selecting available port

spawnia:
## Summary

- Fixes authentication hang when port 8400 is in use by another application
- Adds `_get_available_port()` method that tries the preferred port first, then auto-selects an available port
- Fixes `quota_port` TypeError when `redirect_port` is `None`
- Documents the `REDIRECT_PORT` environment variable in QUICK_START.md

## Problem

When port 8400 is in use by another service (e.g., [Commvault uses port 8400](https://documentation.commvault.com/v11/software/port_requirements_for_commvault.html)), the credential-process hangs indefinitely. The code assumed that if port 8400 was busy, another credential-process was authenticating, and waited for it to finish. This assumption breaks when external services use port 8400.

## Solution

1. **Dynamic port selection**: Try the preferred port (8400 or `REDIRECT_PORT` env var) first. If unavailable, auto-select a free port using `socket.bind((host, 0))`.

2. **Simplified locking**: Removed the port-based locking mechanism from `run()` and `authenticate_for_monitoring()` since it conflicted with external services.

3. **Quota port fix**: Changed `quota_port = self.redirect_port + 1` to `self.preferred_port + 1` to avoid `TypeError` when `redirect_port` is `None` (before authentication has selected a port).

4. **Documentation**: Added troubleshooting section for port conflicts in QUICK_START.md.

## Impact of Removing Locking

The previous locking mechanism prevented multiple credential-process instances from authenticating simultaneously. Without it, if two processes need fresh credentials at the same time, both may open browser windows. Since users typically run many Claude Code instances, this could occur when credentials expire across multiple sessions. Cached credentials mitigate this in most cases, but a more robust locking mechanism (e.g., file-based) may be worth considering.

## Testing

- 199 passing tests (2 pre-existing failures on `main` unrelated to this change)
- Manual testing recommended:
  - Block port 8400 with `nc -l 8400`, verify auth works on auto-selected port
  - Verify auth still works on port 8400 when available

## Licensing

I confirm that this contribution is licensed under the MIT No Attribution license, consistent with the project's LICENSE file.

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -291,6 +291,17 @@ Force re-authentication:
 ~/claude-code-with-bedrock/credential-process --clear-cache
 ```
 
+### Port Conflicts
+
+The credential provider uses port 8400 by default for OAuth callbacks.
+If this port is in use by another application, authentication will automatically use an available port.
+
+To manually specify a different port, set the `REDIRECT_PORT` environment variable:
+
+```bash
+export REDIRECT_PORT=8401
+```
+
 ### Build Failures
 
 Check Windows build status:
--- a/source/credential_provider/__main__.py
+++ b/source/credential_provider/__main__.py
@@ -95,9 +95,10 @@ class MultiProviderAuth:
             )
         self.provider_config = PROVIDER_CONFIGS[self.provider_type]
 
-        # OAuth configuration
-        self.redirect_port = int(os.getenv("REDIRECT_PORT", "8400"))
-        self.redirect_uri = f"http://localhost:{self.redirect_port}/callback"
+        # OAuth configuration - port selection deferred until authentication
+        self.preferred_port = int(os.getenv("REDIRECT_PORT", "8400"))
+        self.redirect_port = None
+        self.redirect_uri = None
 
         # Initialize credential storage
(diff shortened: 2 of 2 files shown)


scouturier:
Thanks @spawnia, great diagnosis and clean fix — the port conflict issue is real and still present on `main`, so this is definitely worth merging!

One thing to add to this PR: `main` has a line `quota_port = self.redirect_port + 1` (around line 1795) that will throw a `TypeError` once your change initializes `redirect_port = None`. Could you add a fix for that in this PR — replacing it with `self.preferred_port + 1` should do it.

A rebase will also be needed — `__main__.py` has had quite a few changes since this branch was cut (keyring support, id_token caching).

spawnia:
Rebased onto current `main` and fixed the `quota_port` issue — changed `self.redirect_port + 1` to `self.preferred_port + 1`. The silent refresh logic that was added to `main` is preserved; only the port-locking blocks are removed. Also cleaned up the now-unused `_wait_for_auth_completion()` method.

dineshSajwan (approved):

PR #114, opened 2026-02-16
fix(deploy): Use correct Cognito OIDC issuer URL for quota monitoring

windson:
## Problem

When deploying the quota monitoring stack with Cognito User Pools, the API Gateway JWT authorizer fails because the hosted UI domain is used as the OIDC issuer URL, which does not serve the `/.well-known/openid-configuration` endpoint.

## Root Cause

The deploy code uses `profile.provider_domain` as the OIDC issuer URL for all provider types. For Cognito, this is the hosted UI domain, not the actual OIDC issuer.

## Fix

For Cognito providers, construct the issuer URL using the standard Cognito IDP format:
`https://cognito-idp.<region>.amazonaws.com/<user-pool-id>`

Other providers (Okta, Azure AD, Auth0) are unaffected.


windson:
Merged as part of #248

Issue #118, opened 2026-02-16
Region handling issues: no warning, cross-region detection, hardcoded ELB account ID


marmutto:
## Description

Multiple region-related issues affect the deployment experience.

### 3a. No warning when deploying to default CLI region

CloudFormation stacks deploy to the default region in your AWS CLI config (often `us-east-1`) without warning. Users intending to deploy in a different region must remember to pass `--region` explicitly.

### 3b. Cognito User Pool stack not detected cross-region

The init wizard says "No Cognito User Pool stack detected" if the stack was deployed in a different region than the CLI default, even if the user later specifies the correct region during init.

### 3c. ALB logs bucket policy hardcodes us-east-1 ELB account ID

The `landing-page-distribution.yaml` template hardcodes the ELB service account for `us-east-1` (`127311923021`). Deploying in any other region fails with an S3 access denied error.

**Error:**
```
AWS::ElasticLoadBalancingV2::LoadBalancer (DistributionALB): Resource handler returned message:
"Access Denied for bucket: claude-code-auth-alb-logs-<account-id>. Please check S3bucket permission"
```

## Suggested Fix for 3c

Replace the hardcoded account ID with a `Mappings` section that maps each region to its correct ELB service account, then use `!FindInMap` to resolve dynamically:

```yaml
AWS: !Sub
  - 'arn:aws:iam::${ELBAccountId}:root'
  - ELBAccountId: !FindInMap [ELBAccountMap, !Ref 'AWS::Region', AccountId]
```

Reference: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html

## Suggested Improvements

- Display the target region prominently before deploying
- Search for Cognito stacks across commonly used regions, or prompt the user for the region
- Use a region-to-ELB-account mapping in the CloudFormation template

wirjo:
Status update on beta branch:

- **3a (no region warning)** — still valid. No confirmation prompt when deploying to default CLI region. Open.
- **3b (cross-region Cognito detection)** — partially mitigated. The init wizard now uses the region explicitly set during `ccwb init`, reducing confusion. Edge cases remain.
- **3c (hardcoded ELB account IDs)** — mostly fixed. The `ELBServiceAccounts` mapping in `landing-page-distribution.yaml` now covers 20+ regions. A few newer regions (eu-south-2, ap-south-2, ap-southeast-4, etc.) may still be missing.

[...]

PR #127, opened 2026-02-18
fix: CloudFormation template bugs (ALB logs region, ECS DependsOn, CodeBuild SSL)

### LINKED ISSUE #121: Windows CodeBuild: SSL certificate error when Nuitka downloads GCC compiler
# Windows CodeBuild: SSL Certificate Error During Nuitka Compilation

## Issue
Windows builds fail during Nuitka compilation phase due to SSL certificate verification errors.

## Problem
Nuitka compilation fails in Windows CodeBuild when downloading GCC compiler from GitHub:
```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate
```

## Root Cause
Python 3.12 on Windows Server 2022 expects SSL certificates at `C:\Program Files\Common Files\SSL\cert.pem`, which doesn't exist in the CodeBuild container.

## Solution
1. Install `certifi` package (provides Mozilla's CA bundle)
2. Set `SSL_CERT_FILE` environment variable to point to certifi's certificate bundle

### CloudFormation Template Changes
Add environment variable to CodeBuild project:
```yaml
Environment:
  Type: WINDOWS_SERVER_2022_CONTAINER
  ComputeType: BUILD_GENERAL1_LARGE
  Image: aws/codebuild/windows-base:2022-1.0
  PrivilegedMode: false
  EnvironmentVariables:
    - Name: SSL_CERT_FILE
      Value: C:\Python312\Lib\site-packages\certifi\cacert.pem
      Type: PLAINTEXT
```

### Buildspec Changes
Add certifi installation to pre_build phase:
```yaml
pre_build:
  commands:
    - echo Installing certifi for SSL certificates...
    - C:\Python312\python.exe -m pip install certifi
```

## Files to Update
- `deployment/infrastructure/codebuild-windows.yaml`

## Notes
- The hardcoded path assumes Python 3.12 installed via `choco install python312`
- If Python version changes, update the `SSL_CERT_FILE` path accordingly
- Validated in actual failed build environment using CodeBuild sandbox mode

### LINKED ISSUE #122: ECS service race condition with HTTPS listener in otel-collector
## Description

When `HasCustomDomain` is true, the `HTTPListener` only redirects HTTP to HTTPS (no target group). The target group association with the load balancer happens in `HTTPSListener`. However, `ECSService` only has `DependsOn: HTTPListener`, so it can start before `HTTPSListener` exists.

## Error

```
target group does not have an associated load balancer
```

## File

`deployment/infrastructure/otel-collector.yaml` line ~460

## Suggested Fix

Add `HTTPSListener` to ECSService DependsOn. CloudFormation ignores DependsOn references to conditional resources when the condition is false, so this is safe.

```yaml
ECSService:
  Type: AWS::ECS::Service
  DependsOn:
    - HTTPListener
    - HTTPSListener
```

peepeepopapapeepeepo:
## Description

Fixes three CloudFormation template bugs that block deployment in various configurations.

Partially addresses #118 (section 3c), fixes #122, fixes #121

> Note: PR #112 also addresses #121 with a different approach (pre-installing MinGW). This PR instead fixes the root cause - broken SSL certificates in the Windows CodeBuild container - by installing `certifi` and configuring `sitecustomize.py`. This covers Nuitka's MinGW download and any future HTTPS downloads.

### Changes

**landing-page-distribution.yaml:**
- ALB access logs S3 bucket policy now uses `logdelivery.elasticloadbalancing.amazonaws.com` service principal instead of hardcoded us-east-1 ELB account ID (`127311923021`). Per [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html). Note: #118 suggests a `Mappings`+`FindInMap` approach; this uses the service principal instead, which is simpler and requires no region mapping table.

**otel-collector.yaml:**
- Added `HTTPSListener` to ECSService `DependsOn`. When `HasCustomDomain` is true, the target group is associated via HTTPSListener, not HTTPListener. Without this, ECS service creation races ahead and fails.

**codebuild-windows.yaml:**
- Added `certifi` to pip install and wrote `sitecustomize.py` to set `SSL_CERT_FILE`. The Windows Server 2022 CodeBuild container has outdated CA certificates causing Nuitka's MinGW download to fail with `SSL: CERTIFICATE_VERIFY_FAILED`.

### Testing

- Deployed full stack in ap-southeast-1 with landing-page distribution, Entra ID, and monitoring enabled
- Verified ALB access logs write successfully
- Verified ECS service starts after HTTPS listener is ready
- Verified CodeBuild Windows build completes successfully

### DIFF
--- a/deployment/infrastructure/codebuild-windows.yaml
+++ b/deployment/infrastructure/codebuild-windows.yaml
@@ -142,7 +142,8 @@ Resources:
               commands:
                 - echo Installing build dependencies...
                 - C:\Python312\python.exe -m pip install --upgrade pip
-                - C:\Python312\python.exe -m pip install nuitka==2.7.12 ordered-set zstandard
+                - C:\Python312\python.exe -m pip install nuitka==2.7.12 ordered-set zstandard certifi
+                - Set-Content -Path 'C:\Python312\Lib\site-packages\sitecustomize.py' -Value 'import os; os.environ["SSL_CERT_FILE"]=r"C:\Python312\Lib\site-packages\certifi\cacert.pem"'
                 - echo Installing application dependencies...
                 - C:\Python312\python.exe -m pip install boto3 requests PyJWT cryptography
                 - C:\Python312\python.exe -m pip install keyring pywin32
--- a/deployment/infrastructure/landing-page-distribution.yaml
+++ b/deployment/infrastructure/landing-page-distribution.yaml
@@ -175,7 +175,7 @@ Resources:
           - Sid: AWSLogDeliveryWrite
             Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::127311923021:root'  # ELB service account for us-east-1
+              Service: logdelivery.elasticloadbalancing.amazonaws.com
             Action: s3:PutObject
             Resource: !Sub '${ALBLogsBucket.Arn}/*'
           - Sid: AWSLogDeliveryAclCheck
--- a/deployment/infrastructure/otel-collector.yaml
+++ b/deployment/infrastructure/otel-collector.yaml
@@ -459,6 +459,7 @@ Resources:
     Type: AWS::ECS::Service
     DependsOn:
       - HTTPListener
+      - HTTPSListener
     Properties:
       ServiceName: otel-collector-service
       Cluster: !Ref ECSCluster


wirjo:
@dineshSajwan Bumping this one too

dineshSajwan (approved):

PR #160, opened 2026-03-17
Add Go credential process with beta opt-in during install

bluedoors:
## Summary

- Rewrites the credential-process in Go for significantly faster startup and lower memory usage
- Packages both Python and Go binaries during `ccwb package` — Go cross-compiles all 5 platforms (macOS arm64/intel, Linux x64/arm64, Windows) in seconds with no Docker required
- Install script presents users with a choice between the stable Python engine and the beta Go engine, defaulting to Python
- If Go toolchain is unavailable at build time, packaging continues gracefully with Python-only
- Distribution (`ccwb distribute`) updated to bundle Go binaries alongside Python ones

## How it works

During `./install.sh`, users see:
```
Credential Engine Selection

  [1] Standard (Python) - Stable, well-tested
  [2] Beta: High-performance (Go) - Faster startup, lower memory usage

Select credential engine [1]:
```

Selecting `[2]` installs the Go binary as `credential-process`. Default is Python. If no Go binary is present in the package, the prompt is skipped entirely.

## Test plan

- [x] Go cross-compilation verified for all 5 platforms (macOS arm64/intel, Linux x64/arm64, Windows)
- [x] Native binary runs and reports correct version (`1.0.0-beta`)
- [x] Version consistent across main.go, Makefile, and package.py
- [x] Python syntax check passes on modified `package.py` and `distribute.py`
- [ ] End-to-end `ccwb package` with Go builds
- [ ] Install script engine selection with both choices
- [ ] Verify Go binary authenticates correctly against IdP
- [ ] Windows `install.bat` engine selection (no Windows host available — reviewer to test)


wirjo:
Noting that PR #338 appears to be a more comprehensive version of this work — it includes:

- Both credential-process AND otel-helper rewritten in Go
- Pre-built binaries for all 5 platforms (no Go toolchain required for admins)
- Keyring split cache (`keyring_split.go`) addressing the DPAPI/securityd latency in #381
- Windows AV false positive mitigation (PE version info via go-winres)
- Tests: migration, keyring split, OIDC, confidential client
- `--prebuilt` flag for zero-compile deployments

#338 should supersede this PR. We could close this in favour of #338.

jamesbland123:
We're closing this pull request in favor of #338.

PR #186, opened 2026-04-02
feat(landing-page-dist): Improve download link handling for Landing Page Distribution

jphuynh:
*Issue #, if available:*

*Description of changes:*
With the current setup, Chrome is preventing the download of the package from the landing zone and raising a warning that is not great for user experience. This change is reworking the download URL so it's exposed via the trusted custom domain instead of the direct signed S3 url. Changing latest.zip into less generic named installers also helps with Chrome not marking them as suspicious.

```
                  'windows': 'claude-code-windows-installer.zip',
                  'linux': 'claude-code-linux-installer.zip',
                  'mac': 'claude-code-mac-installer.zip',
                  'all-platforms': 'claude-code-all-platforms-installer.zip',
```

### DIFF
--- a/deployment/infrastructure/landing-page-distribution.yaml
+++ b/deployment/infrastructure/landing-page-distribution.yaml
@@ -262,6 +262,12 @@ Resources:
                   if path.startswith('/docs'):
                       return serve_documentation(path)
 
+                  # Handle download redirects via custom domain (avoids Chrome Safe Browsing
+                  # warnings that occur when linking directly to s3.amazonaws.com URLs)
+                  if path.startswith('/download/'):
+                      platform = path.split('/download/', 1)[-1].strip('/')
+                      return handle_download(platform)
+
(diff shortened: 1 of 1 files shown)


wirjo:
Thanks @jphuynh — reviewed and rebased onto latest beta. This is clean and well thought out.

**What we like:**
- Platform allowlist prevents path traversal or unexpected routes
- `head_object` check before redirect (no blind 302 to missing keys)
- Proper 404/500 responses instead of the old silent `'#'` fallback
- `ResponseContentDisposition` with descriptive filenames — solves the Chrome Safe Browsing issue at the root cause
- Since the landing page stack already requires a custom domain for OIDC auth, this introduces no new dependencies

**Risk assessment:** Low. Single file (CF template inline Lambda), no schema changes, ALB authentication still gates all traffic. Strictly better UX for end users.

Rebased to latest beta — will merge shortly. Thanks for the contribution!

PR #203, opened 2026-04-02
fix: preserve existing profile fields when updating via init wizard

scouturier:
Fixes #196

## Summary

- When `ccwb init` is run on an existing profile, the wizard was reconstructing the `Profile` object from scratch using only the fields it prompts for — silently dropping any fields set by other commands (e.g. confidential client settings, certificate paths, keyring-backed secrets)
- This fix loads the existing profile first and updates only the fields the wizard manages, preserving everything else
- Also fixes a second gap: five quota fields (`daily_limit`, `burst_buffer_percent`, `daily_enforcement_mode`, `monthly_enforcement_mode`, `check_interval`) were missing from the pre-population step, so re-running init would reset them to defaults even when the user made no changes

## Test plan

- [ ] Run `ccwb init` on a profile that has fields set outside the wizard (e.g. `client_secret_keyring_service`, cert paths); confirm those fields survive the update
- [ ] Run `ccwb init` on a profile with non-default quota settings; confirm values are pre-populated and preserved if unchanged
- [ ] Run `ccwb init` on a new profile (no existing config); confirm clean creation still works

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1445,52 +1445,102 @@ class InitCommand(Command):
         if monitoring_dict.get("hosted_zone_id"):
             monitoring_config["hosted_zone_id"] = monitoring_dict["hosted_zone_id"]
 
-        profile = Profile(
-            name=profile_name,
-            provider_domain=config_data["okta"]["domain"],
-            client_id=config_data["okta"]["client_id"],
-            credential_storage=config_data.get("credential_storage", "session"),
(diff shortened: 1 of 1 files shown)


dineshSajwan:
This has a conflict with [PR 284](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/284). Can you please check if we still need this , if not then can we please close this one

wirjo:
Closing — the init.py has diverged too far to rebase this (15+ new fields since April). Rewritten as a fresh PR against current beta using the same load-and-mutate approach. Thank you @scouturier for the original design!

PR #205, opened 2026-04-03
fix(credential_provider): Increase RoleSessionName truncation limit to preserve full UUIDs

seeinyou:
Summary

  - Increase RoleSessionName identifier truncation from 32 to 52 characters for both sub and email paths
  - Fixes UUID-based IdP user IDs being cut off in CloudTrail/CloudWatch identity fields
  - The 52-char limit (+ 12-char prefix = 64) matches the AWS RoleSessionName maximum

  Test plan

  - Verify with a UUID-format sub claim (36 chars) that the full ID appears in the session name
  - Verify with a long email prefix (>32 chars) that truncation still works correctly
  - Confirm total RoleSessionName length does not exceed 64 characters

Closes #204

### DIFF
--- a/source/credential_provider/__main__.py
+++ b/source/credential_provider/__main__.py
@@ -965,12 +965,14 @@ class MultiProviderAuth:
             # Sanitize to replace invalid characters with hyphens
             session_name = "claude-code"
             if "sub" in token_claims:
-                # Use first 32 chars of sub for uniqueness, sanitized for AWS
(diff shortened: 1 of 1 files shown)


dineshSajwan (commented):
Thank you for the fix, It looks all good to me. I think we should add a test coverage for these new limits .
1. test case for  `RoleSessionName` length
2. test case for IdP sub claim length

PR #219, opened 2026-04-07
docs: document Route53 same-account requirement for custom domains

wirjo:
Fixes #216

ACM certificate validation requires the Route53 hosted zone to be in the same AWS account as the deployment. Cross-account hosted zones (e.g. shared networking account in Control Tower setups) cause deployment failures with no clear error.

Adds a "Custom Domain Requirements" section to the Prerequisites in README.md documenting this limitation and the workaround (migrate or delegate the zone).

1 file, +7 lines.


wirjo:
Superseded by #405 (consolidated batch targeting beta).

PR #220, opened 2026-04-07
fix(deploy): move distribution stack after networking in deploy order

wirjo:
Fixes #214

The landing-page distribution stack requires VPC outputs from the networking stack (`get_stack_outputs(networking_stack_name)` at line 449), but was queued **before** networking in the deploy-all sequence. This caused `ccwb deploy` to fail on fresh environments.

### Fix

Move the distribution block after the monitoring block (which contains networking). One block moved, no conditional logic.

**Before:** `auth → distribution → networking → monitoring → ...`
**After:** `auth → networking → monitoring → ... → distribution → codebuild`

1 file, +6/-5. Backward compatible — same stacks, just reordered.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -157,11 +157,11 @@ class DeployCommand(Command):
             # Deploy all configured stacks in dependency order
             stacks_to_deploy.append(("auth", "Authentication Stack (Cognito + IAM)"))
 
-            # Deploy distribution after networking if it's landing-page type
-            if profile.enable_distribution:
+            # Deploy presigned-s3 distribution early (no networking dependency)
+            if profile.enable_distribution and getattr(profile, "distribution_type", "") != "landing-page":
                 stacks_to_deploy.append(("distribution", "Distribution infrastructure (S3 + IAM)"))
 
-            # Deploy remaining monitoring stacks
+            # Deploy monitoring stacks
             if profile.monitoring_enabled:
                 vpc_config = profile.monitoring_config or {}
                 if vpc_config.get("create_vpc", True):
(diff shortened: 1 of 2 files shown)


source/claude_code_with_bedrock/cli/commands/deploy.py
```diff
@@ -175,6 +175,11 @@ def handle(self) -> int:
                 # Check if quota monitoring is enabled
                 if getattr(profile, "quota_monitoring_enabled", False):
                     stacks_to_deploy.append(("quota", "Quota Monitoring (Per-User Token Limits)"))
+
+            # Deploy landing-page distribution after networking (requires VPC outputs)
+            if profile.enable_distribution and getattr(profile, "distribution_type", "") == "landing-page":
+                stacks_to_deploy.append(("distribution", "Distribution infrastructure (Landing Page)"))
```
scouturier, line 188: if distribution_type == "landing-page" but monitoring_enabled=False, no VPC is created yet the append still runs, it might be causing a CloudFormation error at deploy time instead of a clear message; 

I suggest to add a guard here or add monitoring_enabled as a requirement for landing-page in validators.py

wirjo: Good catch — added the cross-validation in `validators.py` (f3d4608). Now landing-page distribution with `monitoring_enabled=False` is caught at config validation time too, not just at deploy.

wirjo:
Closing — this fix (deploy order: distribution after networking) is already on main. The deploy logic at line 172 now places distribution after the networking stack.

wirjo:
Correction: my earlier close reason was inaccurate — this bug is real and was not already fixed on main. The deploy order does place distribution before networking, which breaks landing-page deployments.

However, #418 provides a better fix for this same issue:
- Only adds networking before distribution when `distribution_type == "landing-page"` (presigned-s3 doesn't need VPC)
- Includes de-duplication logic to avoid deploying networking twice if monitoring also needs it
- Smaller diff (+5/-2 vs +25/-3)

Superseded by #418.

PR #146, opened 2026-03-10
fix: Cognito schema, quota JWT issuer, Okta preview domains, and Windows packaging

pipeflo:
## Summary

Multiple bug fixes found during Cognito and Okta deployments, plus Windows native packaging fixes and documentation:

### Bug Fixes

1. **Cognito `preferred_username` schema conflict** — `Required: true` conflicts with `AliasAttributes`, preventing users from logging in via Hosted UI. Changed to `Required: false`.

2. **Quota JWT authorizer issuer mismatch** — For Cognito, the quota API Gateway JWT authorizer receives the hosted UI domain, but Cognito tokens use `https://cognito-idp.<region>.amazonaws.com/<pool-id>` as the `iss` claim. All quota API requests are rejected.

3. **Okta preview domains rejected** — The `OktaDomain` CloudFormation parameter regex only accepts `.okta.com` and `.okta-emea.com`, rejecting `.oktapreview.com` (Okta sandbox/test environments).

4. **Windows native package build broken** — Three issues:
   - `binary_name` variable never set for Windows target, causing `cannot access local variable` error
   - Nuitka availability check uses `which` (Unix-only), fails on Windows
   - OTEL helper has no native Windows build path — always errors when monitoring is enabled

### Documentation

5. Updated `cognito-user-pool-setup.md` to reflect `preferred_username` as optional.

6. Added **"Building from a Windows Machine"** section to `WINDOWS_BUILD_SYSTEM.md` — covers native Windows builds with Nuitka, cross-platform build matrix, multi-platform distribution strategy, and troubleshooting. The existing docs only covered building via CodeBuild from macOS/Linux.

## Files Changed

| File | Change |
|------|--------|
| `deployment/infrastructure/cognito-user-pool-setup.yaml` | `preferred_username` Required: false |
| `assets/docs/providers/cognito-user-pool-setup.md` | Docs: preferred_username is optional |
| `source/claude_code_with_bedrock/cli/commands/deploy.py` | Cognito issuer URL for quota stack |
| `deployment/infrastructure/bedrock-auth-okta.yaml` | Accept `.oktapreview.com` domains |
| `source/claude_code_with_bedrock/cli/commands/package.py` | Windows binary_name, Nuitka check, native OTEL helper |
| `assets/docs/WINDOWS_BUILD_SYSTEM.md` | New section: Building from a Windows Machine |

## Test plan

- [x] Deployed Cognito User Pool with `preferred_username Required: false`
- [x] Verified users can log in via Hosted UI
[...]

PR #221, opened 2026-04-07
fix: add encoding="utf-8" to all text-mode open() calls

wirjo:
Fixes #209

Python defaults to the system encoding on Windows (often Windows-1252). Any file containing non-ASCII characters causes `UnicodeDecodeError: 'charmap' codec can't decode byte`. This affects CFN templates, installer scripts, and config files.

Adds explicit `encoding="utf-8"` to all 51 text-mode `open()` calls across 19 files. Binary-mode calls (`"rb"`, `"wb"`) are unchanged.

Note: [#146](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/146) fixes 2 of these in `package.py`. This PR covers the remaining 49 plus those 2 for completeness.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/builds.py
+++ b/source/claude_code_with_bedrock/cli/commands/builds.py
@@ -180,7 +180,7 @@ class BuildsCommand(Command):
                     console.print("[red]No recent builds found. Start a build with 'poetry run ccwb package'[/red]")
                     return 1
 
-                with open(build_info_file) as f:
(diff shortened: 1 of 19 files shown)


wirjo:
Closing — the fix is still relevant (many `open()` calls lack `encoding='utf-8'` for Windows compatibility) but this PR's branch is too diverged to rebase. Will recreate as a fresh PR targeting beta when addressing Windows compatibility.

PR #238, opened 2026-04-08
feat: Support resource tagging on all deployed CloudFormation stacks

### LINKED ISSUE #152: [Feature] Support resource tagging on all deployed CloudFormation stacks
**Description:**

The solution does not pass tags to any deployed CloudFormation stack. The `deploy_stack()` method in `cloudformation.py` already accepts a `tags` parameter (line 87), but no caller in `deploy.py` ever provides tags.

**Expected behavior:**

- A `tags` configuration option in the profile config (via `ccwb init` or config file).
- All `deploy_stack()` calls in `deploy.py` should pass the configured tags.
- Tags should propagate to nested resources via CloudFormation stack-level tagging.

**Impact:** Customers cannot use MAP credit tracking, cost allocation, or organizational governance tags on the deployed infrastructure.

juengler:
Closes #152

### What

Adds support for tagging all CloudFormation stacks deployed via `ccwb deploy`. Tags propagate to nested resources automatically via CloudFormation stack-level tagging.

### Changes

- `config.py` — Added `tags: dict[str, str]` field to the `Profile` dataclass. Defaults to empty dict, backward compatible with existing profiles.
- `init.py` — Added optional "Resource Tags" step at the end of `ccwb init` wizard. Users can add key-value pairs interactively.
- `deploy.py` — `deploy_with_cf` helper now reads `profile.tags` and passes them to `cf_manager.deploy_stack()`. Empty tags are a no-op (nothing sent to CloudFormation API).

### Testing

- 192/196 tests pass. The 4 failures are pre-existing on `main` (unrelated JWT algorithm and source region test mismatches).
- Backward compatibility verified: old profile JSON files without `tags` key load without errors.


dineshSajwan:
### Code review

Found 3 issues:

1. **Whitespace-only tag key bypasses validation and causes a deploy-time CloudFormation failure.** The check `if not tag_key: break` accepts strings like `"   "` (spaces only) because non-empty whitespace is truthy in Python. The result is silently stored in the config and passed to CloudFormation, which rejects it at deployment. Fix: call `.strip()` before the guard so the empty-after-stripping case correctly breaks the loop.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/9394d7df8773ce2a750da0cee058d6b36aaced73/source/claude_code_with_bedrock/cli/commands/init.py#L1291-L1298

2. **Reserved `aws:` tag key prefix is not rejected in the wizard.** AWS CloudFormation prohibits any tag key starting with `aws:` (case-insensitive). Without a guard, a user entering e.g. `aws:environment` will pass the wizard successfully but see the deployment fail at the CloudFormation API level. Fix: after collecting `tag_key`, check `if tag_key.lower().startswith("aws:")`, print an error, and `continue` the loop.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/9394d7df8773ce2a750da0cee058d6b36aaced73/source/claude_code_with_bedrock/cli/commands/init.py#L1291-L1305

3. **Configured tags are not shown in the Step 4 configuration review table.** Every other optional feature (monitoring, quota, analytics, VPC, model, regions) has a `table.add_row()` entry so the user can verify it before confirming. Tags collected in Step 3 are silently omitted, so users cannot catch mistakes before the config is saved and deployment starts. Fix: add a row after the AWS Account row — e.g. `if config.get("tags"): table.add_row("Resource Tags", ", ".join(f"{k}={v}" for k, v in config["tags"].items()))`.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/9394d7df8773ce2a750da0cee058d6b36aaced73/source/claude_code_with_bedrock/cli/commands/init.py#L1401-L1408

---

**What the PR does:** adds an optional `tags` field to `Profile`, a tagging step in `ccwb init`, and passes the tags to all `deploy_stack()` calls so every CloudFormation stack and its nested resources inherits them. The serialisation round-trip and backward compatibility with existing profiles are handled correctly via `field(default_factory=dict)`. The `cf_manager.deploy_stack()` utility already converts `dict[str, str]` to the boto3 `[{"Key": k, "Value": v}]` format. The credential-provider binary is unaffected — it has no dependency on CloudFormation tags.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1257,6 +1257,58 @@ def _gather_configuration(self, progress: WizardProgress, existing_config: dict[
...
+                        console.print(f"[green]✓[/green] Tag: {key}={tag_value}")
+                    else:
+                        console.print(f"[yellow]✗[/yellow] Removed tag: {key}")
+
+            # Then allow adding new tags
+            while True:
+                tag_key = questionary.text(
+                    "New tag key (empty to finish):",
+                    default="",
+                ).ask()
```
dineshSajwan, line 1432: I think we should also trim the space , otherwise it may fail at` ccwb deploy`
 add this line `tag_key = (tag_key or "").strip() `

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1257,6 +1257,58 @@ def _gather_configuration(self, progress: WizardProgress, existing_config: dict[
...
+                if not tag_key:
+                    break
+                tag_value = questionary.text(
```
dineshSajwan, line 1442: AWS reserves tag keys that start with `aws:` for its own internal use. May be we should add validation to not use aws: prefix 
```
 if tag_key.lower().startswith("aws:"):      
                      console.print("[red]✗ Tag keys cannot start with 'aws:' (reserved by AWS)")
```
Let me know if you agree with this.

juengler:
Thanks for the review! I rebased on main to resolve the conflicts and addressed your feedback:

- Added tag_key = (tag_key or "").strip() to catch whitespace-only input
- Added aws: prefix validation
- Tags now show in the Step 4 review table
- Also added key length (128) and value length (256) limit checks

Let me know if this looks good.

dineshSajwan:
Thanks @juengler 
can you please update the documentation in quick_start.md

1. update the decision tree, add after option features , something like this
```
Resource Tags (Optional)                                                                                                     
  +│   │   ├── No -> skip to "What ccwb init saves"                                                                                      
  +│   │   └── Yes                                                                                                                      
  +│   │       └── Enter key-value pairs (up to 50)   
```
2. Detail about resource tags, an example how it helps. 
3. are these tags saved in `ccwb init`, if yes then update that too.

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1346,6 +1346,68 @@ def _gather_configuration(self, progress: WizardProgress, existing_config: dict[
...
+                console.print("[dim]Existing tags (edit value or leave empty to remove):[/dim]")
+                for key, value in existing_tags.items():
+                    tag_value = questionary.text(
+                        f"  {key}:",
+                        default=value,
+                    ).ask()
+                    if tag_value is None:
+                        return None
+                    if tag_value:
+                        tags[key] = tag_value
```
dineshSajwan, line 1422: can we add regex validation on tag and allow only those values which are allowed by cloud formation. and also add a length validation of 50 characters.

dineshSajwan:
@juengler can you please resolve the conflicts

dineshSajwan (approved):

Issue #191, opened 2026-04-02
[Feature] Strengthen quota enforcement to prevent client-side bypass


scouturier:
## Summary

The current quota enforcement mechanism relies on client-side configuration (`config.json` and the OTel endpoint setting in Claude). This means a technically proficient user can bypass quota controls by:

- Editing `config.json` directly
- Removing or redirecting the OTel endpoint from Claude settings
- Redirecting endpoints via `/etc/hosts`

For enterprise deployments, quota enforcement should be authoritative and server-side so that individual users cannot opt out of it.

## Proposed improvements

- Move quota enforcement to the server/gateway layer, so it cannot be overridden by client-side changes
- Evaluate a lightweight API gateway approach (similar to existing commercial solutions) that intercepts Bedrock calls and applies quota rules centrally
- The `embedded_config` work (#188) is a step in the right direction for reducing reliance on editable local config — this issue covers the broader gateway-level enforcement goal
- Document the current trust model clearly so operators are aware of the current limitations while a more robust solution is developed

## Benefit

Organisations that need reliable, auditable quota enforcement (e.g. for cost control or compliance) require a server-side guarantee rather than a best-effort client-side control.

PR #244, opened 2026-04-10
[Feature] Strengthen quota enforcement to prevent client-side bypass

usumfabricae:
*Issue #191 

*Description of changes:*
- Per-user Bedrock Application Inference Profiles: Each user gets dedicated inference profiles (one per model) provisioned
  automatically on first login. Metrics are tracked server-side by AWS, eliminating blind spots from misconfigured or missing    
  OTEL clients.   
- Server-side quota enforcement: A new QuotaEnforcer Lambda tags profiles status=enabled/disabled every 5 minutes. IAM         
  conditions block InvokeModel at the AWS level when a profile is disabled — no client-side bypass possible.                     
- Secure profile provisioning via Lambda: A dedicated InferenceProfileProvisionerFunction is the sole principal with
  bedrock:CreateInferenceProfile and bedrock:TagResource. Users lose direct Bedrock management-plane access and gain only        
  lambda:InvokeFunction on that specific ARN.
- OTEL + inference profiles coexistence: Both can now be deployed together. When inference profiles are active, OTEL is        
  configured to filter out token metrics to prevent double-counting in the dashboard.                                            
- Shiv .pyz packaging: Alternative to PyInstaller for more portable cross-platform distribution — requires only Python 3.10+,
  no native binary.                                                                                                              
- Lambda naming standardization: All Lambda functions renamed to a consistent ClaudeCode-* prefix.
                                                                                                                                 
  
  Test plan                                                                                                                      
   
  - Deploy with inference_profiles_enabled = false — verify no behavioral change vs. main                                        
  - Deploy with inference_profiles_enabled = true — verify profiles are created on first login and ~/.claude.json is patched with
   the default model ARN                                                                                                         
  - Verify ABAC: user A cannot invoke user B's inference profile
  - Exceed quota → wait up to 17 min → verify InvokeModel returns AccessDeniedException                                          
  - Deploy with both inference_profiles_enabled = true and OTEL enabled → verify token metrics appear once in the dashboard (not  doubled)                                                                                                                       
  - Run ccwb profiles list and ccwb profiles set-default sonnet-4-6 — verify correct output and ~/.claude.json update            
  - Build .pyz package — verify install script runs --setup-profiles and sets correct model env vars in ~/.claude/settings.json


wirjo:
Thank you @usumfabricae for the thorough work on quota enforcement strengthening — the thinking here is solid and the gap you identified (client-side bypass) is real.

Unfortunately, the PR is too large to merge safely (+6,687/-672, 43 files). At this scale, it's very difficult to review for regressions, and the blast radius of a bug would be significant.

The good news: several of the concerns raised here have since been addressed incrementally on beta:
- Quota enforcement now runs server-side via Lambda (no client bypass possible)
- Cost-based enforcement with per-model pricing is in progress
- Fine-grained policies with per-user/group resolution are live
- Fail-closed security mode (`ERROR_HANDLING_MODE=fail_closed`) is the default

If there are specific remaining gaps you'd like to address, we'd welcome smaller, focused PRs targeting one concern at a time (~300 lines max). Happy to help scope those out.

Thanks again for the contribution!

PR #245, opened 2026-04-10
feat: add IAM principal cost attribution for Bedrock usage

scouturier:
## Summary
- Use the federated user's email as the STS session name so the principal ARN (`assumed-role/RoleName/alice@acme.com`) appears in the CUR 2.0 `line_item_iam_principal` column, enabling per-user cost visibility without IdP-side configuration
- Remove dead code that built session tags but never sent them (`AssumeRoleWithWebIdentity` has no `Tags` parameter)
- Add `COST_ATTRIBUTION.md` covering built-in tracking, IAM principal tags, and optional session tags with IdP-specific examples (Auth0, Okta, Entra ID)
- Cross-reference from each provider setup guide

## Test plan
- [x] Verified on account 373525597636: CUR 2.0 export has `INCLUDE_IAM_PRINCIPAL_DATA: TRUE` and `line_item_iam_principal` in query
- [x] CloudTrail confirms email-based session name (`assumed-role/BedrockCognitoFederatedRole/sebcou@amazon.com`)
- [x] Cost Explorer returns data grouped by `iamPrincipal/` tags
- [x] Documentation reviewed against [AWS IAM principal cost allocation docs](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/iam-principal-cost-allocation.html)


dineshSajwan (approved):
Thank you, this is really helpful.

PR #247, opened 2026-04-11
docs: document ECS service-linked role prerequisite for standalone deploys

wirjo:
Fixes #236

`ccwb deploy` already creates the `AWSServiceRoleForECS` service-linked role automatically (line 583 in deploy.py). However, users deploying the `otel-collector.yaml` template directly via CloudFormation hit a "could not assume ECS service role" error in new accounts.

### Changes

- **Custom Resource Lambda** in `otel-collector.yaml` — idempotently creates `AWSServiceRoleForECS` (ignores if already exists). Scoped IAM: only `iam:CreateServiceLinkedRole` on the ECS SLR ARN. No-ops on stack Delete.
- **`ECSService` DependsOn** — waits for the custom resource before launching Fargate
- **README.md** — documents the prerequisite for awareness
- **Template description** — notes the prerequisite for CloudFormation console visibility

Works for both `ccwb deploy` users (existing code handles it) and standalone CFN deployers (new custom resource handles it).

Related: #71 makes the analytics/monitoring stack deployable independently without requiring the auth stack.

PR #252, opened 2026-04-13
fix: add IAM propagation delay after creating ECS service linked role

scouturier:
## Summary
- Fixes #236 (together with #247 which covers the standalone CloudFormation deploy path)
- After creating `AWSServiceRoleForECS`, IAM propagation takes several seconds before the role is globally consistent
- Without a delay, CloudFormation immediately attempts ECS cluster creation and fails with "Unable to assume the service linked role"
- Adds a 10-second wait only when the role is newly created (no delay if it already exists)

## Test plan
- [x] Deploy monitoring stack on an account where `AWSServiceRoleForECS` does not yet exist
- [x] Verify ECS cluster creation succeeds after the propagation wait
- [x] Verify no delay occurs when the role already exists


wirjo:
Small, targeted fix — 4 lines adding a 10s IAM propagation delay after ECS SLR creation. Prevents race condition on fresh accounts. Rebased on current beta and merging.

PR #254, opened 2026-04-13
feat: implement --output/-o flag for ccwb config export

### LINKED ISSUE #253: Bug: ccwb config export --output flag documented but not implemented
## Summary

`CLI_REFERENCE.md` documents `--output <file>` as a supported option for `ccwb config export`, but the flag was never implemented. The command always outputs to stdout regardless.

## Impact

On Windows PowerShell 5.x (the default on Windows 10/11), shell redirection (`>`) writes UTF-16 LE instead of UTF-8, which breaks JSON parsing downstream. Users following the docs expecting `--output` to work get an unrecognised option error instead.

## Expected behavior

```bash
ccwb config export --output prod-config.json
```

Creates a UTF-8 encoded `prod-config.json` file.

## Actual behavior

```
error: option "--output" does not exist
```

## Fix

Implement `--output/-o` in `ConfigExportCommand` to write UTF-8 output to a file, with stdout as the default (preserving existing behavior).

scouturier:
## Summary

- Implements the `--output/-o` flag for `ccwb config export`, which was documented in `CLI_REFERENCE.md` since v2.0.0 but never wired up in code
- Also wires up `--include-secrets` which was doc-only too
- Fixes the docs to clarify that the default behavior is stdout (not `<profile-name>.json` as previously stated)

Closes #253

## Why --output matters on Windows

Shell redirection (`>`) in PowerShell 5.x (the default on Windows 10/11) writes UTF-16 LE instead of UTF-8, which breaks JSON parsing downstream. Writing via Python's `open(..., encoding="utf-8")` is the safe cross-platform approach.

## Test plan

- [ ] `ccwb config export` — JSON to stdout, status to stderr (unchanged behavior)
- [ ] `ccwb config export --output prod-config.json` — creates UTF-8 file, nothing on stdout
- [ ] `ccwb config export -o prod-config.json` — shorthand works
- [ ] `ccwb config export --include-secrets` — secrets not redacted in output
- [ ] `ccwb config export --help` — both flags appear in help text

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -1013,7 +1013,7 @@ poetry run ccwb config export [profile-name] [options]
 
 **Options:**
 
-- `--output <file>` - Output file path (default: `<profile-name>.json`)
+- `--output, -o <file>` - Write output to a file instead of stdout (uses UTF-8 encoding)
 - `--include-secrets` - Include sensitive values (not recommended)
 
 **What it does:**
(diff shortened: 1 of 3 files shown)


scouturier:
Closing in favour of a clean branch rebased onto main — see replacement PR.

Issue #131, opened 2026-02-21
quota_api_endpoint is not in the config


pbchekin:
`_create_config` does not write `quota_api_endpoint` to the config:

```
        config = {
            profile_name: {
                "provider_domain": profile.provider_domain,
                "client_id": profile.client_id,
                "aws_region": profile.aws_region,
                "provider_type": profile.provider_type or self._detect_provider_type(profile.provider_domain),
                "credential_storage": profile.credential_storage,
                "cross_region_profile": profile.cross_region_profile or "us",
            }
        }
```

wirjo:
Fixed — `quota_api_endpoint` is now a standard field in the Profile dataclass and is set during `ccwb init` when quota monitoring is enabled. The package command includes it in config.json for the Go binary.

PR #256, opened 2026-04-14
fix: Quota enforcement

jphuynh:
**This PR includes 2 commits that fixes quota enforcement**

**Server side change:**
The quota_check Lambda resolves enforcement limits exclusively from the QuotaPolicies DynamoDB table. However, nothing in the deployment pipeline seeds that table — ccwb deploy passes limits as CloudFormation parameters that become Lambda env vars (MONTHLY_TOKEN_LIMIT, MONTHLY_ENFORCEMENT_MODE, etc.), but the Lambda never reads them.

The result is that resolve_quota_for_user() always returns None (no policy), and every request gets {"allowed": true, "reason": "no_policy"} — unlimited access even when the operator configured enforcement_mode: block.

The similar quota_monitor Lambda already handles this correctly: when ENABLE_FINEGRAINED_QUOTAS is false it short-circuits with env var defaults instead of querying DynamoDB. (https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/deployment/infrastructure/lambda-functions/quota_monitor/index.py#L292-L303)

This commit adds a similar pattern to quota_check.

**Client side change:**
This change ensures that the distributed `config.json` includes `quota_api_endpoint`, `quota_fail_mode`, and `quota_check_interval` so the credential provider on users' machines knows where to call quota enforcement. Without that the quotas aren't being properly enforced and Claude Code allows going past the allowances.

-- 

### DIFF
--- a/deployment/infrastructure/lambda-functions/quota_check/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_check/index.py
@@ -20,6 +20,14 @@ POLICIES_TABLE = os.environ.get("POLICIES_TABLE", "QuotaPolicies")
 MISSING_EMAIL_ENFORCEMENT = os.environ.get("MISSING_EMAIL_ENFORCEMENT", "block")
 ERROR_HANDLING_MODE = os.environ.get("ERROR_HANDLING_MODE", "fail_closed")
 
+# Default limits from environment (used when fine-grained quotas are disabled)
+ENABLE_FINEGRAINED_QUOTAS = os.environ.get("ENABLE_FINEGRAINED_QUOTAS", "false").lower() == "true"
+MONTHLY_TOKEN_LIMIT = int(os.environ.get("MONTHLY_TOKEN_LIMIT", "0"))
+DAILY_TOKEN_LIMIT = int(os.environ.get("DAILY_TOKEN_LIMIT", "0"))
+MONTHLY_ENFORCEMENT_MODE = os.environ.get("MONTHLY_ENFORCEMENT_MODE", "block")
+WARNING_THRESHOLD_80 = int(os.environ.get("WARNING_THRESHOLD_80", "240000000"))
+WARNING_THRESHOLD_90 = int(os.environ.get("WARNING_THRESHOLD_90", "270000000"))
+
 # DynamoDB tables
 quota_table = dynamodb.Table(QUOTA_TABLE)
 policies_table = dynamodb.Table(POLICIES_TABLE)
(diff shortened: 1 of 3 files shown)


source/claude_code_with_bedrock/cli/commands/package.py
```diff
@@ -1731,6 +1731,12 @@ def _create_config(
                 console.print("[dim]  AZURE_CLIENT_CERTIFICATE_PATH=<path/to/cert.pem>[/dim]")
                 console.print("[dim]  AZURE_CLIENT_CERTIFICATE_KEY_PATH=<path/to/key.pem>[/dim]\n")
 
+        # Add quota settings so the credential provider can enforce limits
```
scouturier, line 1734: Worth double-checking the `quota_monitoring_enabled` gate here — elsewhere in the codebase (e.g. `package.py:1712`) the pattern is to gate solely on `quota_api_endpoint` being set, since that's the field that activates the feature. A profile could have `quota_api_endpoint` configured without `quota_monitoring_enabled` being `True`, which would cause the fields to be silently dropped from the export.

Would it make sense to simplify to:
```python
if getattr(profile, "quota_api_endpoint", None):
```

deployment/infrastructure/lambda-functions/quota_check/index.py
```diff
@@ -20,6 +20,14 @@
 MISSING_EMAIL_ENFORCEMENT = os.environ.get("MISSING_EMAIL_ENFORCEMENT", "block")
 ERROR_HANDLING_MODE = os.environ.get("ERROR_HANDLING_MODE", "fail_closed")
 
+# Default limits from environment (used when fine-grained quotas are disabled)
+ENABLE_FINEGRAINED_QUOTAS = os.environ.get("ENABLE_FINEGRAINED_QUOTAS", "false").lower() == "true"
+MONTHLY_TOKEN_LIMIT = int(os.environ.get("MONTHLY_TOKEN_LIMIT", "0"))
+DAILY_TOKEN_LIMIT = int(os.environ.get("DAILY_TOKEN_LIMIT", "0"))
+MONTHLY_ENFORCEMENT_MODE = os.environ.get("MONTHLY_ENFORCEMENT_MODE", "block")
+WARNING_THRESHOLD_80 = int(os.environ.get("WARNING_THRESHOLD_80", "0"))
```
scouturier: `WARNING_THRESHOLD_80` and `WARNING_THRESHOLD_90` default to `0` here, but `quota_monitor/index.py` defaults them to `240_000_000` and `270_000_000`.

Aligning the defaults would ensure both Lambdas behave consistently when the env vars are absent:
```python
WARNING_THRESHOLD_80 = int(os.environ.get("WARNING_THRESHOLD_80", "240000000"))
WARNING_THRESHOLD_90 = int(os.environ.get("WARNING_THRESHOLD_90", "270000000"))
```

source/claude_code_with_bedrock/cli/commands/package.py
jphuynh: Ha that's a good point thanks. I'll update the PR

jphuynh:
This is now ready for review. I've tested internally with our E2E workflow to deploy Claude Code with Bedrock.

Quotas are now enforced as expected (Minus the rendering of the emoji on the page:)):
Screenshot 2026-04-15 at 13 59 07

jphuynh:
Added the last commit to fix the emoji rendering

Screenshot 2026-04-15 at 14 48 24

scouturier:
LGTM

Closes issue #131

scouturier (approved):
LGTM

dineshSajwan (approved):

PR #207, opened 2026-04-06
Centralize model catalog, CRIS-aware env vars, and region alignment

wirjo:
## Summary

Replaces hardcoded model IDs, stale region lists, and brittle model resolution with a centralized model catalog (`CLAUDE_MODELS` in `models.py`) that serves as the single source of truth for all model, CRIS profile, and region data across the project.

## Problem

- `package.py` hardcoded a stale Haiku 3.5 model ID as `ANTHROPIC_SMALL_FAST_MODEL`
- Only `ANTHROPIC_MODEL` and `ANTHROPIC_SMALL_FAST_MODEL` were set — Claude Code's other tier env vars (`DEFAULT_SONNET_MODEL`, `DEFAULT_OPUS_MODEL`, `DEFAULT_HAIKU_MODEL`) were unset, causing unpredictable CRIS prefix selection for secondary models
- Model aliases like `opusplan` would resolve secondary models to the wrong geography (e.g. EU user getting US-routed Haiku)
- Auth templates had inconsistent and incomplete `AllowedBedrockRegions` defaults
- Display name dicts in `init.py` duplicated model data that could drift from source of truth
- No CI validation to catch stale model or region data

### Centralized model catalog (`models.py`)
- Added Haiku 4.5 and Haiku 3.5 to `CLAUDE_MODELS` (were completely missing)
- Added `resolve_model_for_tier(tier, cris_prefix)` — resolves the best available model for a tier (haiku/sonnet/opus) with automatic fallback chain when exact CRIS prefix unavailable
- Added `MODEL_TIER_PREFERENCES` and `MODEL_RATE_LIMITS` for downstream use
- Added `get_rate_limits_for_model()`, `get_throttle_metrics()` helpers

### CRIS-aware env vars (`package.py`)
- Set all 5 model env vars using the admin's `cross_region_profile` from `ccwb init`
- Uses `resolve_model_for_tier()` instead of string-parsing model IDs
- Correctly handles aliases (`opusplan`, `opus`, etc.) — `DEFAULT_*` vars ensure right CRIS prefix per tier
- Removed hardcoded Haiku 3.5 suffix

### IAM permissions (5 auth templates)
- Added `bedrock:ListFoundationModels` and `bedrock:ListInferenceProfiles` (read-only, no cost)
- Required for Claude Code's built-in model discovery fallback

### Region alignment
- Aligned `AllowedBedrockRegions` across all auth templates to 31 commercial regions derived from `models.py`
[...]

PR #257, opened 2026-04-14
feat: add Application Inference Profile and support 3-Tier Model Pinning

### LINKED ISSUE #151: [Feature] Add Application Inference Profile support for Bedrock cost allocation
**Description:**

The solution does not support [Application Inference Profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html), which are the AWS-recommended mechanism for cost allocation and usage tracking on Bedrock. Application Inference Profiles allow customers to attach tags to on-demand model invocation requests for cost tracking.

**Expected behavior:**

- Users should be able to configure an Application Inference Profile ARN in their profile config.
- When configured, the profile ARN should be used as the `modelId` in Bedrock API calls instead of the CRIS model ID.

**Impact:** Customers cannot use AWS-recommended cost allocation tagging for Bedrock usage without modifying the Python scripts themselves.

juengler:
## Description

Adds support for [Application Inference Profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html) and migrates model configuration from the legacy `ANTHROPIC_MODEL` + deprecated `ANTHROPIC_SMALL_FAST_MODEL` to Anthropic's recommended three-tier model pinning.

Closes #151

### Three-tier model pinning

The `ccwb package` command now outputs the recommended `ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, and `ANTHROPIC_DEFAULT_HAIKU_MODEL` environment variables in `settings.json`, replacing the legacy `ANTHROPIC_MODEL` + deprecated `ANTHROPIC_SMALL_FAST_MODEL`. All three tiers are auto-derived from the primary model selection:

| User picks | Opus tier | Sonnet tier | Haiku tier |
|---|---|---|---|
| Opus | selected CRIS ID | Sonnet 4.6 (same prefix) | Haiku 4.5 (same prefix) |
| Sonnet | (not set) | selected CRIS ID | Haiku 4.5 (same prefix) |
| Haiku/other | (not set) | (not set) | selected CRIS ID |

### Application Inference Profile support

Admins can optionally configure up to three Application Inference Profile ARNs (one per model tier) during `ccwb init`. When configured, the ARNs replace the CRIS model IDs in the corresponding `ANTHROPIC_DEFAULT_*_MODEL` environment variables.

### IAM policy update

All five auth CloudFormation templates now include `application-inference-profile/*` as an allowed resource for Bedrock invoke actions, alongside the existing `foundation-model/*` and `inference-profile/*` patterns.

### Backward compatibility fix

`Profile.from_dict()` now filters out unknown fields before constructing the dataclass, preventing `TypeError` when loading profiles that contain fields from other branches or older/newer versions.

## Files changed

- `config.py` — Three per-tier inference profile ARN fields on Profile, resilient `from_dict` field filtering
- `validators.py` — `validate_application_inference_profile_arn()` for aws + aws-us-gov partitions
- `package.py` — Three-tier `ANTHROPIC_DEFAULT_*_MODEL` env vars, inference profile ARN override logic
- `init.py` — Optional per-tier inference profile prompt after model selection, display in review/summary
- `context.py` — Display per-tier inference profile ARNs in `ccwb context show`
- `bedrock-auth-okta.yaml` — `application-inference-profile/*` in IAM policy
- `bedrock-auth-azure.yaml` — `application-inference-profile/*` in IAM policy
- `bedrock-auth-auth0.yaml` — `application-inference-profile/*` in IAM policy
- `bedrock-auth-cognito-pool.yaml` — `application-inference-profile/*` in IAM policy
- `cognito-identity-pool.yaml` — `application-inference-profile/*` in IAM policy

## Testing

- 192/196 unit tests pass (4 pre-existing failures unrelated to this change)
- Manually tested end-to-end: created Application Inference Profiles with tags in eu-central-1, configured via `ccwb init`, packaged, and verified Claude Code uses the inference profile ARNs for all three model tiers
- Verified backward compatibility with existing profiles that lack the new fields
- Verified ARN validation rejects invalid formats with helpful error messages

### DIFF
--- a/deployment/infrastructure/bedrock-auth-auth0.yaml
+++ b/deployment/infrastructure/bedrock-auth-auth0.yaml
@@ -112,6 +112,7 @@ Resources:
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
+              - !Sub 'arn:${AWS::Partition}:bedrock:*:*:application-inference-profile/*'
             Condition:
               StringEquals:
                 'aws:RequestedRegion': !Ref AllowedBedrockRegions
--- a/deployment/infrastructure/bedrock-auth-azure.yaml
+++ b/deployment/infrastructure/bedrock-auth-azure.yaml
@@ -104,6 +104,7 @@ Resources:
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
+              - !Sub 'arn:${AWS::Partition}:bedrock:*:*:application-inference-profile/*'
             Condition:
               StringEquals:
                 'aws:RequestedRegion': !Ref AllowedBedrockRegions
--- a/deployment/infrastructure/bedrock-auth-cognito-pool.yaml
+++ b/deployment/infrastructure/bedrock-auth-cognito-pool.yaml
@@ -110,6 +110,7 @@ Resources:
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
+              - !Sub 'arn:${AWS::Partition}:bedrock:*:*:application-inference-profile/*'
             Condition:
               StringEquals:
                 'aws:RequestedRegion': !Ref AllowedBedrockRegions
--- a/deployment/infrastructure/bedrock-auth-okta.yaml
+++ b/deployment/infrastructure/bedrock-auth-okta.yaml
@@ -103,6 +103,7 @@ Resources:
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
+              - !Sub 'arn:${AWS::Partition}:bedrock:*:*:application-inference-profile/*'
             Condition:
               StringEquals:
                 'aws:RequestedRegion': !Ref AllowedBedrockRegions
--- a/deployment/infrastructure/cognito-identity-pool.yaml
+++ b/deployment/infrastructure/cognito-identity-pool.yaml
@@ -143,6 +143,7 @@ Resources:
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
+              - !Sub 'arn:${AWS::Partition}:bedrock:*:*:application-inference-profile/*'
             Condition:
               StringEquals:
                 'aws:RequestedRegion': !Ref AllowedBedrockRegions
--- a/source/claude_code_with_bedrock/cli/commands/context.py
+++ b/source/claude_code_with_bedrock/cli/commands/context.py
@@ -199,6 +199,10 @@ class ContextShowCommand(Command):
             console.print("\n[bold cyan]Bedrock Configuration:[/bold cyan]")
             if profile.selected_model:
                 console.print(f"  Selected Model:       {profile.selected_model}")
+            for tier, attr in [("Opus", "inference_profile_opus_arn"), ("Sonnet", "inference_profile_sonnet_arn"), ("Haiku", "inference_profile_haiku_arn")]:
+                arn = getattr(profile, attr, None)
+                if arn:
+                    console.print(f"  {tier} Inference Profile: {arn}")
             if profile.selected_source_region:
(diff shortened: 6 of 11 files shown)


wirjo:
@juengler This may overlap with: https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/207. Let me DM you

dineshSajwan:
### Code review

Found 4 issues:

1. **`ANTHROPIC_MODEL` is removed with no backward-compatibility shim.** The previous code set `settings["env"]["ANTHROPIC_MODEL"] = profile.selected_model`. This line is gone entirely — only the new `ANTHROPIC_DEFAULT_*_MODEL` variables are written. Any existing Claude Code installation whose `settings.json` was generated before this change will lose the `ANTHROPIC_MODEL` value on next `ccwb package` run, causing Claude Code to fall back to its built-in default model rather than the admin-selected one.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/72529a3c300b97ee3ac33da803b6645abe564000/source/claude_code_with_bedrock/cli/commands/package.py#L2357-L2368

2. **When the user selects Sonnet (or Haiku), `ANTHROPIC_DEFAULT_OPUS_MODEL` is only set if an inference profile ARN is explicitly configured — otherwise it is left unset entirely.** In the `elif "sonnet"` branch, line 2378 gates the Opus env var on `if opus_arn:`. With no ARN, Claude Code gets no `ANTHROPIC_DEFAULT_OPUS_MODEL` and may route Opus requests to a model outside the admin's configured CRIS region. The Opus branch (line 2371) avoids this by always falling back to `model_id` — the same fallback pattern should apply to derived tiers in the Sonnet and Haiku branches.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/72529a3c300b97ee3ac33da803b6645abe564000/source/claude_code_with_bedrock/cli/commands/package.py#L2374-L2386

3. **The ARN validator regex rejects underscores in the profile ID segment, but AWS allows them.** The pattern `[a-zA-Z0-9-]+` at line 343 of `validators.py` will reject a valid ARN like `arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/my_profile_v1`. The AWS `create-inference-profile` naming convention permits underscores. Fix: change to `[a-zA-Z0-9_-]+`.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/72529a3c300b97ee3ac33da803b6645abe564000/source/claude_code_with_bedrock/validators.py#L341-L348

4. **Hardcoded derived model IDs (`claude-sonnet-4-6`, `claude-haiku-4-5-20251001-v1:0`) will produce wrong or unavailable models for EU and APAC users.** The prefix is extracted correctly (`us`/`eu`/`apac`) but the model family strings are hardcoded and do not account for regional model availability. Open PR #207 addresses this exact problem with a centralised `resolve_model_for_tier()` lookup. Merging this PR first will re-introduce the regional availability gap that PR #207 was created to fix.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/72529a3c300b97ee3ac33da803b6645abe564000/source/claude_code_with_bedrock/cli/commands/package.py#L2369-L2386

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

juengler:
Thanks @dineshSajwan @wirjo for the review and for flagging the overlap with #207.

Reworked the PR to build on top of #207's centralized model catalog and addressed all four issues from the automated review. The scope is now purely Application Inference Profile support:

- ANTHROPIC_MODEL preserved with ARN override when a tier is configured
- All three ANTHROPIC_DEFAULT_*_MODEL tiers always set via resolve_model_for_tier(), ARNs layer on top
- Validator regex now allows underscores in profile IDs
- No more hardcoded derived model IDs
Let me know if anything else needs adjusting.

dineshSajwan:
Thanks @juengler for the PR. 
One of the feedback we have got from our customers is that our documentation is not detailed and not updated. This PR has significant changes in the workflow. can you please update the workflow [here ](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/QUICK_START.md#complete-wizard-flow--decision-tree).

 In my view you should update below items , if there is anything else please update the documentation. 
 
 1.STEP 3: Bedrock Model Selection and add details for configuring application profile.
 2.  Also explain what will be the advantage customer will get with application profile and suggest when to use it.
 3. any prerequisite for using application inference profile ? How do we need to create it , we should suggest both console and aws cli for that.
 4. one example how the the dashboard will look when you use application inference profile for a team/department.
 5. if any cloudformation resource is added with this then update the all the document referring it like cost attribution md file, architecture diagram. any update in `ccwb deploy` command.

dineshSajwan (approved):

PR #273, opened 2026-04-17
fix: Honor daily_enforcement_mode independently from monthly in quota check Lambda

### LINKED ISSUE #272: Daily quota enforcement mode is not working as expected when configured to `alert`
There is a bug with the daily quota enforcement when set to alert. The current behaviour is block no matter what value we pass during the configuration.

jphuynh:
## Summary

Fixes the quota check Lambda to honor `DAILY_ENFORCEMENT_MODE` independently from `MONTHLY_ENFORCEMENT_MODE`. Previously, `DAILY_ENFORCEMENT_MODE` was passed as an env var by CloudFormation but never read by the Lambda — daily limits always used the monthly enforcement mode.

## Bug

When daily enforcement mode is set to `block` but monthly is `alert`, exceeding the daily limit does NOT block the user. The Lambda only checked `enforcement_mode` (monthly) and ignored the daily-specific setting entirely.

**Customer impact:** Users at 191% of daily limit continue to receive credentials when they should be blocked.

## Fix

- Read `DAILY_ENFORCEMENT_MODE` env var at module level (env-var path)
- Include `daily_enforcement_mode` in the `get_policy()` return dict (DynamoDB path)
- Use the daily-specific mode when evaluating daily limit exceeded

### Daily Enforcement Tests (8)
- `TestDailyEnforcementEnvVarPath` — validates daily block/alert modes via env vars
- `TestDailyEnforcementFineGrainedPath` — validates daily modes via DynamoDB policies

### Response Schema Contract Tests (11)
Ensures the Lambda's response format remains stable for `credential-process` consumers:
- `test_response_is_valid_json_with_status_code` — proper HTTP response structure
- `test_allowed_response_has_required_keys` — all keys present when allowed
- `test_blocked_response_has_required_keys` — all keys present when blocked
- `test_reason_field_is_valid_enum` — reason uses known values only
- `test_enforcement_mode_is_valid` — enforcement_mode is alert/block/None
- `test_usage_summary_structure` — usage has monthly_tokens, daily_tokens, limits, percent
- `test_policy_field_structure` — policy has type and identifier
- `test_unblock_status_structure` — unblock_status has is_unblocked boolean
- `test_message_field_is_string` — message is always a non-empty string
- `test_monthly_exceeded_sets_reason_correctly` — monthly limit returns correct reason
- `test_no_policy_response` — disabled quota returns no_policy

### Input Validation Contract Tests (4)
Ensures malformed requests produce structured responses instead of crashes:
- `test_missing_email_claim` — JWT without email returns structured error
- `test_missing_authorizer_context` — no authorizer context handled gracefully
- `test_empty_event` — completely empty event does not crash
- `test_missing_email_blocked_by_default` — fail-closed security default

## Changes

- `deployment/infrastructure/lambda-functions/quota_check/index.py` (+17/-12)
- `source/tests/test_quota_check_lambda.py` (+517 lines, new file)

### DIFF
--- a/deployment/infrastructure/lambda-functions/quota_check/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_check/index.py
@@ -325,6 +325,7 @@ def get_policy(policy_type: str, identifier: str) -> dict | None:
             "warning_threshold_80": int(item.get("warning_threshold_80", 0)),
             "warning_threshold_90": int(item.get("warning_threshold_90", 0)),
             "enforcement_mode": item.get("enforcement_mode", "alert"),
+            "daily_enforcement_mode": item.get("daily_enforcement_mode", "alert"),
             "enabled": item.get("enabled", True),
         }
     except Exception as e:
--- a/deployment/infrastructure/lambda-functions/quota_check/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_check/index.py
@@ -25,6 +25,7 @@ ENABLE_FINEGRAINED_QUOTAS = os.environ.get("ENABLE_FINEGRAINED_QUOTAS", "false")
 MONTHLY_TOKEN_LIMIT = int(os.environ.get("MONTHLY_TOKEN_LIMIT", "0"))
 DAILY_TOKEN_LIMIT = int(os.environ.get("DAILY_TOKEN_LIMIT", "0"))
 MONTHLY_ENFORCEMENT_MODE = os.environ.get("MONTHLY_ENFORCEMENT_MODE", "block")
+DAILY_ENFORCEMENT_MODE = os.environ.get("DAILY_ENFORCEMENT_MODE", "alert")
 WARNING_THRESHOLD_80 = int(os.environ.get("WARNING_THRESHOLD_80", "240000000"))
 WARNING_THRESHOLD_90 = int(os.environ.get("WARNING_THRESHOLD_90", "270000000"))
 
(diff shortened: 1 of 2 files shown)


deployment/infrastructure/lambda-functions/quota_check/index.py
```diff
@@ -153,18 +154,20 @@ def lambda_handler(event, context):
...
-                "enforcement_mode": enforcement_mode,
-                "usage": usage_summary,
-                "policy": {
-                    "type": policy.get("policy_type"),
-                    "identifier": policy.get("identifier")
-                },
-                "unblock_status": {"is_unblocked": False},
-                "message": f"Daily quota exceeded: {int(daily_tokens):,} / {int(daily_limit):,} tokens ({daily_tokens/daily_limit*100:.1f}%). Quota resets at UTC midnight."
-            })
+            daily_mode = policy.get("daily_enforcement_mode", "alert")
```
scouturier, line 157: I believe this fix would work for the env-var path, but when ENABLE_FINEGRAINED_QUOTAS=true the policy dict comes from get_policy_from_dynamodb(), which doesn't include daily_enforcement_mode. That means policy.get("daily_enforcement_mode", "alert") at the new line 157 will always fall back to "alert" for per-user/group policies, making daily block mode silently broken for fine-grained quotas. 

To address this, we could add "daily_enforcement_mode": item.get("daily_enforcement_mode", "alert") to the returned dict in get_policy_from_dynamodb(). maybe?

jphuynh: Thanks for the review. My tests didn't include ENABLE_FINEGRAINED_QUOTAS=true. I've pushed the fix for it including tests to make sure it's not missed anymore.

jphuynh:
```
============================= test session starts ==============================
platform darwin -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/jp/github/guidance-for-claude-code-with-amazon-bedrock/source
configfile: pyproject.toml
collected 8 items

tests/test_quota_check_lambda.py::TestDailyEnforcementEnvVarPath::test_daily_block_mode_blocks_when_exceeded PASSED [ 12%]
tests/test_quota_check_lambda.py::TestDailyEnforcementEnvVarPath::test_daily_alert_mode_allows_when_exceeded PASSED [ 25%]
tests/test_quota_check_lambda.py::TestDailyEnforcementEnvVarPath::test_daily_block_mode_allows_under_limit PASSED [ 37%]
tests/test_quota_check_lambda.py::TestDailyEnforcementFineGrainedPath::test_get_policy_returns_daily_enforcement_mode PASSED [ 50%]
tests/test_quota_check_lambda.py::TestDailyEnforcementFineGrainedPath::test_get_policy_defaults_daily_enforcement_mode_to_alert PASSED [ 62%]
tests/test_quota_check_lambda.py::TestDailyEnforcementFineGrainedPath::test_finegrained_daily_block_mode_blocks_when_exceeded PASSED [ 75%]
tests/test_quota_check_lambda.py::TestDailyEnforcementFineGrainedPath::test_finegrained_daily_alert_mode_allows_when_exceeded PASSED [ 87%]
tests/test_quota_check_lambda.py::TestDailyEnforcementFineGrainedPath::test_finegrained_missing_daily_mode_defaults_to_alert PASSED [100%]

============================== 8 passed in 5.21s ===============================
```

wirjo:
Rebased on current main (clean, no conflicts). This is a confirmed critical bug affecting a customer deployment — daily limit enforcement is completely non-functional:

- `DAILY_ENFORCEMENT_MODE` env var is passed by CloudFormation but never read by the Lambda
- Result: daily limits always use the monthly enforcement mode, which may be "alert" even when daily is configured as "block"
- Users at 191% of daily limit are not being blocked

Customer evidence shows the issue is live:
```
Daily Tokens    16.0M    8.3M    191.6%
```
User continues to receive credentials despite exceeding limit with `enforcement: block`.

@dineshSajwan this is a high-priority fix — one file, +17/-12 lines, with 286 lines of tests. Ready to merge.

jamesbland123 (approved):
Approved

PR #276, opened 2026-04-17
feat: add Claude Opus 4.7 model support with 1M context variant

parkerbradshaw88:
## Summary                                                                                                                                                                   
Adds Claude Opus 4.7 (200K) and Opus 4.7 (1M context) as selectable models, with full cross-region inference support across US, EU, AU, JP, and Global profiles.             
                                                                                                                                                                               
  ## Changes                                                                                                                                                                   
  - Registered `claude-opus-4-7` (200K) and `claude-opus-4-7[1m]` (1M) in the model catalog                                                                                    
  - Added CRIS inference profiles for US, EU, AU, JP, and Global                                                                                                               
  - Added analytics pricing entries for both variants                                                                                                                          
  - Updated dashboard widgets to surface usage/cost for the new models                                                                                                         
  - Updated Lambda telemetry handlers to tag the new model IDs                                                                                                                 

  ## Test plan
  - [ ] New models appear in the model selector
  - [ ] Invocations succeed against each inference profile (US, EU, AU, JP, Global)                                                                                            
  - [ ] Analytics pricing resolves for both 200K and 1M variants                                                                                                               
  - [ ] Dashboard widgets render usage/cost for the new models                                                                                                                 
  - [ ] Lambda telemetry emits correct model IDs


wirjo:
Thanks for adding Opus 4.7 support! Quick question on the `opus-4-7-1m` variant — the [Opus 4.7 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html) shows 1M context as native to the base model (`anthropic.claude-opus-4-7`), unlike Opus 4.6 where 200K was the default and `[1m]` was a separate opt-in variant. Can you verify that `anthropic.claude-opus-4-7[1m]` exists as a distinct Bedrock inference profile? If not, the `opus-4-7-1m` entry and its associated CRIS profiles/dashboard/Lambda mappings may not be needed.

wirjo:
Also heads-up: #278 adds `opus-4-7` to `models.py` as well (US + Global profiles) alongside the model alias system. The two PRs will conflict there — suggest rebasing on top of #278 once it merges to avoid duplication.

parkerbradshaw88:
Thanks for the careful review, @wirjo — you're right on both counts.

**Re: `opus-4-7-1m`** — confirmed. Opus 4.7 serves 1M context natively on `anthropic.claude-opus-4-7`, so `[1m]` isn't a distinct inference profile and the separate entry is redundant. Just pushed 6512a96 removing the `opus-4-7-1m` catalog entry and its test assertions. Full-repo grep for `opus-4-7-1m`, `claude-opus-4-7[1m]`, and "Opus 4.7 (1M" now returns zero matches. The base `opus-4-7` entry (which serves 1M natively) stays.

**Re: #278 conflict** — took a look at the current #278 diff: it adds the `get_claude_code_alias` helper + model_alias profile field, but doesn't add `opus-4-7` to `CLAUDE_MODELS` (its new test at line 149 asserts `us.anthropic.claude-opus-4-7` resolves, which depends on this PR landing). So the actual overlap looks minimal now — happy to rebase once either PR merges, whichever order you prefer.

wirjo:
Closing — Opus 4.7 is already on main (added via #207 merge). The 1M context variant does not exist as a separate Bedrock model (Opus 4.7 ships with 1M natively).

PR #280, opened 2026-04-20
fix: EnableFinegrainedQuotas should be set based on profile setting

### LINKED ISSUE #279: EnableFinegrainedQuotas is always set to true regardless of profile setting
[Profile.enable_finegrained_quotas defaults to False](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/458d1dc6c62ffe1880df6e74233a9a17348bb4e1/source/claude_code_with_bedrock/config.py#L61), but the [quota-monitoring.yaml CloudFormation template parameter has Default: 'true'](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/bca686a30194998123d59d916f47d2adc56ebc1c/deployment/infrastructure/quota-monitoring.yaml#L55) and deploy quota never pass the parameter explicitly.

This means fine-grained quotas are always enabled for every deployment regardless of what was configured in the profile.

Introduced in #74.

jphuynh:
The quota-monitoring CF template had Default: 'true' for EnableFinegrainedQuotas, and deploy.py never passed the parameter explicitly, so fine-grained quotas were unconditionally enabled.

Changed the template default to 'false' to keep the opt-in intent of profile.enable_finegrained_quotas and now pass the value explicitly so the profile is always the source of truth.

Closes: https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/issues/279

### DIFF
--- a/assets/docs/QUOTA_MONITORING.md
+++ b/assets/docs/QUOTA_MONITORING.md
@@ -50,7 +50,7 @@ Deploy using `poetry run ccwb deploy` (deploys all enabled stacks) or `poetry ru
 | Critical Threshold      | 90% (202.5M)| Second alert level                             |
 | Check Frequency         | 15 minutes  | Lambda execution interval                      |
 | Alert Retention         | 60 days     | DynamoDB TTL for deduplication                 |
-| EnableFinegrainedQuotas | true        | Enable fine-grained policy support             |
+| EnableFinegrainedQuotas | false       | Enable fine-grained policy support             |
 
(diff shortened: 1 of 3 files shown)


dineshSajwan (approved):
Approving it but Can you please also update `QUOTA_MONITORING.md` to set `EnableFinegrainedQuotas` deafult as false for updated documentation.

jphuynh:
Good catch. Thanks @dineshSajwan it's been updated.

PR #284, opened 2026-04-21
fix: preserve sso_enabled when pre-populating existing profile in ccwb init

### LINKED ISSUE #283: fix: sso_enabled not pre-populated when updating existing profile via ccwb init
## Bug

When running `ccwb init` and selecting an existing profile to update, the **SSO authentication** field always defaults to `True` regardless of the saved value.

## Root Cause

`_check_existing_deployment()` in `cli/commands/init.py` builds an `existing_config` dict from the saved profile but never includes the `sso_enabled` field. The wizard prompt then falls back to the hardcoded default:

```python
default=config.get("sso_enabled", True),  # always True
```

The `sso_enabled` field was introduced in PR #71 but was not added to the existing config pre-population logic.

## Fix

Add `sso_enabled` to the fields copied from the profile in `_check_existing_deployment()`.

scouturier:
## Summary

- Fixes #283
- PR #71 (merged 2026-04-20) introduced `sso_enabled` but left three gaps, all fixed here:
  - `_check_existing_deployment()` never included `sso_enabled` in the returned dict, so the wizard always defaulted to `True` when pre-populating an existing profile
  - The SSO toggle prompt was shown even when updating an existing profile that already had SSO configured — it now skips the prompt and carries the saved value forward silently
  - Old profile files (saved before PR #71) don't have `sso_enabled`; `Profile.from_dict()` now infers it from `provider_domain` so legacy profiles load correctly

## Test plan

- [ ] Run `ccwb init` and update an existing Cognito profile — SSO prompt should not appear
- [ ] Run `ccwb init` and update an existing IAM-only profile (`sso_enabled: false`) — SSO prompt should not appear and SSO should remain disabled
- [ ] Run `ccwb init` on a legacy profile file without `sso_enabled` — should correctly infer `True` if provider domain is set
- [ ] Run `ccwb init` to create a brand new profile — SSO prompt should still appear

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -296,23 +296,27 @@ class InitCommand(Command):
             console.print("\n[bold blue]Step 1: Authentication Configuration[/bold blue]")
             console.print("─" * 40)
 
-            console.print("\n[bold]SSO Authentication[/bold]")
-            console.print("Enable Single Sign-On authentication via identity providers")
-            console.print("(Okta, Auth0, Azure AD, AWS Cognito)")
-            console.print("\nWhen disabled:")
-            console.print("  • Uses AWS IAM roles for access control")
-            console.print("  • Metrics will use anonymous tracking based on IAM identity")
-            console.print("  • No user authentication required\n")
+            # Skip the SSO toggle when updating an existing profile — preserve the saved value
+            if existing_config and "sso_enabled" in existing_config:
+                config["sso_enabled"] = existing_config["sso_enabled"]
+            else:
+                console.print("\n[bold]SSO Authentication[/bold]")
+                console.print("Enable Single Sign-On authentication via identity providers")
+                console.print("(Okta, Auth0, Azure AD, AWS Cognito)")
+                console.print("\nWhen disabled:")
+                console.print("  • Uses AWS IAM roles for access control")
+                console.print("  • Metrics will use anonymous tracking based on IAM identity")
+                console.print("  • No user authentication required\n")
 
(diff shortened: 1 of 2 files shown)


source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1849,6 +1853,10 @@ def _check_existing_deployment(self, profile_name: str) -> dict[str, Any]:
                 },
             }
 
+            # Preserve sso_enabled flag (added in PR #71; missing here caused it to always default to True)
+            if hasattr(profile, "sso_enabled"):
```
dineshSajwan, line 2316: Isn't this still always true ? `sso_enabled` has default value in `config.py` .

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -296,23 +296,27 @@ def _gather_configuration(self, progress: WizardProgress, existing_config: dict[
...
-            console.print("  • Uses AWS IAM roles for access control")
-            console.print("  • Metrics will use anonymous tracking based on IAM identity")
-            console.print("  • No user authentication required\n")
-
-            sso_enabled = questionary.confirm(
-                "Enable SSO authentication?",
-                default=config.get("sso_enabled", True),
-            ).ask()
+            # Skip the SSO toggle when updating an existing profile — preserve the saved value
+            if existing_config and "sso_enabled" in existing_config:
```
dineshSajwan, line 321: if any existing_config want to disable SSO with re-init, how can they do it . This looks of it will be always sso_enabled once set. Thoughts ?

dineshSajwan (commented):
Thank you for the PR, Looks good overall, have some question , please check.

PR #291, opened 2026-04-23
fix: remove hard AWS CLI dependency from installer

wirjo:
Closes #217

## Problem

The installer exits with an error if the AWS CLI is not installed. For organisations with large developer populations (e.g. embedded systems engineers), requiring the AWS CLI purely for an AI coding assistant creates unnecessary friction.

## Fix

One file changed (`package.py`), three targeted edits:

1. **install.sh** — AWS CLI check becomes informational instead of `exit 1`
2. **install.bat** — same for Windows
3. **settings.json** — adds `AWS_CREDENTIAL_PROCESS` env var pointing to the credential binary

## How it works

The credential-process binary calls AWS APIs directly via its bundled SDK — it never needed the AWS CLI. Adding `AWS_CREDENTIAL_PROCESS` to `settings.json` gives the AWS SDK for JavaScript (used by Claude Code) a direct path to invoke the binary, without needing `~/.aws/config` or the CLI.

## Backwards compatible

- Existing installs with AWS CLI: **no change** — `AWS_PROFILE` + `~/.aws/config` still works as before
- The installer still writes `~/.aws/config` (via `cat >>`, not `aws configure`) for users with other AWS tooling
- `AWS_CREDENTIAL_PROCESS` is additive — if `AWS_PROFILE` resolves first, the SDK uses that

Based on the customer-submitted patch attached to #217.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -1807,10 +1807,10 @@ echo
 # Check prerequisites
 echo "Checking prerequisites..."
 
-if ! command -v aws &> /dev/null; then
-    echo "❌ AWS CLI is not installed"
-    echo "   Please install from https://aws.amazon.com/cli/"
-    exit 1
+if command -v aws &> /dev/null; then
+    echo "✓ AWS CLI found (optional)"
+else
+    echo "ℹ  AWS CLI not found — not required. The credential process binary handles authentication directly."
 fi
 
 echo "✓ Prerequisites found"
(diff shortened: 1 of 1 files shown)


dineshSajwan (approved):

dineshSajwan:
Approved it but before merging I think we need to update aws cli check in [init.py](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/82c86c5b68f8c1776f3a302623a3d78595c10ad8/source/claude_code_with_bedrock/cli/commands/init.py#L233) as well . 

```
checks = {
      "AWS CLI installed": self._check_aws_cli(),    #  this may fail when no aws cli
      "AWS credentials configured": self._check_aws_credentials(),
      
  }
```

PR #302, opened 2026-04-28
fix(otel): remove unconditional HTTPSListener DependsOn

### LINKED ISSUE #286: OTEL collector CloudFormation fails without custom domain — unresolved HTTPSListener dependency
## Description

Deploying the monitoring stack without a custom domain (HTTPS disabled) fails with:

```
Template format error: Unresolved resource dependencies [HTTPSListener] in the Resources block of the template
```

## Steps to Reproduce

1. Run `ccwb init`, choose **No** for "Enable HTTPS with custom domain?"
2. Run `ccwb deploy`
3. Monitoring stack (`otel-collector.yaml`) fails CloudFormation validation

## Root Cause

In `deployment/infrastructure/otel-collector.yaml` (~line 490-495), the `ECSService` resource has:

```yaml
DependsOn:
  - HTTPListener
  - HTTPSListener
```

But `HTTPSListener` has `Condition: HasCustomDomain`. When no custom domain is configured, `HTTPSListener` doesn't exist, and CloudFormation cannot resolve the dependency.

## Expected Behavior

Deployment should succeed without a custom domain configured.

## Environment

- Version: 2.3.0
- AWS Region: us-east-1

jwhitcraft:
## Problem

Deploying the monitoring stack **without a custom domain** fails CloudFormation template validation with:

```
Template format error: Unresolved resource dependencies [HTTPSListener] in the Resources block of the template
```

### Root cause

`ECSService` had an unconditional `DependsOn` block that included `HTTPSListener`:

```yaml
DependsOn:
  - HTTPListener
  - HTTPSListener   # <-- problem
```

`HTTPSListener` carries `Condition: HasCustomDomain`, so CloudFormation only creates it when a custom domain is configured. When no custom domain is provided, the resource does not exist but the `DependsOn` reference still resolves — causing a hard validation failure before any stack operation begins.

## Fix

Removed `HTTPSListener` from the `ECSService` `DependsOn` list. `HTTPListener` alone is sufficient to correctly order ECS service creation relative to the load balancer listeners.

## Testing

1. Deploy the monitoring/OTEL stack **without** setting a custom domain.
2. Confirm CloudFormation template validation passes and the stack reaches `CREATE_COMPLETE` (previously it would fail at the validation stage with the unresolved dependency error).
3. Optionally re-deploy **with** a custom domain to confirm the HTTPS path is unaffected.

Closes #286

### DIFF
--- a/deployment/infrastructure/otel-collector.yaml
+++ b/deployment/infrastructure/otel-collector.yaml
@@ -492,7 +492,6 @@ Resources:
     Type: AWS::ECS::Service
     DependsOn:
       - HTTPListener
-      - HTTPSListener
     Properties:
       ServiceName: otel-collector-service
       Cluster: !Ref ECSCluster


wirjo:
This is already fixed by PR #311 (merged), which split ECSService into two conditional variants (ECSServiceHTTP / ECSServiceHTTPS). Closing as duplicate.

PR #282, opened 2026-04-21
feat: add ccwb cowork generate command and CoWork 3P documentation

wirjo:
## Summary

Add support for generating Claude Cowork 3P MDM configuration files and a dedicated CoWork usage dashboard, enabling organizations to deploy Claude Desktop with Amazon Bedrock using the same credential helper infrastructure.

### `ccwb cowork generate` command

Standalone CLI that reads the deployment profile and generates MDM config files (JSON, macOS .mobileconfig, Windows .reg). Uses full CRIS model IDs derived from `ccwb init` selection.

### `ccwb package` integration

CoWork 3P configs auto-generated during `ccwb package` (enabled by default). Same shared utilities as standalone command.

### Model alignment with `ccwb init`

`inferenceModels` uses full CRIS model IDs (e.g., `us.anthropic.claude-opus-4-7`) from the admin's selected model and cross-region profile. One model per tier (opus/sonnet/haiku), newest version, selected first. New `cowork_aliases` field on `ClaudeModel` + `get_cowork_inference_models()` helper in `models.py`.

### CoWork Usage Dashboard

`ccwb deploy cowork-dashboard` — dedicated CloudWatch dashboard (16 widgets):
- **Summary:** Total Sessions, Tokens, Cost (USD), Active Users
- **Time series:** Token/Cost/Sessions Over Time, Token Usage by Model
- **User activity:** Top Users by Tokens, Top Users by Cost, Per-User Summary
- **Breakdowns:** Token types (input/output/cache), Token/Cost by Model
- **Platform:** Sessions by OS, CoWork Versions, Architecture

Queries use `service.name = "cowork"` against the shared metrics log group — no additional infrastructure.

### Documentation

`assets/docs/COWORK_3P.md` — Getting Started workflows (new vs existing deployments), 3 config options (CLI/Setup UI/JSON MDM), MDM key reference tables, deployment steps (macOS/Windows/VDI), monitoring differences, architecture diagram.

## Files Changed

**New:** `cowork.py`, `cowork_3p.py`, `cowork-dashboard.yaml`, `COWORK_3P.md`, 3 images

**Modified:** `models.py`, `config.py`, `init.py`, `package.py`, `deploy.py`, `CLI_REFERENCE.md`, `README.md`, `docs/README.md`

## Security

- XML escaping (`xml.sax.saxutils.escape`) for .mobileconfig
- Registry JSON array values properly escape inner double-quotes
- `isDesktopExtensionSignatureRequired` defaults to `True`
[...]

PR #139, opened 2026-03-05
Enhance Windows installer script to copy Claude Code settings and replace placeholders

Oleh-Tkachuk-Nice:
*Description of changes: Fixed the Windows installation script, which previously failed to run.*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #307, opened 2026-04-29
fix: Windows install.bat fails with '-replace is incorrectly used' error

wirjo:
Fixes #226. Supersedes #139 (rebased on current main with same approach).

## Problem

The generated `install.bat` fails on Windows with:
```
-replace の使い方が誤っています。
```

**Root cause 1:** Multi-line PowerShell command uses `^` line continuation inside double-quoted strings. `cmd.exe` does not treat `^` as continuation inside quotes, so `-replace` is executed as a standalone command.

**Root cause 2:** `!PROFILE_REGION!` uses delayed expansion syntax without `SETLOCAL ENABLEDELAYEDEXPANSION`, so the region is set to the literal string.

## Fix

1. Add `SETLOCAL ENABLEDELAYEDEXPANSION` + `ENDLOCAL`
2. Replace multi-line PowerShell with single-line using `$env:USERPROFILE` and `.Replace()` 
3. Fix `SKIP_SETTINGS` and `OVERWRITE` to use `!var!` delayed expansion
4. Use `powershell -NoProfile` for faster execution

One file, +10/-15 lines.

## Credit

Based on the approach from PR #139 by @Oleh-Tkachuk-Nice, rebased on current main after recent changes (#291, #282, #299).

Co-authored-by: Oleh-Tkachuk-Nice <Oleh-Tkachuk-Nice@users.noreply.github.com>

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -2034,6 +2034,7 @@ echo
         """Create Windows batch installer script."""
 
         installer_content = f"""@echo off
+SETLOCAL ENABLEDELAYEDEXPANSION
 REM Claude Code Authentication Installer for Windows
 REM Organization: {profile.provider_domain}
 REM Generated: {datetime.now().strftime("%Y-%m-%d %H:%M:%S")}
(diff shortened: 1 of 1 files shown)


wirjo:
Closing in favour of #318 which includes a more comprehensive Windows installer fix.

PR #309, opened 2026-04-29
fix: ccwb init fails when AWS session env vars are expired

### LINKED ISSUE #308: bug: ccwb init fails when AWS_* env vars are expired or aws login profile is not set
## Problem

`ccwb init` fails with credential errors when users authenticate via `aws login` (IAM Identity Center) but:

1. **Expired `AWS_*` environment variables are present** — boto3 prioritises env vars over `~/.aws/credentials` and fails with expired tokens instead of falling back to the credentials file.

2. **`AWS_PROFILE` is not set** — `aws login` authenticates the user and writes session credentials to `~/.aws/sso/cache/`, but boto3 needs `AWS_PROFILE` to resolve which SSO profile to use. Without it, boto3 attempts the default credential chain and fails.

## Reproduction

```bash
# User logs in via IAM Identity Center
aws login

# Expired env vars from a previous session are still in the shell
env | grep AWS_
# AWS_ACCESS_KEY_ID=AKIA...  (expired)
# AWS_SESSION_TOKEN=...       (expired)

# ccwb init fails — boto3 uses expired env vars instead of SSO cache
cd guidance-for-claude-code-with-amazon-bedrock/source
poetry run ccwb init
# Error: ExpiredTokenException or InvalidClientTokenId
```

## Current Workaround

```bash
# Option 1: Unset expired env vars
unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN

# Option 2: Set the profile explicitly
export AWS_PROFILE=CorpIT-Prod

# Then re-run
poetry run ccwb init
```

## Proposed Fix

In `ccwb init` (and `ccwb deploy`), add an AWS credential validation check early in the flow that:

1. **Detects expired env vars** — if `AWS_ACCESS_KEY_ID` is set, attempt an STS `get-caller-identity` call. If it fails with expired token, warn the user and suggest unsetting them.

2. **Detects missing AWS_PROFILE when SSO cache exists** — if no credentials resolve but `~/.aws/sso/cache/` has recent tokens, suggest setting `AWS_PROFILE`.

3. **Provide clear error message** — instead of a generic boto3 error, show:
   ```
   ⚠️  AWS credentials expired. Your shell has AWS_ACCESS_KEY_ID set but the token has expired.
   
   Fix: unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
   Or:  export AWS_PROFILE=<your-profile-name>
   
   Then re-run: poetry run ccwb init
   ```

## Context

This is related to PR #306 which adds IAM Identity Center as a first-class auth option. Users choosing IDC will authenticate via `aws login` / `aws sso login` and are more likely to hit this issue. The fix should be part of the prerequisites check in `ccwb init`.

## Impact

Users with IAM Identity Center or any SSO-based AWS access cannot run `ccwb init` without manually managing environment variables first. This is a common enterprise setup.

wirjo:
Fixes #308

## Problem

`ccwb init` fails at the prerequisites check when expired `AWS_ACCESS_KEY_ID` / `AWS_SESSION_TOKEN` environment variables are present. boto3 prioritises env vars over `~/.aws/credentials` and SSO cache — if they are expired, it fails without falling through to other credential sources.

This commonly occurs when admins authenticate via `aws sso login` but have stale env vars from a previous session in their shell.

## Fix

`_check_aws_credentials()` now retries without env vars if the first attempt fails:

1. Try `sts:GetCallerIdentity` with current credential chain
2. If that fails AND `AWS_*` env vars are set, temporarily clear them
3. Retry — if it succeeds, the env vars were expired; leave them cleared and warn
4. If both fail, restore env vars and report failure

No manual `unset AWS_ACCESS_KEY_ID` required — `ccwb init` auto-handles it.

## Impact

One file, +28/-1 lines. Only affects the prerequisites check in `ccwb init`. No behaviour change when credentials are valid.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1668,24 +1668,59 @@ class InitCommand(Command):
             return False
 
     def _check_aws_credentials(self) -> bool:
-        """Check if AWS credentials are configured."""
+        """Check if AWS credentials are configured.
+
+        If AWS_ACCESS_KEY_ID env vars are set but expired, boto3 won't
+        fall through to ~/.aws/credentials or SSO cache. We detect
+        credential-specific errors and retry without env vars.
+        """
+        import os
+
         console = Console()
         try:
             boto3.client("sts").get_caller_identity()
             return True
         except Exception as e:
             err = str(e)
-            console.print(f"    [dim red]Credential error: {err}[/dim red]")
+
+            # Only retry for credential-specific errors (not network/throttling)
+            credential_errors = ["ExpiredToken", "InvalidClientTokenId", "SignatureDoesNotMatch",
+                                 "expired", "ExpiredTokenException"]
+            is_credential_error = any(ce in err for ce in credential_errors)
+
+            # If credential error and env vars are set, retry without them
+            if is_credential_error:
+                env_keys = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]
+                env_backup = {k: os.environ.pop(k) for k in env_keys if k in os.environ}
+
+                if env_backup:
+                    try:
+                        boto3.client("sts").get_caller_identity()
+                        console.print(
+                            "    [yellow]\u26a0[/yellow] [dim]Expired AWS session env vars detected and ignored[/dim]"
+                        )
+                        return True
+                    except Exception:
+                        os.environ.update(env_backup)  # Restore on failure
+
+            # Show error-specific guidance
+            console.print(f"    [dim red]Credential error: {err[:120]}[/dim red]")
             if "ExpiredToken" in err or "expired" in err.lower():
                 console.print(
-                    "    [dim]Hint: Expired credentials in ~/.aws/credentials are blocking the EC2 instance role.\n"
-                    "    Run: [cyan]unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN[/cyan]\n"
-                    "    Then clear the [default] section from ~/.aws/credentials and retry.[/dim]"
+                    "    [dim]Hint: Your session has expired. Try:[/dim]\n"
+                    "    [dim]  \u2022 aws sso login --profile <profile>[/dim]\n"
+                    "    [dim]  \u2022 export AWS_PROFILE=<your-profile>[/dim]"
                 )
             elif "NoCredentialProviders" in err or "Unable to locate credentials" in err:
                 console.print(
-                    "    [dim]Hint: No credentials found. Configure via env vars, ~/.aws/credentials,\n"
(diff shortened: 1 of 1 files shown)


source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1619,11 +1628,38 @@ def _check_aws_cli(self) -> bool:
...
+        and retry without env vars.
+        """
+        import os
+
+        try:
+            boto3.client("sts").get_caller_identity()
+            return True
+        except Exception:
+            pass
+
```
scouturier, line 1685: if the first get_caller_identity() raises a non-credential error (network timeout, service unavailable, throttling), the function silently retries without env vars; should only retry on credential-specific errors (ExpiredToken, InvalidClientTokenId, SignatureDoesNotMatch)

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1619,11 +1628,38 @@ def _check_aws_cli(self) -> bool:
...
+        env_keys = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"]
+        env_backup = {k: os.environ.pop(k) for k in env_keys if k in os.environ}
+
+        if not env_backup:
+            return False  # No env vars to clear — genuinely no credentials
+
         try:
             boto3.client("sts").get_caller_identity()
+            # Success without env vars
+            console = Console()
```
scouturier: instantiating Console() inside the function when the caller likely already has one; consider passing console as a parameter or using a module-level instance for consistency with the rest of init.py

source/claude_code_with_bedrock/cli/commands/init.py
```diff
@@ -1619,11 +1628,38 @@ def _check_aws_cli(self) -> bool:
...
+            console.print(
+                "  [yellow]\u26a0[/yellow] Expired AWS session env vars detected and ignored"
+            )
             return True
         except Exception:
+            # Restore env vars — both paths failed
+            os.environ.update(env_backup)
             return False
 
     def _check_python_version(self) -> bool:
```
scouturier, line 1727: the PR removes error-specific messaging (ExpiredToken hints, NoCredentialProviders hints) that was added by PR #311 ; after rebase, consider preserving those detailed error messages while adding the auto-clear behavior

wirjo:
Rebased on current main (post-#311) and addressed all review feedback:

1. Only retries for credential-specific errors (ExpiredToken, InvalidClientTokenId, SignatureDoesNotMatch) — network timeouts and throttling pass through without retry
2. Uses Console() already instantiated in the method (consistent with #311's pattern)
3. Preserves and improves error-specific messaging from #311

@dineshSajwan ready for merge when you get a chance 👍

wirjo:
Closing — this fix is already in the beta branch. The init wizard now detects expired AWS tokens and prints clear remediation instructions including the `unset` command.

PR #315, opened 2026-04-30
Add bedrock:CallWithBearerToken to IAM policies for Cowork credential-helper support

ghitaelmoussi:
## Problem

Claude Cowork (Desktop) using the `inferenceCredentialHelper` auth method fails with a 403 error because the IAM policies in the CloudFormation templates don't include the `bedrock:CallWithBearerToken` action.

Claude Code (CLI) uses standard STS credentials to call `bedrock-runtime:Converse`. Cowork's credential-helper converts STS credentials into a bearer token, which hits a different Bedrock API path requiring [`bedrock:CallWithBearerToken`](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-permissions.html) — the same action used by [Bedrock API keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html).

This means after a standard `ccwb deploy`, Claude Code works but Cowork returns: *"not authorized to perform: bedrock:CallWithBearerToken"*.

## Changes

Added `bedrock:CallWithBearerToken` to the Bedrock IAM policy in all 5 authentication CloudFormation templates:

- `cfn/okta.yaml`
- `cfn/azure.yaml`
- `cfn/auth0.yaml`
- `cfn/cognito-user-pool.yaml`
- `cfn/cognito-identity-pool.yaml`

## Testing

Deployed with the updated template and verified both Claude Code (CLI) and Cowork (Desktop) work with the same stack.

### DIFF
--- a/deployment/infrastructure/bedrock-auth-auth0.yaml
+++ b/deployment/infrastructure/bedrock-auth-auth0.yaml
@@ -102,6 +102,7 @@ Resources:
             Action:
               - 'bedrock:InvokeModel'
               - 'bedrock:InvokeModelWithResponseStream'
+              - 'bedrock:CallWithBearerToken'
               - 'bedrock:ListFoundationModels'
               - 'bedrock:ListInferenceProfiles'
               - 'bedrock-runtime:InvokeModel'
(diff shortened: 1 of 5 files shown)


dineshSajwan (approved):
LGTM, Thank you for the contribution.

PR #317, opened 2026-04-30
chore(docs): add troubleshooting to QUICK_START.md

dineshSajwan:
End users on Intel Macs hit '[Errno 8] Exec format error' when the admin builds on Apple Silicon without setting up the x86_64 Python environment, producing only a macos-arm64 binary. ARM64 binaries cannot run on Intel (unlike Intel→ARM64 which Rosetta handles transparently).

Adds a dedicated troubleshooting entry with:
- Diagnostic commands to confirm the architecture mismatch
- Root cause explanation (ARM64-only build)
- Step-by-step fix for the admin to add Intel build support
- Note clarifying why Rosetta doesn't help in this direction

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -16,6 +16,19 @@ Complete deployment walkthrough for IT administrators deploying Claude Code with
 - AWS CLI v2
 - Git
 
+**macOS admins — check your machine architecture before building:**
+
+```bash
+uname -m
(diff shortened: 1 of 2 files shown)


jamesbland123 (approved):
Approved

PR #320, opened 2026-04-30
fix: resolve Windows native packaging bugs (issue #168)

### LINKED ISSUE #168: [Bugs, Solved] Can't build Windows binaries
Hi dear mainteiners,

I face issues during attempts to build Windows binaries, specifically `poetry run ccwb package`
gives the error:
Image
while the `poetry env info` -> Virtualenv -> Python: 3.12.10 version shows the correct one.
`where python`
C:\Python312\python.exe
`poetry add --group dev nuitka ordered-set zstandard`
The following packages are already present in the pyproject.toml and will be skipped:
  - nuitka
  - ordered-set
  - zstandard
If you want to update it to the latest compatible version, you can use `poetry update package`.
If you prefer to upgrade it to the latest available version, you can use `poetry add package@latest`._

Nothing to add.

I removed all other versions from my system, cleander the env variables, rebooted, and I only have that one. I restarted guides multiple times, with no progress.
We need specifically Windows version of the Claude Code.

Thank you.

scouturier:
## Summary

Fixes critical Windows native packaging bugs reported in #168. Enables users to build Windows binaries natively using Nuitka without requiring CodeBuild or cross-compilation.

## Attribution

Windows packaging fixes based on work by @pipeflo in PR #146. This PR extracts just the Windows-specific fixes and adds a critical bug fix for the `executable_path` initialization issue.

### 1. UnboundLocalError crash (Critical - not in PR #146)
**Issue:** When `_build_executable()` raises an exception, `executable_path` is never initialized, causing `UnboundLocalError` on line 327 when checking for Windows CodeBuild scenarios.

**Fix:** Initialize `executable_path = None` before the try block.

**Impact:** Prevents crash when Nuitka is missing or build fails for any reason.

### 2. Missing binary_name for Windows
**Issue:** Line 596 had a comment `# binary_name already set above` but it was never actually set, causing undefined variable errors.

**Fix:** Set `binary_name = "credential-process-windows.exe"` for Windows target.

**Impact:** Enables Windows binary generation.

### 3. Unix-only Nuitka detection
**Issue:** Used `which nuitka` command which doesn't exist on Windows, causing "Nuitka not found" errors even when Nuitka is installed.

**Fix:** Use cross-platform `python -m nuitka --version` instead.

**Impact:** Proper Nuitka detection on Windows.

### 4. OTEL helper Windows build path
**Issue:** No native Windows build path for otel-helper; always errored when monitoring was enabled.

**Fix:** Add native Windows build support in `_build_otel_helper()` and `_build_native_otel_helper()`.

**Impact:** Complete Windows packaging with monitoring support.

## Testing

- ✅ Code changes validated against the error in #168
- ✅ Logic verified for all four bug fixes
- ✅ Based on latest main (rebased)

Full end-to-end Windows testing requires a Windows machine with Nuitka installed.

## Files Changed

- `source/claude_code_with_bedrock/cli/commands/package.py` (+15/-5 lines)

Closes #168
Based on PR #146

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -310,6 +310,7 @@ class PackageCommand(Command):
         for platform_name in platforms_to_build:
             # Build credential process
             console.print(f"[cyan]Building credential process for {platform_name}...[/cyan]")
+            executable_path = None  # Initialize to prevent UnboundLocalError if build fails
             try:
                 executable_path = self._build_executable(output_dir, platform_name)
                 # Check if this was an async Windows build
@@ -593,7 +594,7 @@ class PackageCommand(Command):
             binary_name = "credential-process-linux"
         elif target_platform == "windows":
             platform_variant = "x86_64"
-            # binary_name already set above
+            binary_name = "credential-process-windows.exe"
         else:
             raise ValueError(f"Unsupported target platform: {target_platform}")
 
@@ -629,7 +630,7 @@ class PackageCommand(Command):
         # Check if Nuitka is available (through Poetry)
         source_dir = Path(__file__).parent.parent.parent.parent
         nuitka_check = subprocess.run(
-            ["poetry", "run", "which", "nuitka"], capture_output=True, text=True, cwd=source_dir
+            ["poetry", "run", "python", "-m", "nuitka", "--version"], capture_output=True, text=True, cwd=source_dir
         )
         if nuitka_check.returncode != 0:
             raise RuntimeError(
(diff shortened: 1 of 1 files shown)


scouturier:
Closing in favor of #429, which re-implements these Windows packaging fixes on the current `beta` branch (it explicitly carries the `executable_path` UnboundLocalError guard — referenced there as 'PR #320 bug 1' — plus the Nuitka availability/CodeBuild fallback and otel-helper Windows build path).

This PR targeted `main`, predates the Go rewrite (#338) and the move to `beta`, and is currently in a CONFLICTING state, so rebasing it would be more churn than value. The underlying issue #168 is addressed by #429. Closing to consolidate the Windows packaging work there.

PR #321, opened 2026-04-30
fix(package): auto-provision arch-matched Python for macOS cross-arch builds

scouturier:
## Summary

`ccwb package` could silently produce macOS binaries whose outer Mach-O header claimed the requested arch but whose **embedded libpython was a different arch**, so the binary would fail to launch on end-user Macs with errors like `mach-o file, but is an incompatible architecture`.

Root cause: PyInstaller's `--target-arch` only thins the universal2 bootloader — it can't change the interpreter arch of the Python driving PyInstaller. If Poetry's Python was thin x86_64 (common under Rosetta or with Homebrew's x86_64 Python), `poetry run pyinstaller --target-arch=arm64` produced a binary with arm64 shell + x86_64 libpython.

**Fix:** when Poetry's Python can't natively target the requested arch, auto-provision a thin arch-matched CPython 3.12 from [python-build-standalone](https://github.com/astral-sh/python-build-standalone) into `~/.ccwb/build-venvs/<arch>/` and drive PyInstaller under that interpreter. The legacy `~/venv-x86` convention is still honored for backwards compatibility.

Also removes dead/unreachable `macos-universal` code paths and updates docs (CLI_REFERENCE.md, QUICK_START.md) that referenced the obsolete manual `~/venv-x86` setup.

## Test plan

- [x] Reproduced the bug end-to-end on Apple Silicon using a thin x86_64 Python — confirmed outer arm64 / embedded x86_64 binary
- [x] After fix: `ccwb package --target-platform=macos-arm64` on arm64 Poetry → real arm64 binary (outer + embedded libpython)
- [x] After fix: `ccwb package --target-platform=macos-intel` on arm64 Poetry → auto-provisions `~/.ccwb/build-venvs/x86_64/`, produces real x86_64 binary
- [x] Legacy `~/venv-x86` fallback preserved (tested)
- [x] `poetry run pytest tests/cli/commands/test_package*.py` → 13/13 passing
- [x] Windows CodeBuild path validated end-to-end on a real Windows EC2 via SSM (separate CodeBuild flow, unchanged by this PR but sanity-checked)

## Follow-ups (not in this PR)

Sub-agent review surfaced real gaps worth tracking:
- Hash-verify the PBS tarball against the published `.sha256` (supply chain)
- Change `_warn_on_arch_mismatch` → hard-fail (warning in a long log is easily missed)
- Pin a verified PBS release tag + deliberate bumps (reproducibility + protects against PBS dropping 3.12 support)
- Add `~/.ccwb/build-venvs/` to `ccwb cleanup` (~100MB per arch, currently grows forever)
- No offline fallback: first build fails on airgapped networks

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -783,13 +783,9 @@ See [Distribution Comparison](assets/docs/distribution/comparison.md) for detail
   - If Docker is not installed or its daemon is not running, Linux builds are skipped with a warning
   - macOS and Windows builds have **no dependency on Docker**
 
-### Optional: Intel Mac Builds
+### Cross-Architecture macOS Builds
 
-Intel Mac builds require an x86_64 Python environment on Apple Silicon Macs.
-
-See [CLI Reference - Intel Mac Build Setup](assets/docs/CLI_REFERENCE.md#intel-mac-build-setup-optional) for setup instructions.
-
-If not configured, the package command will skip Intel builds and continue with other platforms.
+Both ARM64 and Intel macOS binaries build automatically — no manual setup. When the Poetry Python can't natively target the requested architecture, `ccwb package` auto-provisions a thin arch-matched Python into `~/.ccwb/build-venvs/<arch>/` (one-time ~15MB download per architecture).
 
 ---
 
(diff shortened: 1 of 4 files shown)


wirjo:
Closing — this was superseded by #324 which was merged and solves the same macOS cross-arch build problem (auto-provisioning per-arch venvs from universal2 Python). Thank you @scouturier!

PR #322, opened 2026-05-01
fix(package): refuse macOS builds from non-macOS hosts

scouturier:
## Summary

Fixes a bug where `ccwb package --target-platform=macos-arm64` (or `macos-intel`) run on a Linux host would silently produce a Linux ELF binary written to a macOS-named output file. End users on macOS would then hit `[Errno 8] Exec format error` when the AWS SDK invoked `credential-process` — the macOS kernel cannot load ELF, and Rosetta only translates Mach-O.

PyInstaller is a runtime bundler, not a cross-OS compiler. It emits binaries in the host OS's native format regardless of `--target-arch`. This PR adds a hard guard that refuses macOS targets on non-macOS hosts with a clear error listing three realistic alternatives (macOS workstation, CI macOS runner, drop macOS targets).

## What changes

**Code** (`source/claude_code_with_bedrock/cli/commands/package.py`):
- New module-level helper `_assert_host_os_can_build_macos()` called from `_build_macos_pyinstaller` and the macOS path of `_build_otel_helper_pyinstaller`
- On Darwin: no-op. On any other host: raises `RuntimeError` with remediation options
- Other build targets (linux, windows) in the same `ccwb package` invocation continue normally — only macOS targets are refused

**Docs** (`QUICK_START.md`, `assets/docs/CLI_REFERENCE.md`, `assets/docs/DEPLOYMENT.md`):
- Add a host-OS matrix stating explicitly that macOS binaries must be built on macOS
- Replace ambiguous wording about cross-compilation direction

## Verification

**Reproduced the original bug on Amazon Linux 2023 aarch64 (EC2):**
```
$ poetry run pyinstaller --target-arch=arm64 --name=credential-process-macos-arm64 ...
✓ build completes, no error
$ file /tmp/macbuild/credential-process-macos-arm64
credential-process-macos-arm64: ELF 64-bit LSB executable, ARM aarch64 ... GNU/Linux
```

**With the guard applied, same command path:**
```
$ poetry run python /tmp/test_guard.py
GUARD TRIGGERED — build refused as expected.
---
Cannot build macOS binaries on Linux.

PyInstaller cannot cross-compile across operating systems. On Linux,
it produces Linux-native binaries regardless of --target-arch.

Options for producing macOS binaries:
  1. Run `ccwb package` on a macOS workstation
  2. Use a CI macOS runner (GitHub Actions `macos-latest`, AWS
     CodeBuild macOS project, or a self-hosted Mac runner) and
     collect the artifacts from there
  3. Drop macOS targets from this build — run with
     --target-platform=linux-x64,linux-arm64,windows
     to build only the platforms your current host supports
```

## Test plan

- [x] Existing `test_package.py`, `test_package_async.py`, `test_package_models.py` — 13/13 pass unchanged
- [x] Manual reproduction on Amazon Linux 2023 (EC2) — guard triggers, no invalid binary written
- [x] Manual smoke on macOS — guard no-ops, normal macOS builds unaffected
- [ ] Maintainer: verify on a Linux CI runner that `--target-platform=all` (which on Linux auto-selects linux-x64, linux-arm64) continues to work unchanged — the guard only fires on explicit macOS target selection

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -669,10 +669,20 @@ Pick based on what your developers report:
 
 **Package Workflow:**
 
-1. **Local builds**: macOS/Linux executables are built locally using PyInstaller
+1. **Local builds**: macOS and Linux executables are built locally using PyInstaller. See the host-OS matrix below — PyInstaller cannot cross-compile across operating systems, so **macOS binaries must be built on macOS** and **Linux binaries must be built on Linux** (or on macOS via Docker).
 2. **Windows builds**: Trigger AWS CodeBuild for Windows executables (20+ minutes) - requires enabling CodeBuild during `init`
 3. **Check status**: Monitor build progress with `poetry run ccwb builds`
 4. **Create distribution**: Use `distribute` to upload and generate presigned URLs
+
+**Host-OS requirements for each target:**
+
+| Target binary | Build host must be | Tooling |
+|---|---|---|
+| `macos-arm64`, `macos-intel` | macOS | PyInstaller (native) |
+| `linux-x64`, `linux-arm64` | Linux, **or** macOS with Docker Desktop | PyInstaller (Docker container used when building from macOS) |
+| `windows` | any host with CodeBuild enabled | AWS CodeBuild (remote build) |
+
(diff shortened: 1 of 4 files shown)


jphuynh:
Thanks! I've also hit that issue. Glad to see a PR to fix it.

jphuynh (approved):

wirjo:
+1 on this

wirjo:
Rebased onto current beta and re-opened from upstream branch (can't push to contributor's fork). Thank you @scouturier! New PR incoming.

PR #329, opened 2026-05-02
fix: use absolute path for package output_dir

### LINKED ISSUE #328: ccwb package fails when run outside source directory
## Bug

`poetry run ccwb package` fails with `Binary not created` even though PyInstaller builds successfully.

**Root cause:** `output_dir` uses a relative path (`Path('./dist')`). PyInstaller writes binaries relative to `source_dir` (its `cwd`), but the existence check resolves relative to the user's working directory.

**Fix:** Call `.resolve()` on `output_dir` to make it absolute.

joel-maws:
Closes #328

`output_dir` was a relative path (`Path('./dist')`), causing PyInstaller to write binaries to a path resolved from `source_dir` while the existence check resolved from the user's working directory. Adding `.resolve()` makes it absolute so both paths match regardless of where `ccwb package` is run.

### DIFF
--- a/.github/workflows/build-macos-arm64.yml
+++ b/.github/workflows/build-macos-arm64.yml
@@ -1,0 +1,52 @@
+name: Build macOS ARM64
+
+on:
+  workflow_dispatch:
+    inputs:
+      profile:
+        description: 'ccwb profile name'
+        required: true
+        default: 'allcode-dev-us-east-1'
+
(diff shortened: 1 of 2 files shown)


wirjo:
Closing — this fix is already in the beta branch. `output_dir` uses `.resolve()` for absolute path resolution (line 330 of package.py).

PR #359, opened 2026-05-09
feat: skip stack query when identity_pool_id known, auto-install deps

Andreas-Allcode:
## Summary

- **Skip CloudFormation stack query** when `identity_pool_id` is already set in the profile (e.g. imported config from a different account). Falls back to stack query if not set.
- **Auto-install Python 3 and AWS CLI** in the generated `install.sh` if they're not already present on the user's machine.

### `config.py`
- Added `identity_pool_id: str | None` field to the `Profile` dataclass

### `package.py`
- Package command now checks for `identity_pool_id` (or `federated_role_arn`) in the profile before querying CloudFormation
- Updated `install.sh` template to auto-install:
  - Python 3 via Homebrew (macOS) or apt/yum/dnf (Linux)
  - AWS CLI via Homebrew (macOS) or official installer (Linux)

## Motivation

When the deploying admin doesn't have CloudFormation access to the account where stacks are deployed (e.g. cross-account setup), the package command fails. This allows importing the identity pool ID directly into the profile to bypass the stack query.

Auto-installing Python and AWS CLI reduces friction for end users who may not have these prerequisites.

## Testing

- Tested package build with `identity_pool_id` set in profile — skips stack query successfully
- Generated install.sh includes prerequisite auto-installation logic


wirjo:
Thanks for the contribution @Andreas-Allcode! The skip-stack-query idea when `identity_pool_id` is already known is useful for cross-account setups.

However, this PR includes ~4,500 lines of unrelated code (15 "nexus" files — marketplace, device auth, metering, governance agent, etc.) that appear to be from a separate project. It also adds build workflows and a compiled binary (`ocb`) that don't belong in this repo.

Could you resubmit with just the focused changes?
1. `package.py` — skip CF stack query when `identity_pool_id` is set
2. Auto-install deps in `install.sh` (if desired)

A clean PR with only those changes would be straightforward to review and merge. Happy to help if you need guidance on splitting the branch.

PR #373, opened 2026-05-15
feat(dashboard): track Opus 4.7, Sonnet 4.6, Haiku 4.6 in Token Usage by Model

jiem-ying:
## Summary

The "Token Usage by Model Over Time" Logs Insights query and the "Token Usage by Model" Lambda widget classifier predate the 4.7 / 4.6 generation Anthropic models. As a result, all traffic on those models is silently dropped from the chart entirely (or bucketed into a generic "Other" via the `else` branch in `get_model_display_name`). The deployed dashboard's legend currently shows Opus 4.6 / 4.1 / 4, Sonnet 4.5 / 4 / 3.7, Haiku 4.5 / 3.5 — leaving Opus 4.7, Sonnet 4.6, and Haiku 4.6 invisible.

## Changes

**`deployment/infrastructure/claude-code-dashboard.yaml`** — extends the `if(model like /.../, ...)` ladder and `stats sum(...)` line with three new buckets:
- `opus_4_7` → `Opus 4.7`
- `sonnet_4_6` → `Sonnet 4.6`
- `haiku_4_6` → `Haiku 4.6`

Also adds the explicit Opus 4.6 1M-context split (`opus_4_6_1m`) that was already present in some deployed dashboards but missing from the source YAML (so a fresh stack deploy would lose it).

**`deployment/infrastructure/lambda-functions/token_by_model/index.py`** — adds matching branches to `get_model_display_name()` and assigns distinct colors in `get_model_color()` for the three new model versions.

## Why ordering matters

In both files, version checks must run from newest to oldest within each family. `opus-4-7` must be matched before `opus-4-6` before bare `opus-4`, otherwise the substring overlap silently swallows newer versions. The diff preserves this ordering.

## Test plan

- [x] `aws cloudformation validate-template` passes for the modified YAML.
- [x] Hot-patched a live `ClaudeCodeDashboard` with the new query — Opus 4.7 / Sonnet 4.6 / Haiku 4.6 appear in the legend immediately, and prior 4.7 traffic populates retroactively (Logs Insights re-evaluates underlying logs).
- [ ] Stack redeploy via `aws cloudformation update-stack` — verify dashboard YAML still renders with both legacy and new model entries.
- [ ] Lambda widget shows Opus 4.7 / Sonnet 4.6 / Haiku 4.6 with distinct colors when those models have non-zero usage.


wirjo:
This is a valid fix for the immediate problem — LGTM for merge.

However, the underlying pattern (hardcoded model lists in Lambdas and CW queries) requires manual updates every time Anthropic ships a new model. I've created issue #400 to track making this dynamic. The approach: replace the `if/elif` chains with a parser that derives display names from model IDs automatically, so new models appear in dashboards without code changes.

wirjo:
No longer needed. The dashboard templates on beta were refactored to use dynamic PromQL queries (`topk(500, sum by (model)(...))`) which automatically group by whatever model string appears in the telemetry. Opus 4.7, Sonnet 4.6, and Haiku 4.6 will show up in the charts as soon as traffic flows — no hardcoded model list to update.

PR #380, opened 2026-05-18
feat: auto-install Python 3 and AWS CLI in install.sh

Andreas-Allcode:
## Summary
The generated install.sh now detects missing prerequisites and installs them automatically:

- **Python 3**: via Homebrew (macOS) or apt/yum/dnf (Linux)
- **AWS CLI v2**: via Homebrew (macOS) or official installer (Linux)

## Problem
End users receiving the distribution package often don't have Python 3 or AWS CLI pre-installed, leading to confusing errors during setup.

## Solution
Added prerequisite checks at the top of the generated install.sh that:
1. Check if python3 is available, install if missing
2. Check if aws CLI is available, install if missing
3. Fall back to clear error messages if auto-install fails

## Testing
Tested on macOS ARM64 (Homebrew path) and Ubuntu 22.04 (apt path).

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -1898,7 +1898,71 @@ echo
 echo "Organization: {profile.provider_domain}"
 echo
 
+# Check prerequisites
+echo "Checking prerequisites..."
 
+# Auto-install Python 3 if missing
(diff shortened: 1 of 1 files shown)


wirjo:
Closing — this is a subset of #359 (same author), which includes auto-install along with additional features.

PR #396, opened 2026-05-22
fix: ccwb init crashes with NoneType error during landing page config

wirjo:
## Summary

Fixes a NoneType crash in `ccwb init` when configuring landing page distribution. The `config.update()` call was incorrectly indented, causing it to run outside the conditional block where the config dict was populated.

## Type of Change

## Changes

- `source/claude_code_with_bedrock/cli/commands/init.py` — fix indentation of `config.update()` call (1 line moved)

### Test Environment

**AWS Region:**  us-east-1
**Python Version:**  3.12.3
**Operating System:**  Linux (Ubuntu 24.04)

### CLI Testing

- [x] **`ccwb init`** with landing page distribution no longer crashes
- [x] All existing tests pass locally

### Unit Tests

- [x] All existing tests pass locally
- AST validation passes

### Test Results

```
$ python3 -c "import ast; ast.parse(open('source/claude_code_with_bedrock/cli/commands/init.py').read()); print('OK')"
OK

$ cd source && python3 -m pytest tests/ -k init -q
39 passed, 238 deselected in 0.67s
```

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1274,24 +1274,39 @@ class InitCommand(Command):
                 # IdP domain
                 idp_domain = questionary.text(
                     "IdP domain (e.g., company.okta.com for Okta, company.auth0.com for Auth0):",
-                    default=config.get("distribution", {}).get("idp_domain", ""),
+                    default=config.get("distribution", {}).get("idp_domain") or "",
                 ).ask()
 
-                # Web app client ID
-                idp_client_id = questionary.text(
-                    "Web application client ID (separate from CLI native app):",
-                    default=config.get("distribution", {}).get("idp_client_id", ""),
-                ).ask()
+                if not idp_domain:
+                    console.print("[yellow]Skipping landing page configuration (no domain provided)[/yellow]")
+                    config["distribution"]["type"] = None
+                    config["distribution"]["enabled"] = False
+                else:
+                    # Web app client ID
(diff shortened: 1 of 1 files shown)


mozammil89:
Thanks for this fix — we hit the same NoneType crash during our deployment and this PR addresses it well.

One potential issue I noticed: the `config["distribution"].update(...)` block (around line 1180 in the new code) appears to be indented inside the `except Exception` block. This means the distribution config only gets saved when Route53 listing fails — not on the happy path where a hosted zone is successfully selected.

It should be dedented one level so it runs after the try/except regardless of outcome:

```python
                except Exception as e:
                    console.print(f"[yellow]Could not list Route53 zones: {e}[/yellow]")
                    hosted_zone_id = None

                # Save landing page configuration (should be outside except block)
                config["distribution"].update(
                    {
                        "idp_provider": idp_provider,
                        "idp_domain": idp_domain,
                        "idp_client_id": idp_client_id,
                        "idp_client_secret_arn": secret_arn,
                        "custom_domain": custom_domain,
                        "hosted_zone_id": hosted_zone_id,
                    }
                )
```

We tested the overall flow and it works for the Cognito auto-detection path (where this block isn't reached). But the manual configuration path would fail to persist the config on success.

wirjo:
Good catch @mozammil89 — fixed. The `config["distribution"].update(...)` was indeed indented inside the `except` block (only ran on Route53 failure). Dedented it so it executes after the try/except regardless of outcome. Pushed.

PR #408, opened 2026-05-26
fix: read monitoring token directly from keyring to avoid auth prompts

Andreas-Allcode:
## Summary
The OTel helper now reads cached tokens directly from the system keyring before falling back to credential-process. This avoids triggering browser auth prompts when credential-process uses device flow or OIDC redirect.

## Problem
When credential-process uses device flow (or any browser-based auth), the OTel helper times out waiting for interaction and falls back to anonymous mode. This causes telemetry to report as 'anonymous' instead of the real user.

## Solution
Read the monitoring token directly from keyring cache. The token may be stored as raw JWT or JSON-wrapped `{"token": "..."}`, so both formats are handled. Falls back to credential-process if keyring is empty.

## Testing
Tested on macOS with keyring storage. Token is correctly parsed from JSON-wrapped format and used for authenticated telemetry.

### DIFF
--- a/source/otel_helper/__main__.py
+++ b/source/otel_helper/__main__.py
@@ -312,8 +312,37 @@ def write_cached_headers(headers, token_exp):
 
 
 def get_token_via_credential_process():
-    """Get monitoring token via credential-process to avoid direct keychain access"""
+    """Get monitoring token via credential-process or directly from keyring cache"""
     logger.info("Getting token via credential-process...")
+
+    # First try reading directly from keyring (avoids browser auth prompts)
+    try:
+        import keyring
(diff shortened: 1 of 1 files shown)


wirjo:
Closing — superseded by the Go credential-process binary in beta (#415, #429).

The problem this PR solves (otel-helper triggering browser auth prompts) is already addressed differently in beta: the otel-helper calls `credential-process --get-monitoring-token` which reads from a fast file-based cache and returns instantly without triggering auth. The Go binary never opens a browser unless explicitly asked.

The direct keyring read approach here targets the Python credential-provider's keyring entries, which are being phased out in favor of the Go binary's split cache architecture. Thank you for the contribution @Andreas-Allcode!

PR #415, opened 2026-05-29
feat(go): add Generic OIDC support, silent refresh, and configurable redirect port

pipeflo:
## Summary

- **Generic OIDC support** — enables the Go credential-process to authenticate against any OIDC-compliant IdP (CyberArk, PingFederate, Keycloak, ForgeRock) via `provider_type: "generic"` in config.json
- **Silent refresh** — reuses cached OIDC ID token to obtain fresh AWS credentials without opening a browser popup (matches Python credential-provider behavior)
- **Configurable redirect port** — `REDIRECT_PORT` env var > `redirect_port` in config.json > default 8400
- **Windows AV fix** — `ccwb package --go` no longer strips Windows binaries (avoids Defender Wacatac.B!ml false positive)
- **Config.json generation** — `ccwb package --go` now writes generic OIDC fields and redirect_port to the distributed config.json

## Motivation

The Go rewrite on beta only supported 4 named providers (Okta, Auth0, Azure, Cognito). Customers using CyberArk Identity, PingFederate, Keycloak, or other generic OIDC IdPs could not use `ccwb package --go` — the binary would error with "Unable to auto-detect provider type".

## Test plan

- [x] All existing Go tests pass (`go test ./... -count=1`)
- [x] New tests added: `TestIsKnown_Generic`, `TestConfigFor_GenericProvider`, `TestDetect_CyberArk`, `TestGenericOIDCConfig`
- [x] Cross-compilation builds all 10 binaries (5 platforms × 2 binaries)
- [x] Python `package.py` syntax verified
- [ ] Manual integration test with Keycloak (Generic OIDC) pending
- [ ] Manual integration test with CyberArk Identity pending (customer call 2026-05-29)

### DIFF
--- a/assets/docs/ARCHITECTURE.md
+++ b/assets/docs/ARCHITECTURE.md
@@ -104,9 +104,9 @@ Our implementation returns credentials in the exact format required by the AWS C
 
 The packaging and distribution system bridges the gap between IT administrators who deploy infrastructure and end users who need simple, foolproof installation. The `ccwb package` command creates a self-contained distribution that includes everything users need without requiring technical expertise.
 
-The packaging system uses pre-built native Go binaries stored in `source/go/prebuilt/`. These are generic — they contain zero customer-specific data and work for all deployments. Go cross-compilation produces binaries for all 5 platforms from a single machine, replacing the previous PyInstaller (macOS/Linux) and Nuitka/CodeBuild (Windows) build pipeline.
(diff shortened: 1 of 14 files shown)


wirjo:
@pipeflo pushed a few improvements on top of your changes:

1. **Secrets baseline** — whitelisted the test fixture files (`generic_oidc_cyberark.json` + `migration_test.go`) to fix the secrets-scanner CI failure

2. **Silent refresh timeout** — added a 10-second timeout to `trySilentRefresh()` so it falls through to browser auth if STS is slow/unreachable instead of hanging

3. **AuthOptions struct** — refactored `oidc.Authenticate()` from 7 positional parameters to a single `AuthOptions` struct. Future provider features can be added by extending the struct without changing the function signature. Only one caller (`main.go`) so it's a clean swap.

Let me know if you have any concerns with these changes.

PR #438, opened 2026-06-04
feat: detect Bedrock use without a running OTEL sidecar (sidecar mode)

avivenk:
## Problem

In sidecar mode, per-user quota usage is derived from telemetry the local OTEL
sidecar reports. A developer who stops the sidecar keeps Bedrock access while
their usage goes uncounted — effectively bypassing quota limits.

## Change

Adds a detective control: a scheduled Lambda (`claude-code-sidecar-monitor`) that
joins tamper-proof CloudTrail Bedrock activity (`InvokeModel`/`Converse`, logged
as management events by default) against OTEL-reported usage in DynamoDB. Users
active in CloudTrail but not reporting telemetry are flagged via per-user
`ClaudeCode/SidecarHealth` CloudWatch metrics and an SNS alert (via the existing
quota alert topic).

Opt-in via `ccwb init` or the `EnableSidecarMonitoring` quota-stack parameter
(enabled by default in sidecar mode; auto-disabled in central mode where the
collector runs server-side and cannot be stopped by users).

## How it works

1. Every 15 minutes, the Lambda queries CloudTrail `LookupEvents` for Bedrock
   runtime calls. These are management events captured by default — no trail or
   data-event charges. Caller email is read from the assumed-role session name.
2. It reads `UserQuotaMetrics` to determine which users reported telemetry in
   the same window.
3. `bedrock_active_users − otel_reporting_users` = users with a stopped sidecar.
4. Publishes CloudWatch metrics and sends an SNS alert listing affected users.

## Scope / limitations

- Detective control only — reports bypass, does not block.
- Tamper-proof enforcement via Bedrock model invocation logging is noted as a
  future enhancement in the docs.

## Testing

- New unit tests: `source/tests/test_sidecar_monitor_lambda.py` (8 cases)
- `cfn-lint` clean on `quota-monitoring.yaml` (only expected W3002 package warnings)
- Existing config + cloudformation + quota suites pass (65 tests total)

## Notes

- Additive only (no changes to existing behavior); all new resources gated on a
  `SidecarMonitoringEnabled` CloudFormation condition.
- Added an `[Unreleased]` CHANGELOG entry. Happy to bump `source/pyproject.toml`
  if maintainers prefer a versioned entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Sidecar bypass detection**: New opt-in detective control (sidecar mode) that flags users invoking Amazon Bedrock without a running OTEL sidecar. A scheduled Lambda joins tamper-proof CloudTrail Bedrock activity against per-user OTEL-reported usage (point reads, no table scan), publishes per-user `ClaudeCode/SidecarHealth` CloudWatch metrics, and sends SNS alerts. Configurable via `ccwb init` or the `EnableBypassDetection` quota-stack parameter (disabled by default).
+
 ## [3.0.0-beta.2] - 2026-05-28
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-### Added
-
-- **Sidecar bypass detection**: New opt-in detective control (sidecar mode) that flags users invoking Amazon Bedrock without a running OTEL sidecar. A scheduled Lambda joins tamper-proof CloudTrail Bedrock activity against per-user OTEL-reported usage (point reads, no table scan), publishes per-user `ClaudeCode/SidecarHealth` CloudWatch metrics, and sends SNS alerts. Configurable via `ccwb init` or the `EnableBypassDetection` quota-stack parameter (disabled by default).
-
 ## [3.0.0-beta.2] - 2026-05-28
 
 ### Fixed
--- a/assets/docs/QUOTA_MONITORING.md
+++ b/assets/docs/QUOTA_MONITORING.md
@@ -643,6 +643,58 @@ Configure in your profile:
 
 **Trade-off**: Shorter sessions mean more frequent re-authentication prompts for users, but provide tighter quota enforcement.
 
+## Sidecar Bypass Detection
+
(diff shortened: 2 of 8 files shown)


wirjo:
Thanks @avivenk — this is a solid detective control. Good design choices: additive-only, gated on a condition, reuses existing SNS topic, and the CloudTrail management events approach is free.

A few pieces of feedback:

**1. Please retarget to `beta`** (not `main`). All active development targets beta — main is release-only.

**2. DynamoDB full table scan is a risk at scale.** The `get_otel_reporting_users()` function does a full `Scan` with a filter expression on every 15-min invocation. For deployments with hundreds/thousands of users, this will hit throughput limits and cost more than expected.

Suggestions:
- Use a GSI on `last_updated` so you can `Query` only recently-active users instead of scanning all
- Or maintain a separate lightweight "heartbeat" item per user that the quota Lambda stamps, and query only those
- Or use a TTL-based approach where you only check users whose heartbeat TTL hasn't expired

**3. Minor:** The PR will need a rebase onto beta (it's currently based on main which is behind). Happy to help with that if needed.

Overall this is a great contribution — addresses a real gap in sidecar mode. Looking forward to the updated version targeting beta.

wirjo:
Adding to my previous comment — the naming `EnableSidecarMonitoring` is also confusing since it sounds like it enables the sidecar itself (which is `monitoring_mode=sidecar`). Consider renaming to something like `EnableBypassDetection` or `EnableSidecarComplianceCheck` to distinguish it from the existing monitoring mode concept.

And to clarify: the suggestion is to default this new parameter to `false` (opt-in) rather than `true`, since it's a compliance audit feature — not core monitoring functionality.

avivenk:
Thanks for the thorough review @wirjo  — all five points addressed and pushed:

1. Retargeted to beta ✓ — PR base is now beta, and the branch is rebased onto the latest beta HEAD.

2. Removed the full-table scan ✓ — Rather than add a GSI or a heartbeat item, I flipped the logic: detection only needs to inspect the CloudTrail-active set (users who actually invoked Bedrock in the window), which is the small working set. For each of those users the Lambda now does a single GetItem point read on their current-month record and checks last_updated freshness. Reads scale with active users, not total users — no Scan, no schema change, and no changes to quota_monitor. IAM was narrowed from dynamodb:Scan to dynamodb:GetItem accordingly. (Bonus: a Bedrock-active user with no metrics record is now correctly treated as a stopped sidecar.)

3. Rebased onto beta ✓ — Resolved conflicts in CHANGELOG.md, the quota template Conditions block, and init.py (adopted the new wizard_fields update-in-place structure).

4. Renamed for clarity ✓ — EnableSidecarMonitoring → EnableBypassDetection, SidecarLookbackMinutes → BypassDetectionLookbackMinutes, condition → BypassDetectionEnabled, Lambda → claude-code-bypass-detection. Docs/section retitled "Sidecar Bypass Detection" to distinguish it from the monitoring_mode=sidecar concept.

5. Default to opt-in ✓ — Defaults to false across the CFN parameter, the init prompt, and the profile field, framed as a compliance/audit feature.

Local verification: 68 tests pass (incl. 11 for the new Lambda covering the point-read freshness logic), cfn-lint clean on quota-monitoring.yaml, and the profile field round-trips through serialization.

One open question for you: the [Unreleased] CHANGELOG entry — do you want a 
pyproject.toml
 version bump in this PR, or do you handle that at release time? Happy to add it either way.

Issue #427, opened 2026-06-03
[Windows] Monitoring id_token never persists in keyring (>1280 chars) — repeated browser auth + broken STS silent refresh


peepeepopapapeepeepo:
## Summary

On Windows with `credential_storage=keyring` and an IdP that issues large id_tokens (e.g. Azure AD, ~1.3 KB), the monitoring id_token is **never persisted**, so background telemetry re-triggers browser authentication repeatedly (effectively on every prompt) and STS silent refresh is broken.

## Environment

- Windows 10/11, `credential_storage=keyring` (Windows Credential Manager)
- Azure AD (Entra ID) OIDC provider, Direct STS federation
- id_token measured at 1284 chars (1372 chars once wrapped in the stored JSON)

## Root cause

`save_monitoring_token()` in `source/credential_provider/__main__.py` stores the id_token as a **single** keyring entry (`{profile}-monitoring`). A single Windows Credential Manager entry is capped at `CRED_MAX_CREDENTIAL_BLOB_SIZE` (5*512 = 2560 bytes); the Python `keyring` backend stores values as UTF-16LE (2 bytes/char), so the practical limit is **~1280 characters**. A real Azure id_token exceeds this, so `set_password` raises `(1783, 'CredWrite', 'The stub received bad data')` — and the exception is silently swallowed (`_debug_print` only). The token is therefore never cached.

Reference: https://learn.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentialw (`CredentialBlobSize` / `CRED_MAX_CREDENTIAL_BLOB_SIZE`).

Note: the STS `SessionToken` already works around this exact limit by splitting into `-token1`/`-token2` entries; that split was never applied to the monitoring token.

## Impact

1. The OTEL header cache (keyed to the id_token expiry) can never be repopulated with a valid token, so telemetry re-invokes `credential-process --get-monitoring-token` on every export and opens a browser.
2. `_try_silent_refresh()` depends on `get_monitoring_token()`; with no persisted token, silent STS refresh fails and the main credential flow falls back to a browser when STS creds expire.

Scope: Windows + keyring only. macOS Keychain and the default session-file storage are unaffected (no small per-entry size limit / file storage).

## Steps to reproduce

1. Windows, `credential_storage=keyring`, Azure AD provider.
2. `credential-process --profile <p> --clear-cache`, then run Claude Code.
[...]

Issue #426, opened 2026-06-03
Windows OTEL Helper Binary Missing from CodeBuild Distribution Package


surya-my:
When building Windows packages via CodeBuild, the distribution package contains `credential-process-windows.exe` but is missing `otel-helper-windows.exe`. Windows users cannot collect OpenTelemetry metrics.

**##** **Root Cause**

The CodeBuild buildspec does not validate that `otel-helper-windows.exe` was successfully compiled. If the Nuitka compilation of the OTEL helper fails, the build still reports SUCCESS and the artifact zip only contains `credential-process-windows.exe`.

surya-my:
Windows Deployment Issues
As a Windows user, you'd encounter the following issues across the deployment lifecycle:

1. install.bat Parser Bug — -replace was unexpected at this time.
The generated install.bat uses inline PowerShell commands within a cmd.exe batch file. The -replace operators in the PowerShell one-liner get consumed by cmd.exe's parser, breaking the installation midway. The binary and config.json are copied, but ~/.claude/settings.json placeholder replacement fails.

Impact: Users must manually complete the installation with a PowerShell script (documented in QUICK_START.md as a known workaround).

2. OTEL Helper Binary Missing from Windows Packages
The _build_otel_helper() function in package.py never builds the OTEL helper for Windows — it only checks if the binary already exists and throws an error if not. This is documented in OTEL_WINDOWS_FIX.md and requires a code patch.

Impact: Windows users cannot collect OpenTelemetry metrics without manual workaround.

3. No Windows Equivalent of otel-helper.sh Cache Wrapper
The Unix otel-helper.sh shell script provides fast cached header lookups (avoids PyInstaller binary startup cost). There is no Windows .ps1 or .bat equivalent. Windows users always pay the full PyInstaller startup cost (~2-3 seconds) every time Claude Code needs OTEL headers.

Impact: Performance penalty on Windows for telemetry header generation.

4. os.chmod Calls with POSIX Modes
Throughout the codebase, os.chmod(path, 0o600) and 0o700 modes are used for security-sensitive files. On Windows, these calls technically succeed but don't enforce POSIX-style permissions — Windows uses ACLs. The files are not actually protected to the intended level.

Files affected: 
__main__.py
, 
__main__.py

Impact: Session tokens, credentials cache, and monitoring tokens have weaker file-level protection on Windows than intended.

5. os.rename Not Atomic on Windows
In 
[...]

PR #441, opened 2026-06-04
fix(otel): resilient empty-headers contract + eliminate per-turn latency

scouturier:
## Summary

Makes the `otel-helper` resilient when no monitoring token is available, and eliminates a per-turn credential-process round-trip. This is the **consumer-side** counterpart to #427: when the monitoring token can't be obtained (the failure #427 describes on Windows/keyring, but also fresh installs, post-expiry, and headless/CI hosts), the helper previously returned exit 1 with empty stdout — violating Claude Code's `otelHeadersHelper` contract. Claude Code then logged `otelHeadersHelper did not return a valid value` and **dropped the telemetry batch on every export cycle**.

Related issues:
- #427 — `[Windows] Monitoring id_token never persists in keyring (>1280 chars)`. That issue's own symptom is *"telemetry re-invokes `credential-process --get-monitoring-token` on every export and opens a browser"* — i.e. the no-token state this PR makes safe. #427/#429 fix the **cause** (token persistence); this PR hardens the **consumer** so the telemetry path degrades gracefully when a token is absent for any reason.
- #426 — `Windows OTEL Helper Binary Missing from CodeBuild Distribution Package`. Related otel-helper packaging context; not fixed here.

## What changed

**Contract correctness (Go + Python)**
- The Go helper now emits a valid empty JSON object (`{}`) and exits 0 instead of `return 1` when no token is available or the JWT fails to decode (`emitEmptyHeaders`). Telemetry export proceeds **unattributed** until the user authenticates, at which point real per-user attribution resumes on the next helper invocation.
- The Python helper's last-resort `except` block now prints `{}` and returns 0 for the same contract reason. It deliberately does **not** cache `{}` (its normal no-token path uses STS anonymous mode earlier), documented inline.
- **No fabricated/anonymous identity is ever emitted** — since this product exists for accurate per-user cost attribution, a stand-in identity would pollute dashboards. Unattributed-but-honest is preferred over wrongly-attributed.

**Per-turn latency**
- Serve headers from the `~/.claude-code-session` Layer-1 cache instead of spawning credential-process on every invocation.
- The empty-headers result is cached with a short (120s) TTL so credential-process isn't re-spawned every export cycle while a token is unavailable.
- `EmptyHeadersWriteSafe` guard: the empty write is skipped if it would overwrite a still-valid populated entry (protects against a transient Layer-1 read failure clobbering real attribution).
- `ReadCachedHeaders` treats empty entries with an expired or non-positive TTL as a miss, so attribution always recovers.
- Atomic, Windows-safe cache writes (`os.replace` / `MoveFileEx` replace-semantics).

> Note on recovery: the 120s TTL bounds when the *helper* refreshes, not when Claude Code re-invokes it. Claude Code calls `otelHeadersHelper` at startup and on a debounce (default 29 min, `CLAUDE_CODE_OTEL_HEADERS_HELPER_DEBOUNCE_MS`), so end-to-end attribution recovery is bounded by that interval.

## Measured impact

Benchmarked on Windows Server 2025 and macOS (arm64), simulating a slow (5s) credential-process with no token:

| | call 1 | steady-state | credential-process spawns (per 20 calls) |
|---|---|---|---|
| Before | ~5s | ~5s **every** call | 20 |
| After | ~5s | ~8ms | 1 |

Plus: 0 → eliminated the recurring `otelHeadersHelper did not return a valid value` error and dropped-batch on every export cycle.

## Testing

- Go: `go test ./cmd/otel-helper/ ./internal/otel/` — pass; `gofmt`/`go vet` clean. New regression tests: contract compliance, fresh-empty Layer-1 hit, populated-expired served-not-rewritten, `EmptyHeadersWriteSafe` matrix, untimed-empty miss.
- Python: `test_otel_latency_fixes.py` — pass; added except-path test (`{}` + exit 0).
- End-to-end: verified on Windows + macOS; a real attributed OTLP metric carrying the helper's headers is accepted by the live collector (HTTP 200).
- Confirmed no impact on Keychain/keyring: `internal/storage` (keyring owner) is imported only by `credential-process`, never by `otel-helper`/`internal/otel`.

## Scope / safety

- Touches only otel-helper code paths; the happy path (valid token) is unchanged.
- No anonymous/fabricated attribution; no interactive auth ever triggered from the telemetry path (avoids the prior "browser popup every ~1h" regression).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.0.0-beta.2] - 2026-05-28
 
+### Fixed
+
+- **otel-helper empty-headers contract**: Both the Go and Python otel-helpers now always print a JSON object (`{}` when no monitoring token is available) and exit 0, fixing the recurring "otelHeadersHelper did not return a valid value" log and dropped telemetry batch that occurred on every export cycle when a user was unauthenticated (e.g. before first login, after token expiry, or on headless/CI hosts). Telemetry then exports unattributed — never with fabricated identity — until the user authenticates, at which point real per-user attribution resumes on the next helper invocation
+- **otel-helper per-turn latency**: Eliminated a credential-process round-trip on every turn by serving cached headers from `~/.claude-code-session`, with atomic cache writes that are safe on Windows. An empty-headers result is cached with a short (120s) TTL and is guarded so it can never overwrite still-valid populated attribution; once authenticated, attribution resumes on the next helper invocation (bounded by Claude Code's headers-helper debounce, default 29 min)
+
 ## [2.4.0] - 2026-05-22 - OTLP-First Metrics
 
 ### Added
--- a/source/go/cmd/otel-helper/main.go
+++ b/source/go/cmd/otel-helper/main.go
@@ -9,12 +9,29 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"ccwb-go/internal/config"
 	"ccwb-go/internal/jwt"
 	"ccwb-go/internal/otel"
 	"ccwb-go/internal/version"
 )
+
+// emptyHeadersCacheTTLSeconds bounds how long THIS binary serves an empty-headers
+// result from the file cache before it retries credential-process. It is short
+// enough that retrying is cheap (a credential-process cache hit is ~20ms) yet
+// long enough to keep telemetry export off the credential-process hot path
+// during a sustained unauthenticated window. It is kept above the shell shim's
+// 60s token_exp skew so the shim and binary agree on freshness.
+//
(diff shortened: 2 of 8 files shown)


wirjo:
Rebased on current beta and removed two out-of-scope changes:

1. **Removed `package.py` change** (dropping `CLAUDE_CODE_ENABLE_TELEMETRY=1` from managed settings) — this is a behavioral change for all users that should be a separate, explicitly-documented PR. Admins currently rely on this being set.

2. **Removed `source/dist/claude-settings/settings.json`** — appears to be an accidentally committed local config with hardcoded ALB endpoint and model IDs.

The otel-helper contract fix + cache improvements are solid. Please open a separate PR for the telemetry opt-in/opt-out discussion if you'd like to pursue that change.

The `OTEL_EXPORTER_OTLP_TIMEOUT=5000` addition was also in `package.py` so it was removed with it — feel free to re-add it as a standalone one-liner PR if desired.

Issue #271, opened 2026-04-16
[OAuth] Refresh token discarded — full browser re-auth required every hour


yagoTobi:
**Environment:** All platforms

**Problem**
The credential process discards the refresh token after initial login. When AWS credentials expire (~1 hour), a full browser re-authentication is triggered instead of a silent background renewal.

This is inconsistent with how Claude Enterprise handles token refresh.

**Proposed fix**
Store and use the refresh token to allow silent background credential renewal with no browser interaction.

**Patch**
[oidc-refresh-token.patch](https://github.com/user-attachments/files/26802145/oidc-refresh-token.patch)

wirjo:
**Partial fix landed in beta:**

- PR #415 (Go rewrite) implements silent refresh — reuses cached OIDC ID token to obtain fresh AWS credentials without opening a browser popup
- PR #429 fixes the Python credential-provider's cache writes (`os.replace` instead of `os.rename`) so cached tokens are actually persisted between invocations on Windows

**Current state:** Users running the Go binary (default in beta) get silent refresh automatically. Users on the Python credential-provider path also benefit from the cache write fix (#429), which was causing the token to never persist (hence re-auth every hour).

**Remaining gap:** If the IdP's `refresh_token` grant is disabled server-side (some Okta orgs, Azure AD with short-lived tokens), silent refresh will still fail and fall back to browser auth. This is IdP-configuration dependent and can't be fixed client-side — but we could improve the error message to point users at their IdP's refresh token settings.

wirjo:
Fixed by PR #447 (refresh token persistence, on beta). The credential-process now stores refresh tokens in `~/.claude-code-session/{profile}-refresh.json` and silently re-authenticates without browser for 7-30 days depending on provider. All OIDC providers support `offline_access` scope. Thanks @yagoTobi for reporting!

PR #447, opened 2026-06-05
feat: persist OIDC refresh_token for silent credential renewal

### LINKED ISSUE #446: Cowork 3P: silent refresh fails for session storage users when id_token expires
## Problem

When using `credential_storage: session` (the default), Claude Cowork 3P (Claude Desktop via Bedrock) loses authentication after the OIDC id_token expires (typically 1-24h depending on IdP configuration).

## Root Cause

The authentication flow for Cowork 3P:
1. Claude Desktop → AWS SDK → `~/.aws/config` named profile → `credential_process = credential-process --profile X`
2. credential-process → checks cached AWS credentials → if expired → `_try_silent_refresh()`
3. Silent refresh uses the cached id_token from `~/.claude-code-session/{profile}-monitoring.json`
4. If id_token itself is expired → silent refresh fails → **browser popup required**

The problem: Claude Desktop cannot display or handle a browser popup initiated by credential-process. The process simply times out or returns an error, and the user sees a generic auth failure in Claude Desktop with no clear recovery path.

## Impact

- **Session storage users** (default): Auth breaks every 1-24h (depending on IdP token lifetime)
- **Keyring storage users**: Same issue, but keyring may persist tokens across reboots longer

## Who is affected

Any Cowork 3P deployment where:
- `credential_storage: session` (default)
- The IdP id_token lifetime is shorter than the user's work session
- The user doesn't also use Claude Code in a terminal (which would refresh the token via browser)

## Possible solutions

1. **Persist refresh_token** (relates to #271): If the OIDC refresh_token is cached, credential-process could silently exchange it for a new id_token without browser interaction. Most IdPs issue refresh tokens with 7-30 day lifetimes.

2. **Extend token cache lifetime**: Cache the id_token aggressively and only force re-auth when the IdP rejects it, rather than preemptively expiring based on JWT `exp` claim.

3. **Background token refresh daemon**: A lightweight process that refreshes tokens before they expire, keeping the cache warm for Cowork 3P.

4. **User guidance**: Document that Cowork 3P users should set their IdP's id_token lifetime to at least 8-12 hours to cover a work day.

## Related

- #271 (refresh token discarded)
- #166 (silent refresh — works within id_token lifetime, but not beyond it)

wirjo:
## Problem

After the OIDC id_token expires (typically 1-24h), credential-process must open a browser popup to re-authenticate. This breaks **Cowork 3P (Claude Desktop via Bedrock)** because Claude Desktop spawns credential-process as a subprocess with no ability to display or handle browser popups. Users see generic auth failures until they manually re-authenticate in a terminal.

Relates to #271 (refresh token discarded), closes #446 (Cowork 3P session storage auth failure).

## Solution

Store the OIDC `refresh_token` returned during initial browser authentication, then use it for silent token renewal when the id_token expires. The refresh_token exchange is a server-to-server HTTP POST — no browser, no user interaction.

### Auth flow (new step 3):

```
credential-process runs:
  1. Check cached AWS creds → valid? → return them ✅
  2. Check cached id_token → valid? → STS exchange → return new AWS creds ✅
  3. NEW: id_token expired → check refresh_token → POST to IdP token endpoint
     → IdP returns fresh id_token → STS exchange → return new AWS creds ✅
  4. refresh_token expired/revoked → open browser (last resort)
```

### Why this fixes Cowork 3P:

| Timeline | Before | After |
|---|---|---|
| 0h (first auth) | Browser popup → tokens cached | Same |
| 1-12h (id_token valid) | Silent refresh ✅ | Same |
| 12-24h (id_token expired) | **Browser needed → Cowork breaks ❌** | Refresh token exchange ✅ |
| 7-30 days (refresh_token expires) | N/A | Browser needed once per week/month |

### Provider compatibility:

| Provider | Scope needed | Default lifetime | Notes |
|---|---|---|---|
| Okta | `offline_access` | 7 days | Must enable in Access Policy |
| Azure AD | `offline_access` | 90 days | Available for public clients |
| Auth0 | `offline_access` | 30 days | Enable "Allow Offline Access" |
| Cognito | Automatic | 30 days | No extra scope needed |
| Generic OIDC | `offline_access` | Varies | Standard OIDC spec |

### Graceful degradation:

If the IdP doesn't issue a refresh_token (scope not enabled, admin-restricted, or provider doesn't support it), the `RefreshToken` field is empty and the behavior is **identical to before** — no regression.

## Changes (7 files, +339/-10)

| File | What |
|---|---|
| `internal/provider/endpoints.go` | Add `offline_access` to scopes for all providers |
| `internal/oidc/flow.go` | Add `RefreshToken` field to `AuthResult` struct |
| `internal/oidc/token.go` | Add `RefreshTokenExchange()` function |
| `internal/storage/refresh.go` | Save/Load/Clear refresh_token (file-based, 0600 perms) |
| `internal/storage/refresh_test.go` | 4 unit tests (save, load, empty, clear) |
| `internal/provider/provider_test.go` | Update scope assertion |
| `cmd/credential-process/main.go` | Wire `tryRefreshToken()` into auth flow + persist on auth |

## Security

- Refresh token stored at `~/.claude-code-session/{profile}-refresh.json` with 0600 permissions
- Same security posture as existing credential cache
- Cleared on `--clear-cache` (logout)
- Failed refresh clears stored token (prevents retry loops with revoked tokens)
- Rotated tokens are persisted (handles IdPs that rotate on each use)

## Testing

- 4 unit tests for storage layer (save, load, empty token no-op, clear)
- Provider test updated for new scope
- Cannot compile locally (Go 1.24 required, CI will validate)


wirjo:
@scourtier Would appreciate your review on this one — it adds refresh_token persistence to enable silent credential renewal for Cowork 3P (Claude Desktop). The main flow change is in `cmd/credential-process/main.go` (new `tryRefreshToken()` step between the existing silent refresh and browser auth fallback).

PR #452, opened 2026-06-05
fix: cross-platform ccwb test + destroy distribution/codebuild stacks

scouturier:
## Summary

Two small, self-contained fixes for issues affecting Windows users and resource cleanup:

### 1. `ccwb test` fails on Windows (`/bin/sh` wrapper)
`ccwb test` wrote the temporary profile's `credential_process` as `/bin/sh -c 'CCWB_PROFILE=… <binary>'`. `/bin/sh` does not exist on Windows, so the test profile could never execute and `ccwb test` failed.

The fix passes the profile via the binary's `--profile` flag instead — the same cross-platform convention the install scripts already use. The binary reads `--profile` and `CCWB_PROFILE` identically, so behavior on macOS/Linux is unchanged.

### 2. `ccwb destroy` orphaned the `distribution` and `codebuild` stacks
`deploy.py` can create `distribution` and `codebuild` stacks, but they were absent from `DESTROYABLE_STACKS`, so `destroy --all` left them standing (orphaned S3/IAM and a CodeBuild project).

Both are now added in reverse-dependency order (`distribution` before `networking`; `codebuild` first), with skip-logic mirroring deploy's `enable_distribution` / `enable_codebuild` gates.

## Testing

- **Windows fix** — verified live on Windows Server 2025 via SSM: the old `/bin/sh` form fails with `WinError 2`; the new `--profile` form returns valid credentials with the argument correctly received. Path-with-spaces case covered.
- **macOS regression** — verified with the real shipped binary: old and new forms return identical credentials.
- **Destroy fix** — verified end-to-end against real AWS infra (deployed `distribution` + `codebuild` in a clean region, confirmed the pre-fix list would orphan both, ran the fixed `destroy --all`, confirmed both reached `DELETE_COMPLETE` with no orphaned resources).
- **Unit tests** — added `test_destroy_stacks.py` (destroy covers every deployable stack, correct teardown order) and `test_test_cross_platform.py` (no `/bin/sh`, command is shell-free and shlex-parseable on POSIX + Windows). Full suite: 418 passed, 4 skipped.

## Files
- `source/claude_code_with_bedrock/cli/commands/test.py` — drop `/bin/sh` wrapper, use `--profile`
- `source/claude_code_with_bedrock/cli/commands/destroy.py` — add `distribution`/`codebuild` + skip-logic
- `source/tests/cli/commands/test_destroy_stacks.py` — new
- `source/tests/cli/commands/test_test_cross_platform.py` — new

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -1172,7 +1172,7 @@ poetry run ccwb destroy [stack] [options]
 
 **Arguments:**
 
-- `stack` - Specific stack to destroy: auth, networking, monitoring, dashboard, or analytics (optional)
+- `stack` - Specific stack to destroy: codebuild, analytics, quota, cowork-dashboard, dashboard, monitoring, distribution, networking, s3bucket, or auth (optional)
 
 **Options:**
 
@@ -1181,7 +1181,7 @@ poetry run ccwb destroy [stack] [options]
 
 **What it does:**
 
-- Deletes CloudFormation stacks in reverse order (analytics → dashboard → monitoring → networking → auth)
+- Deletes CloudFormation stacks in reverse dependency order (codebuild → analytics → quota → cowork-dashboard → dashboard → monitoring → distribution → networking → s3bucket → auth), skipping any not enabled for the profile
 - Shows resources to be deleted before proceeding
 - Warns about manual cleanup requirements (e.g., CloudWatch LogGroups)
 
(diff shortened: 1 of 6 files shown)


scouturier:
**Note on the failing Windows check**

The red `Tests (Python 3.12, windows-latest)` job on this PR is **not caused by these changes** — it's a pre-existing failure on `beta` itself.

The 3 failures are in `tests/test_otel_jwt_expiry.py` (unrelated to this PR's `ccwb test` / `destroy` changes). Root cause: the test fixture redirects the otel-helper cache dir via `HOME`, but `Path.home()` reads `USERPROFILE` on Windows. The bug has existed since that test was added (2026-05-10) and only surfaced when #449 made the Windows job blocking.

Fix is up separately in #453 (one-line, test-only, Windows CI green there). Once #453 merges, this branch will need a rebase onto the updated `beta` for its Windows check to go green.

This PR's own new tests all pass on the Windows runner.

wirjo:
Superseded by #467 (rebased to latest beta with credit). Thank you @scouturier for the original fix!

Issue #126, opened 2026-02-18
Feature: Add ALB JWT validation for OTEL Collector endpoint


peepeepopapapeepeepo:
## Summary

The OTEL Collector endpoint has no authentication. Anyone who can reach the ALB can POST metrics. The otel-helper already has access to the Entra ID JWT but only uses it client-side.

## Proposal

Use ALB native `jwt-validation` action (released Nov 2025) to validate JWT tokens at the ALB before forwarding to ECS.

**Changes:**
- `otel-collector.yaml`: Add `OidcIssuerUrl` and `OidcJwksEndpoint` parameters, `HasJwtAuth` condition, and conditional `jwt-validation` action on HTTPS listener
- `deploy.py`: Pass OIDC issuer URL and JWKS endpoint when deploying monitoring stack with Entra ID
- `otel_helper/__main__.py`: Add `Authorization: Bearer <token>` header to output

**Backward compatible:** When `OidcIssuerUrl` is empty (default), behavior is unchanged (forward only).

**References:**
- [AWS announcement](https://aws.amazon.com/about-aws/whats-new/2025/11/application-load-balancer-jwt-verification/)
- [ALB JWT verification docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-verify-jwt.html)

PR #459, opened 2026-06-05
fix: include Bearer token in otel-helper output for ALB JWT validation

wirjo:
## Problem

PR #129 added ALB JWT validation for the OTEL collector HTTPS endpoint, but the **client-side was never implemented** — neither the Go nor Python otel-helper included the `Authorization: Bearer <token>` header in their output. Since Claude Code sends all otel-helper output headers on OTLP requests, the ALB rejects every telemetry submission with `target_status_code: -` and `classification_reason: JWTHeaderNotPresent` when JWT validation is enabled.

**Impact:** Any deployment with HTTPS + `OidcIssuerUrl` configured on the monitoring stack has completely broken telemetry collection.

## Root Cause

The otel-helper already fetches the monitoring token (id_token) and uses it to extract user claims (`x-user-email`, etc.), but it never passed the token itself as an `authorization` header. The ALB's `jwt-validation` action looks for `Authorization: Bearer <token>` — which was never sent.

## Fix

Include the Bearer token in otel-helper output, covering all OTLP client paths:

| Path | How it reaches the ALB | Fix |
|---|---|---|
| **Claude Code CLI** | otel-helper output → OTEL SDK attaches headers → POST to ALB | Go + Python helper: add `authorization: Bearer <token>` to stdout output |
| **CoWork proxy mode** | CoWork → localhost:4318 proxy → forwards to remote ALB | Proxy's `get_user_headers()` now includes Bearer when forwarding |
| **Sidecar (local collector)** | Claude Code → localhost:4318 → CloudWatch directly | No ALB in path — not affected, no change needed |

## Security: Token never persisted to disk

The Bearer token is added to output AFTER the cache write — never stored in the plaintext cache file (`~/.claude-code-session/{profile}-otel-headers.json`). The cache only contains low-sensitivity attribution headers (`x-user-email`, etc.).

For the Layer 1 cache hit path (where cached attribution headers are served), the token is resolved fresh from `CLAUDE_CODE_MONITORING_TOKEN` env var (free) or credential-process `--get-monitoring-token` (fast, ~20ms).

## Token lifecycle

| Step | What happens |
|---|---|
| Cache hit (Layer 1) | Serve cached attribution + resolve fresh Bearer from env/credential-process |
| Cache miss | Full token fetch → extract claims → cache attribution → add Bearer to output |
| Claude Code debounce | Re-invokes otel-helper every ~29 min |
| Token lifetime | Typically 1-24h (IdP-configurable) |
| Proxy mode | Refreshes headers every 5 min (`_HEADER_REFRESH_SECONDS`) |

## Graceful degradation

- **No token available** → empty headers output → no authorization header → ALB rejects (same as today, no regression)
- **JWT disabled on ALB** (`OidcIssuerUrl` empty) → authorization header is a harmless extra HTTP header the collector ignores

## Changes (4 files, +313/-3)

| File | What |
|---|---|
| `go/cmd/otel-helper/main.go` | Add Bearer to output (after cache write); Layer 1 resolves token from env |
| `go/cmd/otel-helper/main_test.go` | 4 tests: bearer present, not in cache, not when no token, cache hit path |
| `otel_helper/__main__.py` | Bearer in normal output + proxy `get_user_headers()` |
| `tests/e2e/test_cross_platform.py` | 3 static analysis guards (auth contract alignment) |

## Static analysis guards (prevents recurrence)

| Test | What it catches |
|---|---|
| `test_alb_jwt_auth_has_go_helper_bearer_output` | Someone removes Bearer from Go helper while ALB has JWT |
| `test_alb_jwt_auth_has_python_helper_bearer_output` | Same for Python helper |
| `test_bearer_token_not_in_cache_write` | Bearer accidentally moved before cache write (credential leak) |

## Testing

- 8/8 Go otel-helper tests pass (4 new + 4 existing)
- 3/3 static analysis guards pass
- Go build compiles clean

Relates to #126, completes PR #129

### DIFF
--- a/source/go/cmd/otel-helper/main.go
+++ b/source/go/cmd/otel-helper/main.go
@@ -73,6 +73,13 @@ func run(testMode bool) int {
 		headers, err := otel.ReadCachedHeaders(profile)
 		if err == nil && headers != nil {
 			debugPrint("Using cached OTEL headers (token still valid)")
+			// Resolve a Bearer token for ALB JWT validation.
+			// Try env var (free) then credential-process cache (fast ~20ms).
+			if t := os.Getenv("CLAUDE_CODE_MONITORING_TOKEN"); t != "" {
+				headers["authorization"] = "Bearer " + t
+			} else if t, err := getTokenViaCredentialProcess(profile); err == nil && t != "" {
+				headers["authorization"] = "Bearer " + t
+			}
 			outputJSON(headers)
 			return 0
 		}
@@ -119,9 +126,13 @@ func run(testMode bool) int {
 	headers := otel.FormatHeaders(userInfo)
 
 	if testMode {
+		// Include bearer in test output for visibility
+		headers["authorization"] = "Bearer " + token
 		printTestOutput(userInfo, headers)
 	} else {
-		// Cache headers for future calls
+		// Cache attribution headers only — never persist the Bearer token to disk.
+		// The token is sensitive and the cache file has weaker protection than
+		// the keyring/credential-process chain that issued it.
 		tokenExp := int64(claims.GetFloat("exp"))
 		if tokenExp > 0 {
 			if err := otel.WriteCachedHeaders(profile, headers, tokenExp); err != nil {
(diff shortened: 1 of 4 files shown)


wirjo:
@peepeepopapapeepeepo This completes the client-side of your PR #129 (ALB JWT validation). The otel-helper was never sending the `Authorization: Bearer` header, so JWT-enabled deployments silently rejected all telemetry. Would appreciate your review since you designed the original auth flow.

wirjo:
Merged to beta. See also #485 by @peepeepopapapeepeepo which extends this fix to cover the Layer-1 cache-hit path and the sidecar shell wrapper — two paths where the Bearer was still missing. Both are now on beta.

PR #468, opened 2026-06-06
fix: smart destroy with retained resource awareness and pre-cleanup

### LINKED ISSUE #360: bug: destroy command doesn't report S3 buckets with DeletionPolicy: Retain
## Summary

When `ccwb destroy` deletes the distribution stack, the `DistributionBucket` (which has `DeletionPolicy: Retain`) is silently skipped by CloudFormation. The destroy command only reports resources that *fail* to delete (`DELETE_FAILED` status), not resources that are intentionally retained (`DELETE_SKIPPED` status).

This causes users to not see cleanup instructions for the `ccwb2-auth-landing-page-dist-*` bucket, which then blocks redeployment with an "already exists" error.

## Steps to reproduce

1. `ccwb deploy` (creates distribution stack with `DistributionBucket` having `DeletionPolicy: Retain`)
2. `ccwb destroy` (empties bucket but CloudFormation retains it silently)
3. `ccwb deploy` again → fails with: `ccwb2-auth-landing-page-dist-* already exists`

## Expected behavior

The destroy command should report retained resources and show cleanup commands:

```
⚠ Distribution stack — retained resources:
    • DistributionBucket (AWS::S3::Bucket): ccwb2-auth-landing-page-dist-*

S3 Buckets (must be emptied first):
  • ccwb2-auth-landing-page-dist-*
    aws s3 rm s3://ccwb2-auth-landing-page-dist-* --recursive
    aws s3 rb s3://ccwb2-auth-landing-page-dist-*
```

## Root cause

`get_failed_resources()` in `cloudformation.py` only checks for `ResourceStatus == "DELETE_FAILED"`. Resources with `DeletionPolicy: Retain` get `DELETE_SKIPPED` status and are never reported.

wirjo:
## Why

`ccwb destroy` fails silently on non-empty S3 buckets and doesn't report resources kept by `DeletionPolicy: Retain`. Users see generic 'manual cleanup required' without knowing what was retained vs what actually failed. Issue #360.

## What

Cherry-picked destroy improvements from PR #330 (credit: @peepeepopapapeepeepo):

1. **`pre_cleanup_stack()`** — reads CF template to identify Retain resources, then empties S3 buckets and cleans Athena workgroups before deletion. Never touches Retain resources.

2. **`get_retained_resources()`** — reports `DELETE_SKIPPED` resources so users know what was intentionally kept.

3. **DELETE_FAILED retry** — instead of giving up, runs pre-cleanup and retries deletion.

4. **Batch delete** — `_empty_bucket` uses 1000-object batches (S3 API hard limit).

5. **Improved output** — per-resource failure details + retained resource list with cleanup instructions.

## Testing

- Syntax validated (Python AST parse)
- No changes to existing test assertions
- Additive methods on `CloudFormationManager` — zero risk to deploy/other commands

## Credit

Co-authored-by: @peepeepopapapeepeepo (from PR #330)

Fixes #360

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/destroy.py
+++ b/source/claude_code_with_bedrock/cli/commands/destroy.py
@@ -119,6 +119,7 @@ class DestroyCommand(Command):
         console.print("\n[bold]Destroying stacks...[/bold]\n")
 
         all_failed_resources = []  # Collect failed resources from all stacks
+        all_retained_resources = []  # Collect intentionally retained resources
         stacks_with_failures = []
 
         for stack in stacks_to_destroy:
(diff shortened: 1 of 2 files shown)


wirjo:
**Risk notes for this change:**

| Action | Risk | Mitigation |
|--------|------|-----------|
| S3 bucket emptying (all objects + versions) | Destructive — irreversible data loss | Only runs during explicit `ccwb destroy`. Skips buckets with DeletionPolicy: Retain. User already confirmed destruction. |
| Athena workgroup force-delete | Destructive — removes saved queries | Only during destroy. Standard `RecursiveDeleteOption=True` pattern. |
| DELETE_FAILED retry | Low — existing behavior was to give up | Now pre-cleans then retries. Worst case: retry also fails (same outcome as before). |
| Template parsing for Retain detection | Low — wrapped in try/except | If parsing fails, pre-cleanup still runs (just won't skip Retain resources). |

**Backward compatible:** No config changes. Only activates during `ccwb destroy`. Existing workflows (destroy clean stacks) see improved output formatting but identical outcomes.

Issue #365, opened 2026-05-12
Per-user dashboard widgets (ActiveUsers, TopUsers, ActiveHours, Commits) always show 0 when Claude Code sends OTEL without user.email


mortizbey:
## Summary

Four dashboard widgets on `ClaudeCodeDashboard` unconditionally show `0` whenever Claude Code's OTEL stream lacks the `user.email` attribute:

- **Active Users** (`ClaudeCode-Widget-ActiveUsers`)
- **Top Users**
- **Active Hours**
- **Commits**

The root cause is in the aggregator / widget Lambdas, which hard-filter on `@message like /user.email/`. Claude Code's native OTEL emitter only emits `user.id` (a SHA-256 hashed machine UUID) by default, so every metric event is silently dropped from per-user aggregation.

This affects any deployment where the `otel_helper` binary is not in the path, which today includes:
- Deployments with `sso_enabled=false` (observability-only mode, added in #71)
- Any administrator testing the OTEL collector directly via `OTEL_EXPORTER_OTLP_ENDPOINT` in `~/.claude/settings.json` without packaging/installing the CLI binary

## Environment

- Guidance version: v2.3.0 (mainline at commit 105c00b)
- Profile: `sso_enabled: false`, `federation_type: cognito`, `monitoring_enabled: true`, `analytics_enabled: true`
- Claude Code: 2.1.139 on macOS arm64
- Region: us-east-1

## Reproduction

1. Run `ccwb init` and answer **No** to "Enable SSO authentication?"
2. Run `ccwb deploy` — all stacks except auth deploy successfully
3. Configure `~/.claude/settings.json` to point at the collector directly:
   ```json
   {
     "env": {
       "CLAUDE_CODE_USE_BEDROCK": "true",
       "CLAUDE_CODE_ENABLE_TELEMETRY": "1",
       "OTEL_METRICS_EXPORTER": "otlp",
       "OTEL_LOGS_EXPORTER": "otlp",
       "OTEL_EXPORTER_OTLP_PROTOCOL": "http/protobuf",
       "OTEL_EXPORTER_OTLP_ENDPOINT": "https://<your-collector-domain>",
       "AWS_PROFILE": "<your-aws-profile>",
       "AWS_REGION": "us-east-1"
     },
     "model": "us.anthropic.claude-opus-4-7"
   }
   ```
4. Use Claude Code actively for 30+ minutes
[...]

PR #476, opened 2026-06-07
fix: dashboard widgets fall back to user.id when user.email absent

wirjo:
## Why

All per-user dashboard widgets (Active Users, Top Users, Token Usage by User, Cost by User) show **0** when the OTEL stream lacks `user.email` — which happens for:
- IDC (Identity Center) users where email is resolved server-side
- Deployments with `sso_enabled=false`
- Users without otel-helper installed (direct OTEL endpoint usage)

The widgets hard-filtered with `group by ("user.email")` / `sum by ("user.email")`, silently dropping all metrics without that attribute.

## Change

Replace exclusive `user.email` grouping with `user.email, user.id`:

```promql
# Before
count(group by ("user.email")({...}))

# After  
count(group by ("user.email", "user.id")({...}))
```

This ensures metrics are counted regardless of which user identifier is present. Users with otel-helper still show by email; IDC/anonymous users show by their hashed machine ID.

## Testing

Regression test verifies:
1. `user.id` appears in dashboard body
2. No widget groups exclusively by `user.email` without `user.id` fallback

## Risk

**Near zero** — only changes dashboard query grouping labels. No infrastructure, no auth, no credential paths affected.

Relates to #365

### DIFF
--- a/deployment/infrastructure/claude-code-dashboard.yaml
+++ b/deployment/infrastructure/claude-code-dashboard.yaml
@@ -43,7 +43,7 @@ Resources:
                 "title": "Active Users",
                 "view": "number",
                 "region": "${MetricsRegion}",
-                "data": { "queries": [{ "id": "a", "type": "cloudwatch-metrics", "language": "PromQL", "query": "count(group by (\"user.email\")({\"claude_code.session.count\", \"@resource.service.name\"=\"claude-code\"}))", "step": 60, "label": "Users" }] },
+                "data": { "queries": [{ "id": "a", "type": "cloudwatch-metrics", "language": "PromQL", "query": "count(group by (\"user.email\", \"user.id\")({\"claude_code.session.count\", \"@resource.service.name\"=\"claude-code\"}))", "step": 60, "label": "Users" }] },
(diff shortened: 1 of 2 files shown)


wirjo:
Closing this PR — the dual group-by approach has a risk of double-counting or splitting users when both `user.email` and `user.id` exist on the same series.

**The problem is nuanced.** Possible approaches:

1. **`label_replace` fallback** — use `label_replace` to coalesce email→display, falling back to user.id. But unclear if CloudWatch PromQL supports this fully.
2. **Separate widget** — keep main widgets on `user.email`, add an 'Unidentified Users' widget for series where `user.email` is empty, grouped by `user.id`.
3. **otel-helper level fix** — have otel-helper always emit `user.email` (falling back to user.id value when email is unavailable), keeping dashboard queries simple.
4. **Server-side resolution** — resolve email from IAM ARN in the quota Lambda (already done in `feat/aws-iam-idc-support` branch for IDC users).

Each has tradeoffs around PromQL compatibility, dashboard readability, and deployment requirements.

@mortizbey — what's your deployment setup? Specifically:
- Are your users going through otel-helper, or sending directly to the OTEL collector endpoint?
- Which IdP are you using (Cognito, Okta, Azure AD)?
- Would a separate 'unidentified users' widget solve your visibility need?

This would help us pick the right approach.

PR #367, opened 2026-05-14
fix: handle SSO disabled mode in review configuration

dineshSajwan:
When SSO authentication is disabled during init, the Step 4 review configuration was trying to access config['okta'] which doesn't exist, causing 'Error: okta' KeyError.

Changes:
- Wrap OIDC provider/client ID display in sso_enabled check
- Show 'IAM-based (SSO disabled)' when SSO is off
- Only show Identity Pool when SSO is enabled
- Only show Cognito/OIDC Provider in resources list when SSO enabled

Fixes the issue where users selecting 'No' for SSO would hit KeyError at review step.

## Why is this change needed

Fixes # (issue)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Infrastructure/deployment change
- [ ] Dependency update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD changes

## Changes Made

- 
- 
- 

## Breaking Changes

- [ ] No breaking changes
- [ ] Breaking changes described below

## Documentation

- [ ] Updated relevant documentation (README, QUICK_START, guides, etc.)
- [ ] Updated CHANGELOG.md (if user-facing change)
- [ ] Updated version in `source/pyproject.toml` (if applicable)
- [ ] Added/updated code comments and docstrings
- [ ] No documentation changes needed

## Additional Notes

---

### Automated Tests

- [ ] All existing tests pass locally (`ccwb test` or CI checks)
- [ ] Added new tests for this change (if applicable)
- [ ] Test coverage maintained or improved
- [ ] Tests are referenced: 

### Manual Testing - CLI Commands

- [ ] **`ccwb init`** - Tested initialization wizard
  - Screenshot/log: 
  
- [ ] **`ccwb deploy`** - Tested deployment
  - Screenshot/log: 
  - Stack deployed: 
  - Stack status: 
  
- [ ] **`ccwb build`** - Tested building binaries
  - Screenshot/log: 
  - Platform(s) tested: 
  
- [ ] **`ccwb package`** - Tested packaging for distribution
  - Screenshot/log: 
  - Distribution method: 
  
- [ ] **`ccwb test`** - Tested Bedrock connectivity
  - Screenshot/log: 
  - Model tested: 

- [ ] **Binary Installation** - Tested `install.sh` or `install.bat`
[...]

PR #477, opened 2026-06-07
fix: handle SSO disabled mode in review configuration

wirjo:
## Why

`ccwb init` crashes at the review step (Step 5) with a `KeyError` when users select **"No" for SSO**. The `_review_configuration` method directly accesses `config['aws']['identity_pool_name']` and `config['monitoring']['enabled']` — both of which don't exist in non-SSO configurations.

This affects users deploying with IAM Identity Center (IDC) or "no auth" mode — they can't complete the init wizard.

## User Impact

| User type | Before this fix | After |
|-----------|----------------|-------|
| SSO disabled (IDC/none) | ❌ Crash at review step — can't complete init | ✅ Review shows "IAM-based (SSO disabled)" cleanly |
| SSO enabled (OIDC) | ✅ Works | ✅ No change |
| Re-init existing profile | ❌ Crash if monitoring section absent | ✅ Safe `.get()` access |

**Error users saw:**
```
KeyError: 'identity_pool_name'
```
at the review configuration step, after completing all wizard questions.

## Changes

| File | What |
|------|------|
| `init.py` | Guard `identity_pool_name` row behind `if sso_enabled` |
| `init.py` | Use `.get()` for `monitoring.enabled` (safe access) |
| `init.py` | Show "IAM, Monitoring" instead of "Cognito, IAM, Monitoring" for non-SSO |
| `init.py` | Guard auth resource listing (Cognito/OIDC) behind `sso_enabled` |
| `test_review_config_sso_disabled.py` | 4 static analysis regression tests |

## Regression Prevention

The 4 tests use **AST-based static analysis** — they parse `_review_configuration` and verify:

1. `config["aws"]["identity_pool_name"]` never appears (must use `.get()`)
2. `config["monitoring"]["enabled"]` never appears (must use `.get()`)
3. `if sso_enabled` guard exists before Cognito/OIDC resource listing
4. `identity_pool_name` display is inside an `sso_enabled` block

These tests catch regressions at the **code pattern level** — if someone adds a new direct dict access in `_review_configuration`, CI fails immediately without needing to exercise the runtime path.

## Suggested Further Improvements

1. **Apply the same pattern repo-wide:** Other commands (deploy, status) likely have similar bare `config["key"]` access that crashes in non-SSO mode. A broader sweep would help.

2. **Config schema validation:** A `validate_config()` function that checks required fields exist before entering command logic would catch these earlier than at the point of access.

3. **Integration test with non-SSO config:** The static analysis tests catch the pattern, but an actual integration test that runs `_review_configuration` with a minimal non-SSO config dict would catch any NEW fields added without guards.

## Risk

**Low** — only changes display logic in the review step. No auth, credential, or deployment paths affected. The review step is read-only (it just renders a table for the user to confirm).

Credit: @dineshSajwan (original PR #367)
Fixes the crash reported in #367

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1926,9 +1926,12 @@ sso_registration_scopes = sso:account:access"""
                 else "Session Files (temporary)"
             ),
         )
-        table.add_row("Infrastructure Region", f"{config['aws']['region']} (Cognito, IAM, Monitoring)")
-        table.add_row("Identity Pool", config["aws"]["identity_pool_name"])
-        table.add_row("Monitoring", "✓ Enabled" if config["monitoring"]["enabled"] else "✗ Disabled")
+        if sso_enabled:
+            table.add_row("Infrastructure Region", f"{config['aws']['region']} (Cognito, IAM, Monitoring)")
+            table.add_row("Identity Pool", config.get("aws", {}).get("identity_pool_name", "—"))
+        else:
+            table.add_row("Infrastructure Region", f"{config['aws']['region']} (IAM, Monitoring)")
+        table.add_row("Monitoring", "✓ Enabled" if config.get("monitoring", {}).get("enabled") else "✗ Disabled")
         if config.get("monitoring", {}).get("enabled"):
             mode = config.get("monitoring", {}).get("mode", "sidecar")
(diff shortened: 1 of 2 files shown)


wirjo:
Closing — this fix has been applied directly to the `feat/aws-iam-idc-support` feature branch (commit 3a08a44), since the SSO disabled crash is specifically the IDC use case. It'll reach beta when the feature branch is promoted.

PR #485, opened 2026-06-07
fix(otel): emit Authorization Bearer across all otel-helper paths to fix ALB JWT 401s

peepeepopapapeepeepo:
## Why

On HTTPS monitoring deployments with ALB JWT validation enabled, **all telemetry was silently dropped**. Every OTLP request was rejected with `401` because the otel-helper never sent `Authorization: Bearer <jwt>`, even though the ALB's `jwt-validation` action requires it. Verified live: the `ClaudeCode` CloudWatch namespace had zero datapoints for ~6 days until this fix.

This was a recurring regression:

- `74c7e4d` dropped the Bearer header during a refactor.
- `09c2c8a` restored it - but Python-only, and it missed the Layer-1 cache-hit path.
- `076da39` (the Go rewrite) never ported it.

So neither the Go nor the Python helper emitted the Bearer on any path.

## What

Emit `Authorization: Bearer <jwt>` on **every** otel-helper output path, in **both** the Go and Python implementations (per the `credential-helper-parity` rule), with the token added **after** the cache write so it is never persisted to disk.

| Path | Change |
|------|--------|
| Binary normal path (Go + Python) | Add Bearer to output after `WriteCachedHeaders` / `write_cached_headers` |
| Binary Layer-1 cache hit (Go + Python) | Resolve Bearer fresh (env var → credential-process) and splice it in |
| Sidecar wrapper (`otel-helper.sh`) | **Major:** the wrapper served the Bearer-less cached `.raw` and exited before the binary, so the in-binary fix was dead code in sidecar mode. The wrapper now splices the Bearer onto stdout at serve-time, reading the token from the env var or the expiry-validated `monitoring.json` (no `jq`, no binary cold-start) |
| Proxy path (CoWork) | `build_proxy_user_headers()` includes the Bearer for forwarded requests |
| No-token cache hit (Go + Python) | **Major:** previously emitted headers with no Bearer and returned 0 **silently**; now logs the omission so an ALB 401 is diagnosable (matches the Layer-3 no-token path) |
| Python `--test` | Now shows the Bearer in `--test` output (parity with the Go helper) |

Attribution headers (`x-user-*`) are unchanged - this only **adds** an `authorization` header, so the `otel-attribution-chain` header contract is untouched.

## Security invariant

The Bearer token is **never written to disk**. It is added to stdout only, after the cache write in every path; the `.raw` / `.json` cache files contain attribution headers only. Guarded by `test_bearer_token_not_in_cache_file` and shell-wrapper scenario 6 (asserts the token never lands in `.raw`).

## Tier 1 compliance

This touches `source/go/cmd/otel-helper/main.go`, a **Tier 1 Critical** file (`review-tiers`):

- [x] Regression test for every changed code path
- [x] Go ↔ Python parity (Bearer injection + no-token log present in both)
- [x] Security invariant verified (Bearer never persisted to disk)
- [x] Cross-platform CI must pass (no admin merge override)

### Test Environment
- **AWS Region**:  ap-southeast-2
- **Python Version**:  3.12.13
- **Operating System**:  macOS 26.5 (arm64); Go 1.26.4

### Testing Performed
- [x] All existing tests pass locally
- [x] **End-to-End**: rebuilt the package, ran live Claude Code traffic, confirmed metrics flowing into CloudWatch
- [x] New regression test added that would have caught the bug (`test_otel_shell_wrapper.sh` - no prior test exercised the wrapper path that hid the bug)

### Test Results
```
# Shell wrapper - new suite, 7 scenarios (Bearer splice, expiry gate, env precedence,
# graceful no-token, cache-miss fallthrough, on-disk secrecy, empty-{} edge case)
$ bash source/tests/test_otel_shell_wrapper.sh
RESULT: 7 passed, 0 failed

# Go otel-helper (Go 1.26.4)
$ go vet ./cmd/otel-helper/
vet OK
$ go test ./cmd/otel-helper/ -count=1
ok      ccwb-go/cmd/otel-helper

# Python - auth-header suite + smoke + otel regression
$ poetry run pytest tests/test_otel_authorization_header.py tests/test_smoke.py \
    tests/test_otel_latency_fixes.py tests/test_regression_429_441.py tests/test_otel_jwt_expiry.py
106 passed, 4 skipped

# Live stack verification (ap-southeast-2)
ALB:        HTTPCode_ELB_4XX (JWT reject) 22/hr -> ~0; HTTPCode_Target_2XX steady
CloudWatch: continuous 1-min datapoints; token.usage ~19M, cost.usage, session.count,
            active_time.total all current (namespace was empty for ~6 days prior)

# Expiry recovery (cleared cache → cold-start path)
$ credential-process --profile wmg-sbx-apse1 --clear-cache   # removes creds + monitoring token
$ AWS_PROFILE=wmg-sbx-apse1 otel-helper --test --verbose
  -> otel-helper silently re-authenticated via credential-process (~16s, no browser prompt),
     minted a fresh token, emitted a valid Bearer, exit 0
  -> attribution cache (.json/.raw) rewritten with x-user-* only; zero authorization/Bearer on disk
```

### Falsification (proof the tests catch the bug)
Each fix was reverted to confirm the matching test goes red:
- Wrapper bypass reverted -> 3 shell scenarios fail (Bearer-presence asserts).
- Empty-`{}` guard reverted -> scenario 7 fails with malformed `{, "authorization": ...}`.
- Go no-token path made to emit empty `Bearer ` -> `authorization must be absent ... got "Bearer "`.

## Relationship to #459

Supersedes #459 (@wirjo), which targets the same bug but misses the **Python Layer-1 cache-hit path** and the **`otel-helper.sh` wrapper bypass**. This PR is a strict superset with full path coverage plus a shell-wrapper test. Thanks @wirjo for the initial diagnosis and fix direction.

## Notes

- Diff is +770/-20, but ~677 lines are tests (the regression coverage `pr-standards` requires). The production change is ~93 lines across 3 files.
- **Pre-existing, out of scope:** Claude Code posts logs to `/v1/logs` (`OTEL_LOGS_EXPORTER=otlp`), but the collector stack defines only a metrics pipeline, so those POSTs return `4xx` and the logs are dropped. Confirmed identical on `origin/main`; metrics are unaffected.

## Follow-ups (not in this PR)

- Extract a single `attachBearer()` choke point to remove the 7-site Bearer duplication (the shape that caused the original `74c7e4d` regression).
- Validate token `exp` on the env-var path in the binary (the credential-process path is already expiry-guarded).
- Central-mode binary spawns credential-process on every Layer-1 cache hit when the env var is absent - mitigate with an in-process Bearer cache (perf, not correctness).
- Collector has no logs pipeline while clients export logs - either add one or stop exporting logs.

### DIFF
--- a/source/go/cmd/otel-helper/main.go
+++ b/source/go/cmd/otel-helper/main.go
@@ -73,8 +73,8 @@ func run(testMode bool) int {
 		headers, err := otel.ReadCachedHeaders(profile)
 		if err == nil && headers != nil {
 			debugPrint("Using cached OTEL headers (token still valid)")
-			// Resolve a Bearer token for ALB JWT validation.
-			// Try env var (free) then credential-process cache (fast ~20ms).
+			// Resolve Bearer fresh — the cache stores attribution headers only,
+			// never the token itself. Try env var (free) then credential-process (~20ms).
 			if t := os.Getenv("CLAUDE_CODE_MONITORING_TOKEN"); t != "" {
 				headers["authorization"] = "Bearer " + t
 			} else if t, err := getTokenViaCredentialProcess(profile); err == nil && t != "" {
@@ -126,13 +126,12 @@ func run(testMode bool) int {
 	headers := otel.FormatHeaders(userInfo)
 
 	if testMode {
-		// Include bearer in test output for visibility
+		// Include Bearer in test output so --test shows the full header set.
 		headers["authorization"] = "Bearer " + token
 		printTestOutput(userInfo, headers)
 	} else {
-		// Cache attribution headers only — never persist the Bearer token to disk.
-		// The token is sensitive and the cache file has weaker protection than
-		// the keyring/credential-process chain that issued it.
+		// Cache attribution headers only — Bearer token must never be persisted
+		// to the plaintext cache file. Add it to output AFTER the cache write.
 		tokenExp := int64(claims.GetFloat("exp"))
 		if tokenExp > 0 {
 			if err := otel.WriteCachedHeaders(profile, headers, tokenExp); err != nil {
@@ -141,10 +140,6 @@ func run(testMode bool) int {
 		} else {
 			debugPrint("JWT has no exp claim, skipping cache write")
 		}
-
-		// Add Bearer token AFTER cache write — output only, never persisted.
-		// This enables ALB JWT validation (PR #129 / issue #126) while keeping
-		// the sensitive token out of the plaintext cache file.
 		headers["authorization"] = "Bearer " + token
 		outputJSON(headers)
 	}
--- a/source/go/cmd/otel-helper/main.go
+++ b/source/go/cmd/otel-helper/main.go
@@ -79,6 +79,12 @@ func run(testMode bool) int {
 				headers["authorization"] = "Bearer " + t
 			} else if t, err := getTokenViaCredentialProcess(profile); err == nil && t != "" {
 				headers["authorization"] = "Bearer " + t
+			} else {
+				// No token from env var or credential-process. Emit cached attribution
+				// anyway (otelHeadersHelper contract), but log so an ALB 401 is
+				// diagnosable instead of silent — the no-token path below logs the same way.
+				debugPrint("Layer 1 cache hit but no Bearer token available " +
+					"(env var empty, credential-process failed); emitting headers without authorization")
 			}
 			outputJSON(headers)
 			return 0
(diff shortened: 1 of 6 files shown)


wirjo:
Thanks @peepeepopapapeepeepo — excellent work identifying these gaps and testing against a live Entra ID stack. The two paths you found (Layer-1 cache bypass + shell wrapper) would have been very hard to catch without real-world testing.

Particularly impressed with:
- The shell wrapper test suite (first coverage this component has ever had)
- The security invariant being maintained (Bearer never in cache file)
- The graceful fallback when no token is available

Both this and #459 are now merged to beta. The full Bearer fix is complete across all 3 layers (Go binary, Python helper, shell wrapper).

PR #486, opened 2026-06-07
feat: skip CloudFormation stack query when identity_pool_id is in profile

Andreas-Allcode:
## Summary
Skip the `DescribeStacks` call when `identity_pool_id` or `federated_role_arn` is already set in the profile config.

## Problem
In cross-account or pre-configured setups, the CloudFormation stack may not exist in the current account, causing `ccwb package` to fail. Users had to manually work around this.

## Solution
Check if the profile already has the needed values before querying CloudFormation. Falls back to the existing stack query when neither value is set.

## Changes
- `source/claude_code_with_bedrock/cli/commands/package.py` — single focused change

## Testing
Tested with pre-configured profiles where `identity_pool_id` is set directly in config.json. Package generation succeeds without CloudFormation permissions.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -206,34 +206,42 @@ class PackageCommand(Command):
             return 1
 
         # Get actual Identity Pool ID or Role ARN from stack outputs
-        console.print("[yellow]Fetching deployment information...[/yellow]")
-        stack_outputs = get_stack_outputs(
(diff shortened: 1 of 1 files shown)


wirjo:
Closing — this is already addressed by PR #436 (preserve existing profile fields when re-running ccwb init), which ensures identity_pool_id and other profile fields are preserved across re-init without needing to query CloudFormation. Thanks @Andreas-Allcode for raising this!

wirjo:
Note: PR #436 is on the `beta` branch. Check there for the profile preservation logic that covers this use case.

PR #502, opened 2026-06-09
fix: resolve UnboundLocalError for region across all landing-page IdPs during init

peepeepopapapeepeepo:
## Why
When AWS setup is skipped (progress marked complete), the `region` variable is never assigned but is later used in the landing-page distribution section, causing `UnboundLocalError` that crashes the `ccwb init` wizard.

**Root cause:** `region` is bound only at the AWS-setup prompt (inside `if not skip_aws:`). The landing-page block runs unconditionally and uses `region` for both Cognito stack detection and the Secrets Manager client. When `skip_aws=True` (last_step in `aws_complete`/`monitoring_complete`/`bedrock_complete` with no existing config), neither path binds it.

A previous fix bound `region` only inside `if idp_provider == "cognito":`, so it rescued just **1 of 5** providers — Okta, Azure, Auth0, and Google still crashed at the Secrets Manager client (`region_name=region`).

**Trigger:** First-time init where the user completes AWS setup, progress marks it complete, then continues to distribution with a non-Cognito landing-page IdP in the same session.

## What
Resolve `region` once at the top of the landing-page block so it is bound for every IdP provider, regardless of `skip_aws`. Remove the now-redundant per-cognito assignment.

```python
# top of `if distribution_type == "landing-page":`
region = config.get("aws", {}).get("region", get_current_region())
```

Falls back to: (1) saved config (populated during AWS setup), or (2) the AWS CLI default region.

## Testing
- Reproduced the crash deterministically for all 5 providers under `skip_aws=True`.
- New regression test `TestLandingPageRegionResolution` drives the real `_gather_configuration` down the `skip_aws` landing-page path, parametrized over all 5 providers. It **fails on the prior cognito-only fix** for okta/azure/auth0/google and **passes** with this change.
- Full suite: 914 passed. (One pre-existing, unrelated failure on `beta` — `test_cfn_contracts.py` flags `cloudtrail:LookupEvents` in `quota-monitoring.yaml`; not touched by this PR.)

Per `review-tiers.md`, `init.py` is Tier 2 — regression test included.

---

## ⚠️ Merge dependency — merge #503 first

This PR's CI is **red until #503 merges**, through no fault of its own code.

The `pytest-ci` gate runs the **full `pytest tests/` suite**, which currently hits a **pre-existing failure on `beta`**: `test_cfn_contracts.py` rejects `cloudtrail:LookupEvents` in `quota-monitoring.yaml` (the allow-list is missing `cloudtrail`). This PR doesn't touch that test, so it inherits beta's failure. **#503 fixes the allow-list.**

**Action:** merge **#503 first**, then **re-run this PR's checks** — GitHub rebuilds the PR merge ref against the updated `beta`, so it goes green automatically (no rebase or extra commits required).

Verified locally by simulating the GitHub merge ref:
| Simulated CI state | cfn contract gate |
|---|---|
| This PR on current `beta` | ❌ red (`cloudtrail` not in allow-list) |
| This PR after #503 in `beta` | ✅ green (21 passed) |

Depends on #503.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1220,6 +1220,10 @@ class InitCommand(Command):
             console.print("\n[bold]Landing Page Configuration[/bold]")
             console.print("Configure IdP authentication for the distribution landing page")
 
+            # Resolve region for landing-page infra ops (Cognito detection + Secrets Manager) even
+            # when AWS setup was skipped (skip_aws=True), so region is bound for all IdP providers.
+            region = config.get("aws", {}).get("region", get_current_region())
+
             # IdP provider selection
             idp_choices = [
                 questionary.Choice("Okta", value="okta"),
@@ -1246,7 +1250,7 @@ class InitCommand(Command):
                 console.print("\n[bold]Cognito Configuration Detection[/bold]")
                 console.print("Searching for deployed Cognito User Pool stack...")
 
-                # Try to auto-detect Cognito stack
+                # Try to auto-detect Cognito stack (region resolved at top of landing-page block)
                 cognito_stack_info = detect_cognito_stack(region)
 
                 if cognito_stack_info:
--- a/source/tests/cli/commands/test_init.py
+++ b/source/tests/cli/commands/test_init.py
@@ -615,6 +615,103 @@ class TestSsoDisabledRendering:
         assert param_map["OktaClientId"] == "abc123xyz"
 
 
+class TestLandingPageRegionResolution:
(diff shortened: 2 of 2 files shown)


wirjo (approved):
LGTM. Verified locally:

- All 34 init tests pass (including the 5 new parametrized provider tests)
- Targets `beta` ✅, mergeable ✅
- The cfn_contracts failure is pre-existing (fixed by #503 already on beta) — resolves on merge
- Tier 2 change (init.py) with regression test included ✅
- Minimal fix: 1 line of actual code change + good test coverage

Risk: Very low. Ship it.

PR #410, opened 2026-05-28
test: add E2E CLI lifecycle, contract, and CloudFormation validation tests

wirjo:
## Why

This repo has grown from a reference architecture into production infrastructure (400+ issues, active community), but the test suite only covers unit-level logic. There are no tests that exercise:

1. **The full CLI lifecycle** — a user creating a profile, switching contexts, validating config, and exporting/importing. Bugs like #390 (init NoneType crash), #285 (KeyError at review step), and #283 (sso_enabled not pre-populated) all happen in these paths and ship undetected.

2. **Cross-component contracts** — the credential-provider binary outputs JSON that must conform to the [AWS credential_process spec](https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-credential_process.html). The quota_check Lambda returns responses that credential-provider parses. If either side changes a field name, users silently lose authentication. No tests verify this agreement.

3. **CloudFormation template integrity** — 20 templates totalling thousands of lines of YAML, but nothing validates they parse correctly, have valid IAM action prefixes, or contain expected resources. Issue #375 (invalid `bedrock-runtime:` prefix — actually valid, but the class of bug where typos in IAM actions cause silent permission failures) would be caught here.

These gaps mean regressions reach users through PRs that "pass CI" because CI never exercised the affected paths.

## What

**110 new tests** in `source/tests/e2e/` across 3 files:

### `test_cli_lifecycle.py` (14 tests)
- Profile create → save → reload round-trip (no data loss)
- Multiple profile coexistence
- Profile update without clobbering unrelated fields
- Export/import JSON fidelity
- `context list`, `context use`, `context current`, `context show` command execution
- `config validate` on well-formed profiles
- Quota command instantiation and JSON structure

### `test_contracts.py` (19 tests)
- AWS credential_process output spec: required keys, Version=1, ASIA* prefix, ISO 8601 expiration
- No sensitive data (IdToken, RefreshToken) leaks into credential output
- Profile fields that credential-provider reads exist with correct types
- Config JSON key names stable (contract with external binary)
- Quota check response structure: allowed/blocked schemas, usage sub-fields
- OTEL helper monitoring token structure

### `test_cfn_contracts.py` (77 tests)
- All 20 CloudFormation templates parse as valid YAML (with intrinsics)
- All templates contain Resources and Description
- IAM actions use valid AWS service prefixes (validates against known list)
- Quota monitoring Lambda environment variables match what code reads
- DynamoDB key schema validation

[...]

Issue #375, opened 2026-05-15
IAM policies in bedrock-auth-*.yaml templates use invalid "bedrock-runtime:" action prefix


qing-at-work:
## Bug

The four authentication CloudFormation templates ([`bedrock-auth-okta.yaml`](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/deployment/infrastructure/bedrock-auth-okta.yaml), [`bedrock-auth-auth0.yaml`](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/deployment/infrastructure/bedrock-auth-auth0.yaml), [`bedrock-auth-azure.yaml`](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/deployment/infrastructure/bedrock-auth-azure.yaml), [`bedrock-auth-cognito-pool.yaml`](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/main/deployment/infrastructure/bedrock-auth-cognito-pool.yaml)) all reference IAM actions with a `bedrock-runtime:` prefix in their `BedrockAccessPolicy` managed policy:

```yaml
Action:
  - 'bedrock:InvokeModel'
  - 'bedrock:InvokeModelWithResponseStream'
  - 'bedrock:CallWithBearerToken'
  - 'bedrock:ListFoundationModels'
  - 'bedrock:ListInferenceProfiles'
  - 'bedrock-runtime:InvokeModel'              # invalid
  - 'bedrock-runtime:InvokeModelWithResponseStream'  # invalid
  - 'bedrock-runtime:ConverseStream'           # invalid
  - 'bedrock-runtime:Converse'                 # invalid
```

## Why these are invalid

`bedrock-runtime` is the **API endpoint hostname** (`bedrock-runtime.<region>.amazonaws.com`), not an IAM service prefix. Per the [AWS Service Authorization Reference for Amazon Bedrock](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonbedrock.html), every Bedrock IAM action lives under the `bedrock:` prefix, and the entire data-plane API surface (`InvokeModel`, `InvokeModelWithResponseStream`, `Converse`, `ConverseStream`) is authorized via `bedrock:InvokeModel`.

[...]

PR #503, opened 2026-06-09
fix: correct IAM action-prefix allow-list in cfn contract test

peepeepopapapeepeepo:
## Why
The cfn contract test's `VALID_ACTION_PREFIXES` allow-list has two errors that together (a) turn beta CI red and (b) silently disable the #375 guard:

1. **Missing `cloudtrail`** — `quota-monitoring.yaml` legitimately uses `cloudtrail:LookupEvents` (added in #438 for sidecar bypass detection), but the allow-list never gained `cloudtrail`, so the test fails deterministically:
   ```
   AssertionError: Invalid IAM action prefix 'cloudtrail' in quota-monitoring.yaml: cloudtrail:LookupEvents.
   ```
2. **Spurious `bedrock-runtime`** — the test is documented to *catch* the bogus `bedrock-runtime:` prefix (#375, see `.claude/rules/iam-actions.md`), yet lists it as valid, so `bedrock-runtime:InvokeModel` would pass straight through.

Verified against the AWS Service Authorization Reference:
- `cloudtrail` is a valid IAM prefix — `cloudtrail:LookupEvents` is a real Read action.
- Amazon Bedrock's only IAM prefix is `bedrock` (singular). `bedrock-runtime` is an SDK/API endpoint name, **not** an IAM namespace.

## What
One-line correction to `VALID_ACTION_PREFIXES`: add `cloudtrail`, remove `bedrock-runtime`.

## Testing
- **Repro:** `quota-monitoring.yaml` parametrization failed before, passes after.
- **Regression — guard restored:** no template uses `bedrock-runtime:` (grep), so removal surfaces no hidden violation; re-injecting a `bedrock-runtime:InvokeModel` action into a template now correctly **fails** the test again (it would have passed before).
- Full `test_cfn_contracts.py` suite green (88 passed).

Credit: `cloudtrail:LookupEvents` introduced by @avivenk in #438; the test allow-list introduced by @wirjo in #410.

### DIFF
--- a/source/tests/e2e/test_cfn_contracts.py
+++ b/source/tests/e2e/test_cfn_contracts.py
@@ -99,7 +99,7 @@ class TestIAMPolicyValidity:
 
     # Valid IAM action prefixes for services used in this project
     VALID_ACTION_PREFIXES = {
-        "bedrock", "bedrock-runtime", "cognito-identity", "cognito-idp", "sts", "logs",
+        "bedrock", "cloudtrail", "cognito-identity", "cognito-idp", "sts", "logs",
         "cloudwatch", "s3", "s3express", "s3-object-lambda", "s3outposts",
         "dynamodb", "lambda", "iam", "ssm",
         "firehose", "glue", "athena", "kms", "codebuild", "ec2",


wirjo (approved):
LGTM. Verified:

- `cloudtrail:LookupEvents` is used in `quota-monitoring.yaml` (sidecar bypass detection from #438) — valid IAM prefix
- `bedrock-runtime` is an SDK endpoint name, not an IAM namespace — removing it restores the #375 guard as intended
- `grep -r 'bedrock-runtime' deployment/infrastructure/*.yaml` returns nothing — no hidden violations

One-line test-only change, fixes CI, no runtime impact. Very low risk.

PR #521, opened 2026-06-11
fix(package): treat Windows-only CodeBuild build as success, not failure

### LINKED ISSUE #520: ccwb package (Windows-only) wrongly reports "No binaries were successfully built" and exits 1
### Describe the bug
Running `ccwb package` for **Windows only** on a non-Windows host (macOS/Linux) prints:

```
Error: No binaries were successfully built.
Please check the error messages above.
```

and exits `1` — even though the Windows build was submitted to CodeBuild successfully and is running. Because the command returns non-zero early, it also **skips config/installer generation**, so no distributable package is assembled for the Windows build.

### Reproducible steps
1. On a profile with `enable_codebuild: true`, run `ccwb package` and select only Windows (on macOS/Linux).
2. Observe: the command prints `Error: No binaries were successfully built.` and exits 1.
3. Meanwhile `ccwb builds` shows the Windows CodeBuild build as submitted/running.

### Root cause
Windows builds run asynchronously in CodeBuild: `_build_executable("windows")` starts the build and returns `None` (no local binary yet), so `built_executables` is empty. The success check is a bare `if not built_executables:` (`source/claude_code_with_bedrock/cli/commands/package.py`), which treats the normal async-Windows state as total failure. `main` guards this case with a `windows_codebuild_pending` check; `beta` does not.

### Version / environment
- Repo: `beta` (v2.5.0). **Not present on `main`.**
- Host: macOS (non-Windows); profile with `enable_codebuild: true`.

### Why CI did not catch it
No regression test exercised the Windows-only async path (`_build_executable` → `None`) through the success guard.

peepeepopapapeepeepo:
## Description

A Windows-only `ccwb package` on a non-Windows host submits the build to CodeBuild asynchronously, so `_build_executable` returns `None` and `built_executables` is empty. The bare `if not built_executables:` guard then reports failure and exits 1, skipping config/installer generation. This PR restores the `windows_codebuild_pending` escape hatch (present on `main`): an async Windows submission is treated as success, the user is told the build is running, and the config/installer are still generated for distribution.

## Why is this change needed

Windows-only packaging on macOS/Linux is unusable on `beta`: it errors with `No binaries were successfully built` and produces no package, even though the CodeBuild build was started successfully. `main` is unaffected.

Fixes #520

## Type of Change

## Changes Made

- `source/claude_code_with_bedrock/cli/commands/package.py`: compute `windows_codebuild_pending` (true only when `windows` is in the build set AND `profile.enable_codebuild`); change the guard to `if not built_executables and not windows_codebuild_pending:`; print a "building in CodeBuild — run `ccwb builds` / `ccwb distribute`" message for the pending case and continue to config/installer generation.
- `source/tests/cli/commands/test_package_windows_codebuild_pending.py` (new): asserts a Windows-only async build (`_build_executable` → `None`, `enable_codebuild=True`) exits 0; and that a genuine "no binary + CodeBuild disabled" case still exits 1 (so the fix does not over-suppress real failures).

## Additional Notes

- The two cases are distinguished by `enable_codebuild`, so the existing behavior for genuine build failures is preserved (verified by the existing `test_package_executable_path_init.py`, which still passes).

### What was tested

- [x] CLI commands tested (`ccwb init/deploy/build/package/test`)

## Testing Evidence

**Test Environment:** ap-southeast-2, Python 3.12, macOS (Apple silicon)

**Test Results:**

```text
# 1. Full unit suite (rebased on beta) — pass
919 passed, 22 skipped, 1 xfailed, 1 xpassed

# 2. New regression tests (2/2 on the fix)
$ poetry run pytest tests/cli/commands/test_package_windows_codebuild_pending.py -q
2 passed

# 3. FALSIFICATION — the same tests run against beta's package.py (no fix):
#    test_windows_only_async_build_is_not_a_failure  -> FAILED  (assert 1 == 0)
#      stdout: "Error: No binaries were successfully built."   <- exact reported symptom
#    test_windows_only_without_codebuild_still_errors -> PASSED (genuine failure still errors)
#  => proves the test catches the bug and the fix resolves it.

# 4. Existing guard test unaffected
$ poetry run pytest tests/cli/commands/test_package_executable_path_init.py -q
1 passed

# 5. CI gates verified locally (incl. main-only): bandit (no new findings vs beta),
#    semgrep --config auto (0 findings), detect-secrets (no new secrets). cfn_nag /
#    validate-regions / go-tests are N/A (diff touches only package.py + a test).
```

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -496,11 +496,28 @@ class PackageCommand(Command):
                         except Exception as e:
                             console.print(f"[yellow]Warning: Could not build OTEL helper for {platform_name}: {e}[/yellow]")
 
-        # Check if any binaries were built
-        if not built_executables:
+        # A Windows build runs asynchronously in CodeBuild and produces no local
+        # binary now (_build_executable returns None), so built_executables can be
+        # empty even though the build was submitted successfully. Treat that as a
+        # success and still generate the config/installer for distribution.
+        windows_codebuild_pending = any(
+            platform_name == "windows" and platform_name not in [p for p, _ in built_executables]
+            for platform_name in platforms_to_build
+        ) and bool(profile and getattr(profile, "enable_codebuild", False))
+
+        # Check if any binaries were built (or are pending in CodeBuild)
+        if not built_executables and not windows_codebuild_pending:
             console.print("\n[red]Error: No binaries were successfully built.[/red]")
             console.print("Please check the error messages above.")
             return 1
+
+        if windows_codebuild_pending and not built_executables:
+            console.print("\n[bold cyan]Windows binaries are building in AWS CodeBuild[/bold cyan]")
+            console.print("Local configuration files will be generated now for distribution.")
+            console.print("\nTo check build status:")
+            console.print("  [cyan]poetry run ccwb builds[/cyan]")
+            console.print("\nOnce complete, retrieve binaries with:")
+            console.print("  [cyan]poetry run ccwb distribute[/cyan]\n")
 
         # Create configuration
         console.print("\n[cyan]Creating configuration...[/cyan]")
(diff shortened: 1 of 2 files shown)


wirjo (approved):
✅ LGTM — CI all green, low risk.

- Narrowly scoped guard: only fires for Windows + enable_codebuild=True
- getattr() backward compat with older Profile objects
- Both positive and negative test cases included
- Doesn't affect non-Windows packaging paths

Nice work @peepeepopapapeepeepo!

PR #526, opened 2026-06-11
refactor(otel): centralize the Bearer-attach literal into one helper per language

### LINKED ISSUE #525: otel-helper: the Bearer-attach literal is duplicated at 8 sites, one cost-attribution-critical contract with no single source of truth
### Describe the issue
The `Authorization: Bearer <jwt>` header that the OTEL collector's ALB jwt-validation action checks is constructed from a hand-written literal — `"Bearer " + token` (Go) / `f"Bearer {token}"` (Python) — plus an omit-on-empty rule, repeated at **8 sites**:

- `source/go/cmd/otel-helper/main.go`: lines 79, 81, 136, 149
- `source/otel_helper/__main__.py`: lines 647, 831, 878, 928

There is no single source of truth for this contract, and no single assertion point for the Go↔Python parity the repo otherwise enforces (`.claude/rules/credential-helper-parity.md`).

### Impact
This is a maintainability / regression-surface issue, **not a live bug** — current `beta` emits the header correctly on every path (all otel-helper tests green). The concern is the *blast radius* of the duplication: the literal feeds the header an ALB rejects with a 401 when wrong or missing, which drops the telemetry batch and makes per-user cost attribution go dark. With 8 copies, a future refactor can skew or drop one while the other seven still pass tests — an invisible split. This class of break has bitten this code before (the attach path was once removed as collateral in an unrelated refactor, then restored).

### Root cause
Organic growth: each new output path (Layer-1 cache hit, test mode, normal mode, proxy mode) re-implemented the same literal inline rather than calling a shared setter.

### Maintainer decision requested: Option A vs Option B

| | Option A — keep 8 inline copies | Option B — centralize (proposed PR) |
|---|---|---|
| Effect | Status quo; literal stays inline at each site | One setter per language (`attachBearer` / `_attach_bearer`); every site routes through it |
| Behavior change | none | none (proven byte-identical, Go == Python) |
| Touches critical files? | no | yes — Tier-1 `main.go`, Tier-2 `__main__.py` |
| Regression surface | 8 independent drift points | 1 choke point + 1 parity assertion + a single-occurrence guard |
| Best when | "don't churn critical files for no functional gain" | "harden a fragile, cost-attribution-critical contract against silent drift" |

The attached PR implements **Option B**. **Please decide A vs B — feel free to close the PR if you'd rather not touch the Tier-1/Tier-2 helpers for a no-functional-change refactor.** Both positions are reasonable.

### Version / environment
- Repo: `beta` (v2.5.0). The duplication is **identical on `main`** (same line numbers; `main` is currently an ancestor of `beta`) — a pre-existing state, not a recent regression.

### Why CI didn't surface it
It's not a failure — all 8 copies currently agree, so every behavioral test passes. Duplication is invisible to behavior-level tests by definition; only a structural/static guard catches it (the attached PR adds one).

peepeepopapapeepeepo:
> 🟡 **Maintainer decision requested — proposal, not a unilateral change.** This is a preventive DRY refactor with **no live bug** behind it (Option B in #525). It touches the Tier-1 Go helper and Tier-2 Python helper for zero functional change. If you'd rather keep the 8 inline copies (Option A) — e.g. "don't churn critical files without a functional reason" — please **close this PR freely**, no attachment.

## Description

The `Authorization: Bearer <jwt>` literal and its omit-on-empty rule were hand-written at 8 sites (4 in the Go otel-helper, 4 in the Python otel-helper). This introduces one setter per language — Go `attachBearer`, Python `_attach_bearer` — and routes every site through it, so the contract has a single source of truth and the Go↔Python parity check has a single assertion point.

The setters are intentionally **exp-agnostic**: token-expiry buffers diverge across the codebase (600s in `storage/monitoring.go`, 60s in the credential provider, shell wrapper, and `read_cached_headers`), so folding an expiry check into the shared setter would create a parity hazard. Expiry validation stays at the token source, where it already lives.

## Why is this change needed

The literal feeds the header the OTEL collector's ALB jwt-validation action checks; when it's wrong or missing the ALB 401s, the telemetry batch is dropped, and per-user cost attribution goes dark. Eight inline copies are eight independent chances for a future edit to skew or drop one while the other seven still pass tests — a silent attribution split. This is a *preventive* hardening: current `beta` emits the header correctly on every path.

Fixes #525

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Changes Made

- `source/go/cmd/otel-helper/main.go`: add `attachBearer(headers, token)` (exp-agnostic, documented); replace the 4 inline literals. Call-site ordering preserved — the post-`WriteCachedHeaders` attach stays after the cache write, so the token still never lands in the cached map.
- `source/otel_helper/__main__.py`: add `_attach_bearer(headers, token)`; replace the 4 inline literals. Surrounding `if token:` guards left intact (no control-flow simplification).
- `source/go/cmd/otel-helper/main_test.go`: new `TestAttachBearer` — non-empty token sets `Bearer <token>`, empty token omits the key, existing attribution preserved.
- `source/tests/test_otel_bearer_parity.py` (new): behavioral `_attach_bearer` test (mirror of the Go side) + cross-language static parity guard + single-choke-point guard (the literal must appear exactly once per language in non-test source).
- `source/tests/e2e/test_cross_platform.py`: teach the existing "Bearer added after `WriteCachedHeaders`" ordering guard to recognize the centralized `attachBearer(` call. The security property is unchanged; only the pattern matcher was updated (and re-verified to still fire if the attach is moved before the cache write).

## Additional Notes

- Two look-alike Bearer sites are deliberately **left untouched**: the quota-check call in `credential_provider/__main__.py` (unrelated HTTP request) and the shell splice in `otel-helper.sh` (a shell wrapper can't share a Go/Python function, and it has its own `token_exp > now+60` gate). Centralizing those would be over-DRYing.

---

### What was tested

- [x] CLI commands tested (`ccwb test`)

## Testing Evidence

**Test Environment:** Python 3.12, Go 1.26, macOS arm64; live test against an Azure-federated profile (ap-southeast-2).

**Test Results:**

```text
# 1. Full unit suite (rebased on beta) — pass
923 passed, 22 skipped, 1 xfailed, 1 xpassed

# 2. Go otel-helper package — pass (incl. new TestAttachBearer)
$ cd source/go && go test ./cmd/otel-helper/...
ok   ccwb-go/cmd/otel-helper

# 3. New parity + guard tests — pass
$ poetry run pytest tests/test_otel_bearer_parity.py -q
7 passed

# 4. ZERO-BEHAVIOR-CHANGE proof (real Azure JWT, deployed profile):
#    otel-helper normal-mode JSON is BYTE-IDENTICAL before vs after the refactor.
#    (The --test diff was non-deterministic Go map-iteration order in
#    printTestOutput — the SAME binary differs from itself run-to-run — and is
#    identical once sorted. Not a behavior change.)

# 5. FALSIFICATION:
#    a) new tests vs beta's pre-refactor files -> ImportError (_attach_bearer
#       absent) + literal count == 4 (guard asserts == 1): all fail on beta,
#       pass only on the fix.
#    b) the updated e2e ordering guard still FIRES when attachBearer is moved
#       before WriteCachedHeaders (auth_line < cache_write_line -> assert fails),
#       proving the security check wasn't defanged.

# 6. End-to-end on real infra: ccwb package (native Go build) + ccwb test
$ poetry run ccwb test --profile <azure-profile>
   Found OTEL helper: otel-helper-macos-arm64
   Summary: 6 passed, 0 warnings, 0 failed   (Auth, IAM Role, Bedrock, Inference
   Profiles, Quota Monitoring) — packaged binary attaches Bearer + attribution
   against a real monitoring token.

# 7. CI gates verified locally (incl. main-only): bandit on changed Python ->
#    no issues; semgrep --config auto -> only a pre-existing urllib finding on a
#    line not in this diff (baseline noise); detect-secrets --baseline -> no new
#    secrets (test JWTs are alg:none unsigned). go vet clean.
#    cfn_nag / validate-regions N/A (no CFN / region-data change).
```

### DIFF
--- a/source/go/cmd/otel-helper/main.go
+++ b/source/go/cmd/otel-helper/main.go
@@ -45,6 +45,16 @@ func debugPrint(format string, args ...interface{}) {
 	}
 }
 
+// attachBearer sets the Authorization header. Intentionally exp-AGNOSTIC:
+// the Layer-1 cache-hit path never decodes the token, so token validation
+// lives at the token source (see storage.GetMonitoringToken), NOT here.
+// Do not add an exp check inside this function.
+func attachBearer(headers map[string]string, token string) {
+	if token != "" {
+		headers["authorization"] = "Bearer " + token
+	}
+}
+
 func main() {
 	testMode := flag.Bool("test", false, "Run in test mode with verbose output")
 	verboseFlag := flag.Bool("verbose", false, "Show verbose output")
@@ -76,9 +86,9 @@ func run(testMode bool) int {
 			// Resolve Bearer fresh — the cache stores attribution headers only,
 			// never the token itself. Try env var (free) then credential-process (~20ms).
 			if t := os.Getenv("CLAUDE_CODE_MONITORING_TOKEN"); t != "" {
-				headers["authorization"] = "Bearer " + t
+				attachBearer(headers, t)
 			} else if t, err := getTokenViaCredentialProcess(profile); err == nil && t != "" {
-				headers["authorization"] = "Bearer " + t
+				attachBearer(headers, t)
 			} else {
 				// No token from env var or credential-process. Emit cached attribution
 				// anyway (otelHeadersHelper contract), but log so an ALB 401 is
@@ -133,7 +143,7 @@ func run(testMode bool) int {
 
 	if testMode {
 		// Include Bearer in test output so --test shows the full header set.
-		headers["authorization"] = "Bearer " + token
+		attachBearer(headers, token)
 		printTestOutput(userInfo, headers)
 	} else {
 		// Cache attribution headers only — Bearer token must never be persisted
@@ -146,7 +156,7 @@ func run(testMode bool) int {
 		} else {
 			debugPrint("JWT has no exp claim, skipping cache write")
 		}
-		headers["authorization"] = "Bearer " + token
+		attachBearer(headers, token)
 		outputJSON(headers)
 	}
 
--- a/source/go/cmd/otel-helper/main_test.go
+++ b/source/go/cmd/otel-helper/main_test.go
@@ -8,6 +8,41 @@ import (
 	"testing"
 )
 
+// TestAttachBearer is the behavioral half of the Go↔Python Bearer parity check
+// (the static half lives in tests/test_otel_bearer_parity.py). It pins the
+// contract the single attachBearer choke point must hold: a non-empty token
+// yields exactly "Bearer <token>" under the "authorization" key, and an empty
+// token attaches nothing (sending "Bearer " with no JWT is worse than omitting
+// the header — the ALB would 401 on it).
+func TestAttachBearer(t *testing.T) {
+	t.Run("non-empty token sets Bearer header", func(t *testing.T) {
+		headers := map[string]string{}
+		attachBearer(headers, "abc.def.ghi")
+		if got := headers["authorization"]; got != "Bearer abc.def.ghi" {
+			t.Errorf("authorization = %q, want %q", got, "Bearer abc.def.ghi")
+		}
+	})
+
+	t.Run("empty token omits the key", func(t *testing.T) {
+		headers := map[string]string{}
+		attachBearer(headers, "")
+		if _, ok := headers["authorization"]; ok {
+			t.Errorf("empty token must not set 'authorization', got %q", headers["authorization"])
+		}
+	})
+
+	t.Run("existing attribution is preserved", func(t *testing.T) {
+		headers := map[string]string{"x-user-email": "a@b.com"}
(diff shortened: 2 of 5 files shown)


wirjo (approved):
✅ LGTM — CI all green (Go Tests ✅, Python Tests ✅, Integration ✅, Semgrep ✅).

- Centralizes Bearer-attach into single helper per language (Go: attachBearer, Python: _attach_bearer)
- Empty-token guard prevents sending "Bearer " with no JWT
- Parity tests verify Go↔Python produce identical output
- Static analysis test ensures literal can't be duplicated again

Solid refactor. Thanks @peepeepopapapeepeepo!

Issue #499, opened 2026-06-08
Bug: quota stack in fine-grained mode never records usage and never applies per-user policies


gubaruch:
## Summary

Two defects in `deployment/infrastructure/quota-monitoring.yaml` make the quota stack effectively non-functional in fine-grained mode (`enable_finegrained_quotas: true`):

1. **`QuotaMonitorRole` is missing `dynamodb:UpdateItem` on `UserQuotaMetrics`.** The monitor Lambda fetches usage from CloudWatch PromQL and tries to upsert per-user counters into DynamoDB — every call silently fails with `AccessDeniedException`. As a result, **`UserQuotaMetrics` stays empty** and `ccwb quota usage <email>` always reports `0` regardless of real usage.

2. **`QuotaCheckFunction` is missing the `ENABLE_FINEGRAINED_QUOTAS` env var.** PR #280 fixed the deploy-time parameter wiring (`deploy.py` now passes `EnableFinegrainedQuotas=true` to CFN when the profile sets it), and the **monitor** Lambda receives it correctly (line 202). But the **check** Lambda's `Environment.Variables` block (line ~269) never sets it. Inside `quota_check/index.py:resolve_quota_for_user`, line 269 checks `if not ENABLE_FINEGRAINED_QUOTAS ...` — when the env var is missing it short-circuits to the env-default policy and **never queries the `QuotaPolicies` table**. Per-user and per-group policies set via `ccwb quota set-user`/`set-group` are silently ignored; everyone gets the deployment-wide default.

Together, these mean fine-grained quota enforcement does nothing in practice: the engine can't see real usage, and even if it could, it would never apply per-user policies.

## Version / environment

- main @ `ccc46ea` (current HEAD as of 2026-06-08)
- `enable_finegrained_quotas: true`
- `quota_monitoring_enabled: true`

## Steps to reproduce

1. Deploy the stack with fine-grained quotas:
   ```bash
   ccwb init  # or hand-edit profile: enable_finegrained_quotas=true, quota_monitoring_enabled=true
   ccwb deploy quota
   ```
2. Set a tight per-user policy:
   ```bash
   ccwb quota set-user alice@example.com --monthly-limit 50K --enforcement block
   ```
3. Push usage past 50K tokens for `alice@example.com` via the OTel collector (or use any source emitting `claude_code.token.usage` with `user.email=alice@example.com`).
[...]

PR #532, opened 2026-06-12
fix(quota): make quota enforcement functional end-to-end

aidinkhosrowshahi:
## Summary

Per-user/group quota enforcement is currently non-functional. Three independent defects each break a different stage of the pipeline (record usage -> resolve policy -> authorize the check call). With all three present, a user who is far over their limit is never blocked.

## Root causes & fixes

**1. Monitor can't persist usage - missing `dynamodb:UpdateItem`**
`lambda-functions/quota_monitor/index.py` (`update_quota_metrics`) calls `update_item` for atomic increments, but `QuotaMonitorRole` in `quota-monitoring.yaml` only grants `Query/PutItem/GetItem/Scan`. Every roll-up fails with `AccessDeniedException ... dynamodb:UpdateItem on .../UserQuotaMetrics`.
Fix: add `dynamodb:UpdateItem` to the `UserQuotaMetrics` statement.

**2. `/check` ignores per-user policies - missing env var**
`quota_check/index.py` reads `ENABLE_FINEGRAINED_QUOTAS`, but the template sets it only on `QuotaMonitorFunction`, not `QuotaCheckFunction`. The check Lambda therefore always defaults to `false` and applies the global env limits, silently ignoring the `QuotaPolicies` table.
Fix: pass `ENABLE_FINEGRAINED_QUOTAS: !Ref EnableFinegrainedQuotas` to `QuotaCheckFunction`.

**3. Okta quota authorizer issuer mismatch**
`deploy.py` builds the quota JWT authorizer issuer from the bare `provider_domain` (`https://<domain>`), but Okta authenticates via the default custom authorization server, so tokens carry `iss=https://<domain>/oauth2/default` (consistent with `credential_provider`'s `/oauth2/default/v1/*` endpoints). The authorizer rejects every `/check` request; with `quota_fail_mode=open` this silently disables enforcement (and with `fail_mode=closed` it would block all access).
Fix: append `/oauth2/default` to the issuer for `provider_type == "okta"`.

## Reproduction

1. Deploy with quota monitoring + `enable_finegrained_quotas: true`, Okta provider.
2. Set a low per-user policy (`ccwb quota set-user <email> --monthly-limit 10K --enforcement block`).
3. Generate >10K tokens; observe the monitor log `AccessDenied` on `UpdateItem` (bug 1).
4. After fixing 1, invoke `/check` - it returns `allowed:true` with the global 225M limit, ignoring the policy (bug 2).
5. After fixing 2, the credential-process still issues credentials because the authorizer 401s the call (bug 3).
6. With all three fixed, `/check` returns `allowed:false / monthly_exceeded` and the credential-process prints `ACCESS BLOCKED - QUOTA EXCEEDED`.

## Changes

- `deployment/infrastructure/quota-monitoring.yaml` - add `dynamodb:UpdateItem`; add `ENABLE_FINEGRAINED_QUOTAS` to `QuotaCheckFunction`.
- `source/claude_code_with_bedrock/cli/commands/deploy.py` - Okta issuer `/oauth2/default`.

Verified end-to-end on a live deployment (Okta direct-IAM federation).

### DIFF
--- a/deployment/infrastructure/quota-monitoring.yaml
+++ b/deployment/infrastructure/quota-monitoring.yaml
@@ -154,6 +154,7 @@ Resources:
                 Action:
                   - dynamodb:Query
                   - dynamodb:PutItem
+                  - dynamodb:UpdateItem
                   - dynamodb:GetItem
                   - dynamodb:Scan
                 Resource:
@@ -273,6 +274,7 @@ Resources:
           DAILY_TOKEN_LIMIT: !Ref DailyTokenLimit
           DAILY_ENFORCEMENT_MODE: !Ref DailyEnforcementMode
           MONTHLY_ENFORCEMENT_MODE: !Ref MonthlyEnforcementMode
+          ENABLE_FINEGRAINED_QUOTAS: !Ref EnableFinegrainedQuotas
       Code: ./lambda-functions/quota_check/
 
   # HTTP API Gateway (v2 - cheaper and faster than REST API)
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -778,6 +778,15 @@ class DeployCommand(Command):
                     oidc_issuer_url = profile.provider_domain
                     if oidc_issuer_url and not oidc_issuer_url.startswith(("http://", "https://")):
                         oidc_issuer_url = f"https://{oidc_issuer_url}"
+                    # Okta authenticates via its custom default authorization server, so
+                    # issued tokens carry iss=https://<domain>/oauth2/default. The quota JWT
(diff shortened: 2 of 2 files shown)


aidinkhosrowshahi:
Closing this in favor of a focused PR rebased on current `main`.

On review, two of the three fixes here are already resolved in `main` via #499 (the `dynamodb:UpdateItem` permission and the `ENABLE_FINEGRAINED_QUOTAS` env var on QuotaCheckFunction). This branch was cut from a stale base (ccc46ea), so it duplicates that work.

The remaining genuinely-unfixed item (the Okta quota authorizer issuer missing `/oauth2/default`), plus a related per-user daily-enforcement gap, will be submitted as a clean PR against current `main`.

PR #533, opened 2026-06-12
fix(quota): repair three quota-enforcement defects (Okta issuer, per-user daily block, monitor same-day write)

aidinkhosrowshahi:
## Summary

Two independent quota-enforcement defects, both still present on `main` (v2.4.0). Each prevents a configured limit from actually blocking the user. (Branched fresh off current `main`; supersedes my earlier stale #532, whose other two fixes are already resolved via #499/#500/#501.)

## 1. Okta quota authorizer issuer mismatch

`deploy.py::_resolve_oidc_config` builds the quota JWT authorizer issuer from the bare `provider_domain` (`https://<domain>`). But Okta authenticates via the **default custom authorization server**, so issued tokens carry `iss=https://<domain>/oauth2/default` (consistent with `credential_provider`'s `/oauth2/default/v1/*` endpoints and the IAM-provider issuer logic elsewhere in `deploy.py`).

Result: API Gateway's JWT authorizer rejects every `/check` request (the backend Lambda is never invoked). With `quota_fail_mode=open` this **silently disables enforcement**; with `fail_mode=closed` it would deny all access.

**Fix:** append `/oauth2/default` to the issuer for `provider_type == "okta"` (idempotent — won't double-append).

## 2. Per-user/group daily limits can never hard-block via the CLI

The quota check Lambda already honors `daily_enforcement_mode` independently (PR #273), but the **write path** never sets it: `QuotaPolicy`, `QuotaPolicyManager.create_policy/update_policy`, and the `ccwb quota set-user/set-group/set-default` commands only ever handled a single `enforcement_mode` (monthly). `daily_enforcement_mode` stayed unset and defaulted to `alert`, so a daily cap created with `--enforcement block` only alerted and never blocked.

**Fix:**
- Add `daily_enforcement_mode` to `QuotaPolicy` (+ DynamoDB (de)serialization), defaulting to `alert` (backward compatible).
- Thread it through `create_policy` / `update_policy`.
- Add a `--daily-enforcement` flag to `set-user`, `set-group`, and `set-default`.

## Reproduction (issue 2)

```bash
ccwb quota set-user alice@example.com --monthly-limit 225M --daily-limit 1000 --enforcement block
# generate >1000 tokens for alice, let the monitor record it
# invoke /check for alice -> allowed:true (daily breach only "alerts")
```
With the fix:
```bash
ccwb quota set-user alice@example.com --monthly-limit 225M --daily-limit 1000 \
  --enforcement alert --daily-enforcement block
# /check -> allowed:false, reason daily_exceeded
```

## Changes

- `source/claude_code_with_bedrock/cli/commands/deploy.py` — Okta `/oauth2/default` issuer.
- `source/claude_code_with_bedrock/models.py` — `QuotaPolicy.daily_enforcement_mode` + (de)serialization.
- `source/claude_code_with_bedrock/quota_policies.py` — thread `daily_enforcement_mode` through create/update.
- `source/claude_code_with_bedrock/cli/commands/quota.py` — `--daily-enforcement` flag on all three `set-*` commands (+ shared parse helper).
- Tests: corrected the Okta issuer expectation, added no-double-append guard, and `daily_enforcement_mode` persistence coverage. **All 109 quota tests pass.**

Verified end-to-end on a live Okta direct-IAM deployment: after both fixes, `/check` returns `allowed:false / daily_exceeded` and the credential-process prints `ACCESS BLOCKED - QUOTA EXCEEDED`.

### DIFF
--- a/deployment/infrastructure/lambda-functions/quota_monitor/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_monitor/index.py
@@ -124,8 +124,22 @@ def update_quota_metrics(usage_data):
             daily_reset = existing.get("daily_date") != current_date
 
             update_expr = "ADD total_tokens :delta, input_tokens :inp, output_tokens :out, cache_tokens :cache"
+            expr_values = {
+                ":delta": Decimal(str(int(delta))),
+                ":inp": Decimal(str(int(usage.get("input_tokens", 0)))),
+                ":out": Decimal(str(int(usage.get("output_tokens", 0)))),
+                ":cache": Decimal(str(int(usage.get("cache_tokens", 0)))),
+                ":ts": now.isoformat().replace("+00:00", "Z"),
+                ":ttl": ttl,
+                ":email": email,
+            }
             if daily_reset:
                 update_expr += " SET daily_tokens = :delta, daily_date = :date, last_updated = :ts, #ttl = :ttl, email = :email"
+                # :date is only referenced on the reset path; including it otherwise
+                # makes DynamoDB reject the whole UpdateItem with a ValidationException
+                # ("Value provided in ExpressionAttributeValues unused"), which silently
+                # froze same-day usage accumulation after the first daily write.
+                expr_values[":date"] = current_date
             else:
                 update_expr += ", daily_tokens :delta SET last_updated = :ts, #ttl = :ttl, email = :email"
 
@@ -133,16 +147,7 @@ def update_quota_metrics(usage_data):
                 Key={"pk": f"USER#{email}", "sk": f"MONTH#{current_month}"},
                 UpdateExpression=update_expr,
                 ExpressionAttributeNames={"#ttl": "ttl"},
-                ExpressionAttributeValues={
-                    ":delta": Decimal(str(int(delta))),
-                    ":inp": Decimal(str(int(usage.get("input_tokens", 0)))),
-                    ":out": Decimal(str(int(usage.get("output_tokens", 0)))),
-                    ":cache": Decimal(str(int(usage.get("cache_tokens", 0)))),
-                    ":date": current_date,
-                    ":ts": now.isoformat().replace("+00:00", "Z"),
-                    ":ttl": ttl,
-                    ":email": email,
-                },
+                ExpressionAttributeValues=expr_values,
             )
         except Exception as e:
             print(f"Error updating quota for {email}: {e}")
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -1345,6 +1345,17 @@ class DeployCommand(Command):
             if issuer_url and not issuer_url.startswith(("http://", "https://")):
                 issuer_url = f"https://{issuer_url}"
 
+        # Okta authenticates via its default custom authorization server, so issued
+        # tokens carry iss=https://<domain>/oauth2/default. The quota JWT authorizer
+        # must match that exact issuer or every /check request 401s (and, with
+        # fail-open, silently disables enforcement).
+        if (
+            profile.provider_type == "okta"
+            and issuer_url
+            and not issuer_url.rstrip("/").endswith("/oauth2/default")
+        ):
+            issuer_url = f"{issuer_url.rstrip('/')}/oauth2/default"
+
         # Auth0 tokens include trailing slash in iss claim, so authorizer must match
         if profile.provider_type == "auth0" and issuer_url and not issuer_url.endswith("/"):
             issuer_url += "/"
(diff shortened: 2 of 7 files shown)


aidinkhosrowshahi:
Added a third fix to this branch (commit `c44ed2e`):

**Monitor `UpdateItem` fails on same-day usage writes**

`quota_monitor/index.py::update_quota_metrics` always supplies `:date` in `ExpressionAttributeValues` but only references it on the daily-reset branch. On every subsequent **same-day** write, DynamoDB rejects the `UpdateItem` with:
```
ValidationException: Value provided in ExpressionAttributeValues unused in expressions: keys: {:date}
```
Effect: per-user usage records **once per UTC day** (the first/reset write) then silently freezes until the next midnight reset — so usage is drastically under-counted and enforcement never triggers for the rest of the day. (Reproduced live: a counter stuck for ~2.5h immediately resumed accumulating after the fix, and the `/check` daily block then fired.)

Fix: build `ExpressionAttributeValues` incrementally and include `:date` only on the reset path. Present in current `main`.

wirjo:
## Review

**Risk: Medium** (Tier 1: `deploy.py` | Tier 2: `quota_monitor/index.py`, `quota.py`, `models.py`)
**Retargeted:** `main` → `beta`

---

### Auth Flow Relevance

| Defect | Okta | Azure | Google | Generic OIDC | IDC |
|--------|------|-------|--------|--------------|-----|
| 1. Issuer mismatch | ✅ **All Okta users** | ❌ | ❌ | ❌ | ❌ |
| 2. Daily enforcement | ✅ | ✅ | ✅ | ✅ | ✅ (universal) |
| 3. Same-day DDB write | ✅ | ✅ | ✅ | ✅ | ✅ (universal) |

### Detailed Assessment

**Defect 1 (Okta issuer):** Clean, surgical fix. The `/oauth2/default` suffix is Okta's standard Custom Authorization Server path. The idempotent check (`if not endswith('/oauth2/default')`) is correct — prevents double-appending if someone already configured the full URL. Mirrors the existing `auth0` trailing-slash fix on the adjacent line. ✅

**Defect 2 (daily_enforcement_mode):** Well-structured feature addition:
- New field with `alert` default = backward compatible
- DynamoDB serialization handles missing field gracefully (`policy_data.get()`)
- CLI exposes as `--daily-enforcement` (separate from `--enforcement` for monthly)
- Shared `_parse_enforcement()` helper reduces duplication

**Potential simplification:** The `_parse_enforcement()` helper uses a tuple return `(mode, ok)` pattern. Could use Python's native `Enum.__members__` lookup instead, but the current approach is explicit and readable. No change needed.

**Defect 3 (conditional :date):** This is the most critical fix. The bug causes **all same-day usage recording to silently fail** after the first daily write. The fix correctly splits `expr_values` construction so `:date` is only included when the UPDATE expression references it.

**Potential regression risk:** If the `daily_reset` flag is ever computed incorrectly, the `:date` value won't be present → DDB won't update the date → next invocation still thinks it's a "same day" → usage accumulates correctly but date never advances. This is the **safe failure mode** (over-counts daily usage rather than under-counting).

### Suggestions (non-blocking)

1. ℹ️ The `set-default` command diff appears truncated (only `set-user` and `set-group` shown). Confirm `set-default` also has `--daily-enforcement` threaded.

2. ℹ️ Consider adding to release notes: "Existing deployments must redeploy the quota-monitor Lambda (`ccwb deploy --stack monitoring`) to pick up defect 3 fix."

3. ℹ️ Test name `test_okta_issuer_no_double_append` is good — confirms the guard works.

### Verdict

**Approve.** All three fixes are well-reasoned, backward compatible, and verified on a live deployment. High confidence this doesn't regress — the Okta fix is additive (only touches Okta path), the daily mode defaults to existing behavior, and the DDB fix makes the code strictly more correct.

Awaiting CI results on the beta target.

wirjo:
Thanks for this @aidinkhosrowshahi — excellent debugging, especially tracking down the DDB ValidationException root cause. These are real bugs affecting production deployments.

---

## Backward Compatibility & Regression Check

**Defect 1 (Okta issuer):**
- Only runs when `provider_type == "okta"` — no other auth flows touched
- Idempotent guard (`if not endswith('/oauth2/default')`) — safe to re-run
- Existing deployments with manually-correct issuer: no-op
- **No regression path**

**Defect 2 (daily_enforcement_mode):**
- New field defaults to `"alert"` — exact same behavior as today
- Existing DynamoDB records without this field: `policy_data.get("daily_enforcement_mode", "alert")` → graceful fallback
- Only changes behavior when admin explicitly sets `--daily-enforcement block`
- **No regression path**

**Defect 3 (DDB :date unused attribute):**
- Moves `expr_values` construction above the branch and conditionally adds `:date` only on reset path
- Safe failure mode: if `daily_reset` logic were wrong → over-counts daily (conservative), never under-counts
- Same DDB key structure, same TTL, same email — no schema change
- **No regression path** ✅

---

## Merge Plan

Retargeted to `beta`. Will squash-merge once CI passes — no additional changes needed.

Note for release: existing deployments need to redeploy the quota-monitor Lambda (`ccwb deploy --stack monitoring`) to pick up defect 3 fix.

PR #535, opened 2026-06-12
style: apply ruff format to Python source

spawnia:
## Why

The project has `ruff-format` as a pre-commit hook, but the existing Python source doesn't pass it. This makes contributing harder since any touched file triggers formatting noise in unrelated PRs.

## What

Ran `ruff format` on all Python files under `source/`. No functional changes — formatting only (18 files).

This is intentionally separate from #534 (targeted fixes) per the contributing guidelines:

> If you also reformat all the code, it will be hard for us to focus on your change.

🤖 Generated with Claude Code


wirjo:
Closing in favour of #552 (rebased onto latest `beta` and merged). Thanks @spawnia for keeping the codebase clean! 🧹✨

PR #537, opened 2026-06-12
fix: active SSO auth for IDC users with quota enforcement and OTEL attribution

### LINKED ISSUE #401: feat: enable quota enforcement for IAM Identity Center users
## Problem

Quota enforcement currently requires an OIDC ID token (JWT) for user identification. IAM Identity Center users authenticate via `aws sso login` which produces AWS credentials but no OIDC token. As a result, per-user token quotas **cannot be enforced** on the IDC path.

## Proposed Solution

Unify quota enforcement to work with both auth types using the **existing** quota infrastructure (API Gateway + Lambda + DynamoDB). No new infrastructure required.

### Changes

**1. API Gateway route:** Remove JWT authorizer requirement. Let the Lambda handle authentication internally.

**2. Quota Lambda (`quota_check/index.py`):** Accept identity from either source:
- JWT Bearer token (OIDC users) — existing path, extract email from claims
- IAM SigV4-signed request (IDC users) — extract email from caller ARN session name

```python
def get_user_email(event):
    # OIDC: extract from JWT
    auth_header = event.get("headers", {}).get("authorization", "")
    if auth_header.startswith("Bearer "):
        claims = decode_jwt(auth_header[7:])
        return claims.get("email")
    
    # IDC: extract from IAM caller ARN
    # AWSReservedSSO_PermSet_abc123/alice@company.com
    arn = event.get("requestContext", {}).get("identity", {}).get("caller", "")
    session_name = arn.split("/")[-1] if "/" in arn else ""
    if "@" in session_name:
        return session_name
    
    return None
```

**3. Credential process:** Use SigV4 signing for IDC profiles (AWS credentials already available), JWT for OIDC profiles:

```python
if self.has_id_token():
    response = requests.get(url, headers={"Authorization": f"Bearer {token}"})
else:
    # Sign with AWS credentials (IDC users)
    response = requests.get(url, auth=sigv4_auth)
```

## Scope

- ~15 lines in `quota_check/index.py`
- ~15 lines in `credential_provider/__main__.py`  
- ~5 lines in `quota-monitoring.yaml` (route auth type)

## Impact

- IDC users get the same per-user quota enforcement as OIDC users
- No new infrastructure deployed
- Backward compatible: OIDC path unchanged

wirjo:
## Problem

Three critical bugs in the IAM Identity Center credential path make it unusable for enterprise customers:

1. **credential-process hangs on every prompt** — passthrough mode uses ambient credential chain with 30s timeout. If SSO session is expired, the SDK hangs attempting to refresh, blocking Claude Code.

2. **Quota enforcement doesn't work** — the passthrough path never calls `quota.Check()`. Even with quota stack deployed and limits exceeded, IDC users are never blocked.

3. **OTEL helper hangs on startup** — `--get-monitoring-token` flag dispatch runs AFTER the passthrough early-exit. IDC users can never reach the monitoring token path.

Closes #401

### Active SSO Auth (new `idc.go`)
Instead of relying on ambient credentials from a prior `aws sso login`, the credential-process now drives the full SSO OIDC device-authorization flow internally using the AWS SDK's `ssocreds` package:
- Checks `~/.aws/sso/cache/` for valid token
- If expired → opens browser for SSO device auth (exactly like `aws sso login`)
- No manual step required — seamless like the OIDC path

### Flag Dispatch Before Auth (`main.go` restructure)
All flags (`--get-monitoring-token`, `--show-tags`, `--check-expiration`, `--get-tag`, `--clear-cache`) now dispatch BEFORE auth-type branching. This fixes the otel-helper hang.

### Quota + OTEL in Both Paths
Both IDC and legacy passthrough now:
- Call `quota.Check()` with SigV4 auth (via `CheckWithIAM`)
- Call `writeOtelCacheFromSTS()` for per-user attribution

## Architecture

```
credential-process --profile X
  ├── Flag dispatch (--get-monitoring-token, etc.) ← runs for ALL auth types
  ├── cfg.IsIDC() → runIDC() (active SSO via ssocreds SDK)
  ├── !cfg.IsSsoEnabled() → runPassthrough() (legacy ambient chain, 10s timeout)
  └── else → run() (OIDC path, unchanged)
```

## Changes

| File | Change |
|------|--------|
| `source/go/internal/config/config.go` | Add IDC fields (`auth_type`, `idc_start_url`, `idc_account_id`, `idc_permission_set_name`, `idc_region`) + `IsIDC()` helper |
| `source/go/cmd/credential-process/main.go` | Restructure: flags first, then IDC → passthrough → OIDC |
| `source/go/cmd/credential-process/idc.go` (new) | Active SSO auth via `ssocreds` + quota + OTEL |
| `source/go/cmd/credential-process/passthrough.go` | Add quota check + OTEL write, reduce timeout 30s→10s |
| `source/go/go.mod` | Promote `credentials`, `sso`, `ssooidc` to direct deps |
| `source/claude_code_with_bedrock/cli/commands/package.py` | Write IDC fields to config.json |

## Backward Compatibility

| Scenario | Behavior |
|----------|----------|
| Existing OIDC deployments | **No change** — `run()` path untouched |
| Existing IDC (no `idc_start_url` in config.json) | Legacy passthrough (now with quota + OTEL) |
| New IDC deployments (`auth_type=idc` + IDC fields) | Full active SSO experience |

## Requirements for IDC quota attribution

- IDC permission set must use email as the role session name (default for most Identity Center setups)
- The email is extracted from the assumed-role ARN: `arn:aws:sts::ACCOUNT:assumed-role/ROLE/user@company.com`

## Testing

- [ ] Go compilation verified (requires Go 1.24 — CI will validate)
- [ ] Existing OIDC tests unaffected (no changes to `run()` path)
- [ ] Existing passthrough tests: passthrough still works, now with quota/OTEL additions
- [ ] IDC-specific tests to be added in follow-up

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -2135,7 +2135,18 @@ RUN pyinstaller \
 
         # Add the appropriate federation field based on type
         if not sso_enabled:
-            pass  # No OIDC/Cognito fields needed — credential-process uses ambient chain
+            # IDC path: include IDC fields so credential-process can drive SSO auth.
+            auth_type = getattr(profile, "auth_type", None)
+            if auth_type == "idc":
+                config[profile_name]["auth_type"] = "idc"
+                if getattr(profile, "idc_start_url", None):
+                    config[profile_name]["idc_start_url"] = profile.idc_start_url
+                if getattr(profile, "idc_account_id", None):
+                    config[profile_name]["idc_account_id"] = profile.idc_account_id
+                if getattr(profile, "idc_permission_set_name", None):
+                    config[profile_name]["idc_permission_set_name"] = profile.idc_permission_set_name
+                idc_region = getattr(profile, "idc_region", None) or profile.aws_region
+                config[profile_name]["idc_region"] = idc_region
         elif federation_type == "direct":
             config[profile_name]["federated_role_arn"] = federation_identifier
             config[profile_name]["federation_type"] = "direct"
--- a/source/credential_provider/__main__.py
+++ b/source/credential_provider/__main__.py
@@ -2229,6 +2229,18 @@ class MultiProviderAuth:
                   "Check your AWS CLI configuration or run 'aws sso login'.", file=sys.stderr)
             return 1
 
+        # Quota enforcement (SigV4-signed — email resolved from IAM ARN server-side).
+        if self._should_check_quota():
+            self._debug_print("IDC/passthrough: performing SigV4 quota check")
+            quota_result = self._check_quota(token_claims={}, id_token=None)
+            if not quota_result.get("allowed", True):
+                reason = quota_result.get("reason", "quota exceeded")
+                message = quota_result.get("message", "Usage quota exceeded. Contact your administrator.")
+                print(f"Error: {message}", file=sys.stderr)
+                print(f"Reason: {reason}", file=sys.stderr)
+                return 1
+            self._save_quota_check_timestamp()
+
         output = {
             "Version": 1,
             "AccessKeyId": frozen.access_key,
--- a/source/go/cmd/credential-process/idc.go
+++ b/source/go/cmd/credential-process/idc.go
@@ -51,8 +51,18 @@ func (a *credentialApp) runIDC() int {
 	roleName := a.cfg.IDCPermissionSetName
 
 	if startURL == "" || accountID == "" || roleName == "" {
-		fmt.Fprintln(os.Stderr, "Error: incomplete IDC configuration.")
-		fmt.Fprintln(os.Stderr, "Required fields in config.json: idc_start_url, idc_account_id, idc_permission_set_name")
+		fmt.Fprintln(os.Stderr, "Error: incomplete IDC configuration in config.json.")
+		if startURL == "" {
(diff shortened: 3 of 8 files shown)


wirjo:
## Testing Instructions

### Quick regression check (OIDC must not break):
```bash
cd source/go && go test ./...
cd source && python -m pytest tests/ -v
./credential-process --profile <oidc-profile>  # should work as before
./credential-process --profile <oidc-profile> --get-monitoring-token
```

### IDC path (new):
```bash
# After ccwb init + ccwb package with IDC:
./credential-process --profile <idc-profile>
# → Browser opens for SSO auth, outputs credential JSON

# Subsequent calls use cached token (no browser):
./credential-process --profile <idc-profile>

# Flags work for IDC now:
./credential-process --profile <idc-profile> --get-monitoring-token
./credential-process --profile <idc-profile> --show-tags
./credential-process --profile <idc-profile> --clear-cache

# Quota enforcement (if stack deployed, set limit to 0):
./credential-process --profile <idc-profile>
# → Should output "Error: quota exceeded"
```

### Legacy passthrough (sso_enabled=false, no idc_start_url):
```bash
aws sso login --profile <profile>  # manual auth first
./credential-process --profile <legacy>  # outputs creds + runs quota check
# With expired session → errors within 10s (not 30s hang)
```

If all regression + IDC tests pass, this is safe to merge to `feat/aws-iam-idc-support`.

PR #540, opened 2026-06-12
fix(go): try refresh_token before browser auth in --get-monitoring-token

HristoStoyanovYotpo:
## Why

Every Claude Code prompt opens 1-2 browser tabs to re-authenticate, even when the keyring/session has a valid refresh_token capable of silent renewal. This breaks the user experience for terminal users and silently fails for **Cowork 3P (Claude Desktop)** where browser popups aren't possible — the exact scenario [PR #447](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/447) was designed to fix.

### Root cause

Two paths invoke `credential-process` per Bedrock call:
1. AWS SDK → `credential_process` profile resolution → main `run()` flow
2. Claude Code's `otelHeadersHelper` → `otel-helper` → `credential-process --get-monitoring-token`

Path 1 calls `trySilentRefresh()` and `tryRefreshToken()` before falling back to a browser. Path 2 (`getMonitoringToken()`) skipped both and went straight to `authenticate()` (browser popup) when the cached monitoring token had less than 10 min remaining (per the buffer in `storage.GetMonitoringToken` at [`monitoring.go:31`](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/beta/source/go/internal/storage/monitoring.go#L31)).

### Why it surfaced now

The asymmetric buffers compound the problem:
- Monitoring token: `<= 600` seconds → considered stale 10 min before expiry
- AWS credentials: `<= 30` seconds → considered valid until 30 sec before expiry

So there's a ~9.5-minute window where AWS creds are still valid (path 1 silent) but the monitoring token is "stale" (path 2 → browser popup).

The recent removal of sidecar-mode caching (`otelcol` + `otel-helper.sh`) made every OTEL export hit `credential-process --get-monitoring-token` directly, exposing this gap on every prompt instead of every few hours.

This contradicts PR #447's stated goal:

> Enable credential-process to silently refresh AWS credentials after the OIDC id_token expires, without requiring browser interaction. This is critical for Cowork 3P (Claude Desktop via Bedrock) where no browser popup is possible.

## What

In `getMonitoringToken()`, attempt `tryRefreshToken()` before falling back to browser-based `authenticate()`. Mirrors the order already used by the main `run()` flow.

`trySilentRefresh()` is intentionally **not** called here — its first step is `storage.GetMonitoringToken()`, which we just observed returned empty. Only `tryRefreshToken()` (which uses the separately-stored `refresh_token`) is meaningful.

## Test plan

- [x] **New unit test** `TestGetMonitoringToken_RefreshTokenWiring`: with no refresh_token stored, `tryRefreshToken` returns nil cheaply, preserving the fall-through to browser auth for the genuinely-no-credentials case.
- [x] **New unit test** `TestGetMonitoringToken_RefreshTokenStoredCallsExchange`: with a stored refresh_token and unreachable IdP, `tryRefreshToken` is attempted (failure clears the refresh_token per the existing contract — proves the wiring exists).
- [x] All existing tests pass: `go test ./...` (full Go module), targeted Python tests `test_silent_refresh.py`, `test_credential_process_contract.py`, `test_monitoring_keyring_chunks.py`.
- [x] Manual verification on macOS: with valid keyring entries, `~/claude-code-with-bedrock/credential-process --profile <name> --get-monitoring-token` now returns a token without opening a browser.

## Tier-1 review checklist (per [`.claude/rules/review-tiers.md`](.claude/rules/review-tiers.md))

This change touches `source/go/cmd/credential-process/main.go` (Tier 1 — all users affected if broken).

- [x] Regression test covers the changed code path
- [x] Backward compat: behavior unchanged when no refresh_token is stored (still falls through to browser auth)
- [x] Cross-platform CI must pass (no admin merge override) — tests use `t.Setenv` and handle Windows `USERPROFILE`
- [ ] Parity check: no Python equivalent — `--get-monitoring-token` is a Go-only flag added in PR #441 era
- [x] Test with all auth types: change is gated on `refresh_token` presence; `idc`/`none` profiles never reach this path (they use `runPassthrough`)

## Open questions for reviewers

1. **Should the 10-min buffer in `storage.GetMonitoringToken` be reduced** to align with the 30s buffer used for AWS credentials? With this PR the consequence is "do a refresh_token exchange instead of returning the still-valid token" — cheap (~200ms) but unnecessary work. Worth a follow-up.

2. **Re-reading the token after `tryRefreshToken()`** is slightly awkward. Cleaner alternative: change `tryRefreshToken()` to also return the id_token, or extract the saved-token-output logic into a helper. Open to reviewer preference.

## Related

- Completes PR #447 (refresh_token persistence) for the monitoring-token path
- Fixes the regression introduced when sidecar mode (`otelcol` + `otel-helper.sh`) was removed

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -218,7 +218,27 @@ func (a *credentialApp) getMonitoringToken() int {
 		return 0
 	}
 
-	// No cached token — trigger authentication
+	// Cached monitoring token expired or near-expiry. Try refresh_token
+	// exchange (PR #447) before opening a browser — this is the only path
+	// that works for Cowork 3P (no browser popup possible) and eliminates
+	// per-prompt auth interruptions for terminal users whose monitoring
+	// token has aged past the 10-min buffer in storage.GetMonitoringToken
+	// while their refresh_token is still valid (typically 7-30 days).
+	//
+	// Note: trySilentRefresh() is intentionally not attempted here — its
+	// first step is storage.GetMonitoringToken(), which we just observed
+	// returns empty. Only refresh_token (separately stored) is meaningful.
+	if creds := a.tryRefreshToken(); creds != nil {
+		_ = creds // tryRefreshToken already saved AWS creds and the fresh monitoring token
+		if t, terr := storage.GetMonitoringToken(a.profile, a.cfg.CredentialStorage); terr == nil && t != "" {
+			fmt.Println(t)
+			return 0
+		}
+		debugPrint("refresh_token exchange succeeded but monitoring token unreadable; falling through to browser auth")
+	}
+
+	// No refresh_token available, refresh failed, or refresh produced no
+	// readable monitoring token — fall back to browser authentication.
 	debugPrint("No valid monitoring token found, triggering authentication...")
 	authResult, err := a.authenticate()
 	if err != nil {
--- a/source/go/cmd/credential-process/monitoring_refresh_test.go
+++ b/source/go/cmd/credential-process/monitoring_refresh_test.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"os"
+	"testing"
+
+	"ccwb-go/internal/config"
+	"ccwb-go/internal/storage"
+)
+
+// TestGetMonitoringToken_RefreshTokenWiring is the regression guard for
+// the bug where --get-monitoring-token would open a browser tab on every
+// expired-token call instead of using the cached refresh_token.
+//
+// PR #447 added refresh_token-based silent renewal but only wired it into
+// the main run() flow. The --get-monitoring-token path (called by
+// otel-helper for every OTEL export cycle) skipped tryRefreshToken() and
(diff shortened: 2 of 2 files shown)


wirjo:
## Review

**Risk: Low** — Adds a `tryRefreshToken()` call before the browser auth fallback in `--get-monitoring-token`.

**Why merge:** Without this, every Claude Code prompt opens 1-2 browser tabs for re-authentication when the monitoring token is near expiry — even though a valid refresh_token exists. For CoWork 3P (Claude Desktop), browser popups aren't possible at all, so telemetry silently breaks.

**Safety:**
- Falls through to existing browser auth if refresh fails (no behavioral change on failure)
- No-op if no refresh_token is stored (backward compatible with fresh installs)
- Clears revoked refresh_tokens on exchange failure (cleanup)
- 2 regression tests covering success + failure paths

**Note:** Tier 1 file (credential-process/main.go) but change is additive and well-guarded. CI pending — will verify go tests pass before merge.

PR #545, opened 2026-06-13
fix(ci): allowlist the otel-helper test fixture JWT for secret scanner

### LINKED ISSUE #544: Scanners / secrets-scanner CI gate fails: unaudited fixture JWT in source/go/cmd/otel-helper/main_test.go
## Describe the bug

The `Scanners / secrets-scanner` workflow fails on `beta` (and on PRs targeting it). The gate scans the repo against `.secrets.baseline`, then audits the result; a detected secret that is neither baselined nor allowlisted makes the audit error out.

## Reproducible steps

From a clean `beta` checkout, run the gate exactly as `.github/workflows/scanners.yml` does:

```bash
detect-secrets scan --baseline .secrets.baseline
cat .secrets.baseline | jq '[.results|to_entries|.[].value[]
  | { "filename": .filename, "is_secret": .is_secret }
  | select(.filename | startswith("source/tests/") | not)
  | if .is_secret == null or .is_secret == true then .filename else empty end]
  | unique | if length>0 then error("potential secrets in: \(.)") else empty end'
```

Result:

```
jq: error (at <stdin>:NNN): potential secrets in: ["source/go/cmd/otel-helper/main_test.go"]
```

## Root cause

`source/go/cmd/otel-helper/main_test.go:332` holds a fixture JWT used to exercise the `CLAUDE_CODE_MONITORING_TOKEN` path. It is an inert test token — `alg:none`, no signature, payload `{"email":"test@example.com","exp":9999999999}`. `detect-secrets` flags it on pattern, but the line is neither recorded in `.secrets.baseline` nor marked with an inline allowlist pragma, so it surfaces as an unaudited finding (`is_secret: null`).

The audit's `source/tests/` exemption doesn't apply because the file lives under `source/go/`. The fixture was added in #526; no allowlist marker accompanied it.

## Impact / Evidence

- `Scanners / secrets-scanner` is red on `beta` and on every PR targeting it, so the signal is meaningless until cleared and would mask a real future finding.
- Other fixtures in the repo are already allowlisted inline (e.g. `source/tests/test_credential_passthrough.py`, `source/go/internal/oidc/confidential.go` use `# pragma: allowlist secret` / `// pragma: allowlist secret`) — this one was missed.

## Version / environment

- Branch: `beta` (HEAD `3bbce7d`). Also affects `main` after promotion.
- `detect-secrets` 1.5.0.

## Why CI missed it

The gate runs on `pull_request`, but fork PRs skip secret-bearing workflows, so the failing state reached `beta` without a blocking signal on the introducing PR.

peepeepopapapeepeepo:
## Description

Marks the inert fixture JWT in `source/go/cmd/otel-helper/main_test.go` with an inline `// pragma: allowlist secret` comment so `detect-secrets` skips it, clearing the `Scanners / secrets-scanner` gate.

## Why is this change needed

The secrets-scanner gate is red on `beta`: detect-secrets flags the fixture JWT at `main_test.go:332`, and because the line is neither baselined nor allowlisted, the workflow's jq audit errors with `potential secrets in: ["source/go/cmd/otel-helper/main_test.go"]`. The file is under `source/go/`, outside the audit's `source/tests/` exemption.

Fixes #544

## Type of Change

- [x] CI/CD changes

## Changes Made

- Add `// pragma: allowlist secret` to the fixture JWT line in `main_test.go`.
- Matches the repo's existing convention (`test_credential_passthrough.py`, `confidential.go`) — preferred over a `.secrets.baseline` entry: it's local to the fixture, visible to reviewers, and survives line-number changes.

## Additional Notes

The token is an inert fixture — `alg:none`, no signature, payload `{"email":"test@example.com","exp":9999999999}` — used to drive the `CLAUDE_CODE_MONITORING_TOKEN` path. No real credential value.

### What was tested

- [x] Ran the workflow gate locally (`detect-secrets scan --baseline .secrets.baseline` + the jq audit): errors on `beta`, exits 0 with this change; `.secrets.baseline` needs no edit.
- [x] `cd source/go && gofmt -l cmd/otel-helper/main_test.go` clean; `go vet` clean.
- [x] `go test -count=1 ./cmd/otel-helper/` -> `ok` (fixture still exercises its path).
- [x] `cd source && poetry run pytest tests/` — full suite green (950 passed).

### DIFF
--- a/source/go/cmd/otel-helper/main_test.go
+++ b/source/go/cmd/otel-helper/main_test.go
@@ -329,7 +329,7 @@ func TestRun_CacheHit_ResolvesTokenFromEnv(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	envToken := "eyJhbGciOiJub25lIn0.eyJlbWFpbCI6InRlc3RAZXhhbXBsZS5jb20iLCJleHAiOjk5OTk5OTk5OTl9."
+	envToken := "eyJhbGciOiJub25lIn0.eyJlbWFpbCI6InRlc3RAZXhhbXBsZS5jb20iLCJleHAiOjk5OTk5OTk5OTl9." // pragma: allowlist secret
 	t.Setenv("CLAUDE_CODE_MONITORING_TOKEN", envToken)
 
 	r, w, _ := os.Pipe()


wirjo:
## Review

**Risk: Very Low** — 1-line comment addition to allowlist a test fixture JWT.

**Why merge:** The secrets-scanner CI gate is red on beta because `detect-secrets` flags the fixture JWT in `otel-helper/main_test.go`. This blocks all PRs from showing clean CI. The JWT is inert (expired, self-signed, used only in unit tests).

**Pattern:** Matches the repo's existing convention — other test files already use `pragma: allowlist secret` for similar fixtures.

Approving and merging to unblock CI for all open PRs.

PR #547, opened 2026-06-13
fix(init): use API-returned ARN for distribution IdP secret

### LINKED ISSUE #546: ccwb deploy fails on landing-page distribution: "Secrets Manager can't find the specified secret" when IdP secret already exists
## Describe the bug

Deploying the landing-page distribution stack fails at the ALB HTTPS listener with `ResourceNotFoundException` from Secrets Manager, whenever the IdP client secret already exists in Secrets Manager (i.e. on any re-run of `ccwb init` for an existing profile).

```
AWS::ElasticLoadBalancingV2::Listener (HTTPSListener):
Secrets Manager can't find the specified secret.
(Service: AWSSecretsManager; Status Code: 400; Error Code: ResourceNotFoundException)
```

## Reproducible steps

1. `ccwb init` with a landing-page distribution + a confidential IdP (Okta/Azure/Auth0), so a client secret is stored in Secrets Manager. Deploy — succeeds (first run).
2. Re-run `ccwb init` for the same profile (e.g. to change any setting). The secret now already exists, so init takes the `update_secret` path.
3. `ccwb deploy` -> distribution stack fails on `HTTPSListener` with the error above.

## Root cause

`source/claude_code_with_bedrock/cli/commands/init.py` stores the secret, then records its ARN. On first creation it uses the ARN from the API response (correct). But on the `ResourceExistsException` / `update_secret` path it **hand-builds** the ARN (init.py:1367, and again in the outer except at :1374):

```python
secret_arn = f"arn:aws:secretsmanager:{region}:{account_id}:secret:{secret_name}"
```

Secrets Manager appends a random 6-char suffix to every secret ARN (e.g. `...-FhLi4n`). The hand-built string omits it. That ARN is saved to the profile, passed to the template by deploy.py (`AzureClientSecretArn={profile.distribution_idp_client_secret_arn}`), and resolved by the listener via `{{resolve:secretsmanager:${AzureClientSecretArn}:SecretString}}` (landing-page-distribution.yaml:946) — which requires the full ARN, so the lookup fails.

## Impact / Evidence

- Blocks `ccwb deploy` for landing-page distribution on any profile whose IdP secret already exists — i.e. every deploy after the first `init`. Observed live (Azure IdP, `ap-southeast-1`).
- First-ever deploys are unaffected (the create path returns the correct ARN), which is why this only surfaces on re-init.

## Version / environment

- Branch `beta` (HEAD `3bbce7d`); present since the landing-page feature (#64). Also affects `main`.

## Why CI missed it

No test exercised the `update_secret` (existing-secret) path; the secret-storage logic was inline in the `init` wizard with no unit-testable seam.

peepeepopapapeepeepo:
## Description

Use the ARN returned by the Secrets Manager API (which includes its random suffix) for the distribution IdP client secret, on both the create and update paths, instead of hand-building it.

## Why is this change needed

When the IdP secret already exists, `init` took the `update_secret` path and then hand-built the secret ARN as `arn:...:secret:<name>`, omitting the random 6-char suffix AWS appends. The landing-page ALB HTTPS listener resolves the secret via `{{resolve:secretsmanager:<arn>}}`, which needs the full ARN, so `ccwb deploy` failed with `ResourceNotFoundException`. First-ever deploys worked (create path returns the correct ARN); this surfaced on any re-`init`.

Fixes #546

## Type of Change

## Changes Made

- Extract secret create/update into `_store_idp_secret()`, which always returns `response["ARN"]` from the API (create and update alike).
- The outer `except` now recovers the real ARN via `describe_secret` before falling back to a hand-built one.
- Add `TestStoreIdpSecret` covering create and update paths.

## Additional Notes

Storage (keyring vs Secrets Manager) and which IdPs use a confidential secret are unchanged — only the recorded ARN is corrected.

### What was tested

- [x] `cd source && poetry run pytest tests/` — full suite green (935 passed).
- [x] New `TestStoreIdpSecret` fails against the prior hand-built-ARN behavior and passes on this change (falsified).
- [x] Verified locally: bandit (no new findings vs beta), semgrep (0), detect-secrets (0).
- [x] CI gates: pytest-ci + scanners apply; go-tests / cfn_nag / validate-regions are N/A (no `source/go/**`, no CloudFormation, no region data in this diff).

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1349,29 +1349,26 @@ class InitCommand(Command):
 
                     secret_name = f"{config['aws']['identity_pool_name']}-distribution-idp-secret"
 
-                    # Try to create or update secret
-                    try:
-                        secret_response = secrets_client.create_secret(
-                            Name=secret_name,
-                            SecretString=idp_client_secret,
-                            Description=f"IdP client secret for "
-                            f"{config['aws']['identity_pool_name']} distribution landing page",
-                        )
-                        secret_arn = secret_response["ARN"]
-                    except secrets_client.exceptions.ResourceExistsException:
-                        # Secret already exists, update it
-                        secret_response = secrets_client.update_secret(
-                            SecretId=secret_name,
-                            SecretString=idp_client_secret,
-                        )
-                        secret_arn = f"arn:aws:secretsmanager:{region}:{account_id}:secret:{secret_name}"
+                    secret_arn = self._store_idp_secret(
+                        secrets_client,
+                        secret_name,
+                        idp_client_secret,
+                        description=f"IdP client secret for "
+                        f"{config['aws']['identity_pool_name']} distribution landing page",
+                    )
 
                     console.print(f"[green]✓[/green] IdP client secret stored in Secrets Manager: {secret_name}")
 
                 except Exception as e:
                     console.print(f"[red]Error storing secret in Secrets Manager: {e}[/red]")
                     console.print("[yellow]You'll need to configure the secret manually before deployment[/yellow]")
-                    secret_arn = f"arn:aws:secretsmanager:{region}:{account_id}:secret:{secret_name}"
+                    # Storing failed, so we have no API response. Recover the real
+                    # ARN (with its random suffix) via describe_secret if the secret
+                    # exists; only fall back to a hand-built ARN as a last resort.
+                    try:
+                        secret_arn = secrets_client.describe_secret(SecretId=secret_name)["ARN"]
+                    except Exception:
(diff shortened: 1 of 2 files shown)


wirjo:
## Review

**Risk: Very Low** — Extracts secret storage into a helper that uses the API-returned ARN (with random suffix) instead of hand-building it.

**Why merge:** `ccwb deploy` is broken for anyone who re-runs `ccwb init`. The hand-built ARN omits the 6-char random suffix AWS appends to Secrets Manager ARNs, causing `ResourceNotFoundException` when the ALB tries to resolve the secret. First-time deploys work (create path returns correct ARN); the bug surfaces on any subsequent init.

**Safety:**
- Fallback path preserved: if API fails, tries `describe_secret` → hand-built ARN
- Regression tests cover both create and update paths
- No config changes needed from users — fix is transparent

Production blocker fix. Approving and merging.

PR #554, opened 2026-06-14
feat(codebuild): cross-region CodeBuild deployment for unsupported regions

peepeepopapapeepeepo:
## Description

Adds opt-in cross-region CodeBuild: when the main region lacks the Windows Server 2022 container fleet, the user picks a supported CodeBuild region in `ccwb init`, and `deploy` deploys the CodeBuild stack there while the rest of the infrastructure stays in the main region. Every CLI command that touches CodeBuild resolves the region through one helper so the client, stack outputs, and S3 bucket never split across regions.

Proposes #553 (maintainer decision requested — see issue).

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Changes Made

- **config**: `codebuild_region` (chosen build region) + `codebuild_prior_regions` (regions abandoned via re-init/disable, for cleanup).
- **helpers**: `CODEBUILD_WINDOWS_REGIONS` + `find_nearest_codebuild_region()` (continent-aware; us-east-1 fallback when no in-continent region).
- **init**: prompt for a supported region when the main one is unsupported; warn explicitly when the nearest crosses continents (data residency); record abandoned regions on a region change AND when CodeBuild is skipped/disabled.
- **deploy**: non-interactive; deploys the codebuild stack with its own region-bound manager; orphan detection + deletion and `--show-commands` also use the codebuild region; legacy unsupported profiles skip with a pointer to init rather than failing the whole deploy.
- **builds / package / package_cb / distribute**: route codebuild client + stack + S3 bucket through `get_codebuild_region` (distribute downloads the Windows artifact from the build region).
- **destroy**: delete the codebuild stack in its own region, plus any prior regions (ungated from `enable_codebuild`), so a cross-region/moved/disabled stack is never orphaned.

## Additional Notes

Opt-in and backward-compatible: `codebuild_region` defaults to `None` -> resolves to `aws_region`, so single-region profiles are byte-identical. >300 lines because it's one cohesive feature spanning the init->deploy->package->builds->distribute->destroy lifecycle; splitting the field from its usage would land half-wired intermediate states.

### What was tested

**Verified live on real AWS** — full lifecycle, main infra `ap-southeast-1`, CodeBuild `ap-southeast-2` (the find_nearest_codebuild_region pick for apse1):
- [x] `ccwb init` region picker -> offered "ap-southeast-2 (nearest supported)", saved codebuild_region=ap-southeast-2
[...]

PR #568, opened 2026-06-16
Fix invalid YAML in Windows CodeBuild buildspec

jcjorel:
## Problem

Every Windows binary build via AWS CodeBuild fails in the `DOWNLOAD_SOURCE` phase (before Nuitka runs) with:

```
Phase context status code: YAML_FILE_ERROR
Message: Expected Commands[2] to be of string type: found subkeys instead
at line 40, value of the key 'tag' on line 39 might be empty
```

This blocks the CodeBuild Windows build path entirely. The Linux x64 / ARM64 CodeBuild builds are unaffected.

## Root cause

In `deployment/infrastructure/codebuild-windows.yaml`, the Windows project's inline buildspec `post_build.commands` has two **unquoted** commands containing `"FATAL: ` — a colon followed by a space:

```yaml
- if (-not (Test-Path credential-process-windows.exe)) { Write-Error "FATAL: credential-process-windows.exe was not built. Nuitka compilation failed."; exit 1 }
- if (-not (Test-Path otel-helper-windows.exe)) { Write-Error "FATAL: otel-helper-windows.exe was not built. Nuitka compilation failed. Consider using 'ccwb package --go' for reliable Go-based builds."; exit 1 }
```

In YAML, `key: value` (colon + space) denotes a mapping. Because these list items are unquoted plain scalars, the parser interprets the `: ` as a mapping separator and reads the command as a nested mapping instead of a string — which is exactly the "found subkeys instead" error CodeBuild reports.

## Fix

Wrap each command in single quotes so the whole string (including `FATAL: `) is a literal scalar. The second command's inner `'ccwb package --go'` single quotes are doubled (`''ccwb package --go''`) per YAML single-quote escaping rules. This is a **quoting-only** change — command behavior is unchanged.

## Verification

Extracted the corrected buildspec block and parsed it with a YAML loader: the `phases` (install / pre_build / build / post_build) parse correctly and all five `post_build` commands are now string scalars (previously the parse aborted).

## When this bites

Reproducible whenever the CodeBuild + Nuitka path is the only route to a Windows binary — e.g. the `ccwb package --go` (Go cross-compile) alternative isn't available because Go can't be installed on the build host. Encountered while deploying with infrastructure in `eu-central-1`.

### DIFF
--- a/deployment/infrastructure/codebuild-windows.yaml
+++ b/deployment/infrastructure/codebuild-windows.yaml
@@ -180,8 +180,8 @@ Resources:
               commands:
                 - echo Build completed
                 - echo Validating all required binaries exist...
-                - if (-not (Test-Path credential-process-windows.exe)) { Write-Error "FATAL: credential-process-windows.exe was not built. Nuitka compilation failed."; exit 1 }
-                - if (-not (Test-Path otel-helper-windows.exe)) { Write-Error "FATAL: otel-helper-windows.exe was not built. Nuitka compilation failed. Consider using 'ccwb package --go' for reliable Go-based builds."; exit 1 }
+                - 'if (-not (Test-Path credential-process-windows.exe)) { Write-Error "FATAL: credential-process-windows.exe was not built. Nuitka compilation failed."; exit 1 }'
+                - 'if (-not (Test-Path otel-helper-windows.exe)) { Write-Error "FATAL: otel-helper-windows.exe was not built. Nuitka compilation failed. Consider using ''ccwb package --go'' for reliable Go-based builds."; exit 1 }'
                 - Get-ChildItem *.exe | ForEach-Object { Write-Host "$($_.Name) - $([math]::Round($_.Length/1MB, 1)) MB" }
 
           artifacts:


wirjo:
Thanks @jcjorel — good catch. This is a real build blocker and the root cause analysis is excellent.

I've retargeted this PR to `beta` (our merge strategy: beta first, promote to main via releases).

## Regression Risk: None

This is a **quoting-only** change — the PowerShell commands are byte-identical at runtime. The YAML parser now correctly reads them as string scalars instead of choking. There is no scenario where this makes a previously-working build fail.

## Backward Compatibility: Full

- No config changes, no API changes, no IAM changes
- The fix is purely syntactic — same commands, just properly quoted
- Linux builds are unaffected (different buildspec)

## Simplification Suggestion

The fix is already minimal (2 lines), so there's nothing to simplify in the quoting itself. However, one small improvement for maintainability:

These PowerShell validation commands are getting long. A future cleanup (not for this PR) could extract them into a shared validation script in the repo and call it from the buildspec:

```yaml
# Future: single command, no quoting issues
- .\scripts\validate-windows-binaries.ps1
```

But that's a separate concern — this PR should land as-is since it's a minimal, correct fix for a hard failure.

## Verdict

✅ Ready for merge once CI passes. Very low risk, zero regression potential, full backward compatibility.

wirjo:
Thanks @jcjorel — great catch and excellent root cause analysis! 

However, this fix was already included in #554 (merged to beta Jun 14) which reworked the Windows CodeBuild buildspec as part of the cross-region CodeBuild feature. The `beta` branch already has the quoted commands (with an explanatory comment even).

Closing as already fixed on beta. The fix will reach `main` in the next release promotion. Appreciate you taking the time to diagnose this — the analysis is spot on. 🙏

Issue #541, opened 2026-06-12
CoWork 3P telemetry never reaches the dashboard or ccwb quota usage (central monitoring mode)


adiospeds:
## Why

We deployed ccwb in central monitoring mode (OIDC/Azure, custom-domain OTEL collector) and enabled CoWork 3P. The shipped `ClaudeCoWorkDashboard` stays **completely empty**, and CoWork usage never appears in `ccwb quota usage <email>` — even though CoWork is actively used and Claude Code telemetry on the *same* deployment works perfectly.

After tracing it end-to-end (config → ALB → collector → CloudWatch → dashboard/quota) **and capturing CoWork's real OTLP output on the wire**, the CoWork monitoring path turns out to be non-functional as designed. This is a feature gap, not a single-deployment misconfig: every piece between CoWork and the dashboard expects a data shape CoWork does not emit.

## Environment

- ccwb central monitoring mode, OIDC auth (Azure), custom-domain collector (`https://telemetry.redacted-domain.com`)
- CoWork (Claude Desktop) **v1.12603.1**, 3P/Bedrock, macOS arm64
- Claude Code telemetry on the same stack works (real `user.email`, full EMF in `/aws/claude-code/metrics`, namespace `ClaudeCode`)

## What I found

I stood up a throwaway local OTLP receiver, repointed CoWork's MDM `otlpEndpoint` at it (no AWS changes), restarted Claude, and drove ~7 turns. Findings:

### 1. CoWork emits the **logs/events** signal only — never metrics
18 POSTs to `/v1/logs`, **0** to `/v1/metrics`. OTLP scope: `claude-desktop.events`.
But the CoWork dashboard (`deployment/infrastructure/cowork-dashboard.yaml`) and the quota Lambda (`lambda-functions/quota_monitor/index.py`) both query the **metric** `claude_code.token.usage` via PromQL. There is no such metric for CoWork → queries never match.

### 2. The collector has **no `logs` pipeline**
`deployment/infrastructure/otel-collector.yaml` defines only:
```yaml
service:
  pipelines:
    metrics:
      receivers: [otlp]
      processors: [attributes, resource, batch/metrics]
      exporters: [otlphttp, awsemf]   # otlphttp only in the no-analytics variant
```
[...]

PR #570, opened 2026-06-16
fix(monitoring): force ECS redeploy after collector config change

wirjo:
## Why

The OTEL collector config is stored in SSM and resolved once at container start. When `ccwb deploy --stack monitoring` updates the config, the ECS task keeps serving the stale version because the task definition is byte-identical (references the SSM param by name, not version). Users get 404 on `/v1/logs` until they manually force a redeploy.

## What

- After successful monitoring stack deploy, call `ecs:UpdateService` with `forceNewDeployment=true`
- Non-fatal: if the ECS API call fails (e.g. first deploy where service doesn't exist yet), print a warning with the manual command
- **Tests**: 3 new tests covering success, failure-skip, and non-fatal error handling

## Backward Compatible

Adds an ECS API call after deploy — no config changes needed. First-time deploys where the ECS service doesn't exist yet gracefully degrade to a warning. Existing deployments benefit immediately (no more stale configs).

## Verification

```bash
cd source && poetry run pytest tests/cli/commands/test_deploy_ecs_redeploy.py -v
```

---

Credit: @adiospeds identified gap 3 during testing (#541)

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -766,6 +766,28 @@ class DeployCommand(Command):
                     template, stack_name, params, task_description="Deploying monitoring collector..."
                 )
 
+                # Force ECS service redeploy so the collector picks up the new config.
+                # The collector config is stored in SSM and resolved at container start;
+                # a CFN update alone won't restart the running task.
+                if result == 0:
+                    try:
+                        import boto3
+
+                        ecs_client = boto3.client("ecs", region_name=profile.aws_region)
+                        ecs_client.update_service(
+                            cluster="claude-code-otel-cluster",
+                            service="otel-collector-service",
+                            forceNewDeployment=True,
+                        )
+                        console.print("[dim]Forced ECS service redeploy to load new collector config[/dim]")
+                    except Exception as e:
+                        # Non-fatal: stack deployed fine, just couldn't force redeploy
+                        console.print(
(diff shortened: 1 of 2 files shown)


wirjo (commented):
## Self-Review

**Risk: Very Low** | **Tier:** 2 (deploy.py) | **Size:** +99/-0, 2 files

**Automated checks:** ✅ 3 tests pass locally

**Review:**
- **Blast radius:** All monitoring stack deploys now also trigger an ECS redeploy — this is beneficial (ensures fresh config) and the downside is a brief task restart (~30s service interruption during rolling deploy)
- **Backward compat:** ✅ Safe — ECS failure is caught and logged as a non-fatal warning. First-time deploys where the service doesn't exist yet gracefully degrade.
- **Test coverage:** ✅ 3 tests — success, failure-skip, non-fatal error
- **Simplification:** ✅ Minimal — one `try/except` block with a single `update_service` call
- **One concern:** ✅ Single purpose (force ECS redeploy)

**Potential concern:** The ECS service name (`otel-collector-service`) and cluster name (`claude-code-otel-cluster`) are hardcoded. These match the CFN template's `ServiceName` and `ClusterName` properties, so they're correct — but if someone customizes the stack name, the ECS call would fail (gracefully, due to try/except).

**Verdict: ✅ Ship it** — the non-fatal fallback means this can never break a deploy, only improve it.

---

@adiospeds — could you review this one? Specifically:
1. On your deployment, did the cluster/service names match `claude-code-otel-cluster` / `otel-collector-service`, or are they customized?
2. Is a ~30s rolling restart acceptable for your use case, or would you prefer a flag to opt out?

PR #571, opened 2026-06-16
fix(dashboard): support both CoWork telemetry schemas in metric filters

wirjo:
## Why

The CoWork dashboard stays empty even when events flow to `/aws/claude-cowork/events`. The metric filters target only the **Local Agent Mode** schema (`lam_session_turn_completed`), but real interactive CoWork usage emits the **Claude-Code-for-Desktop** schema (`claude_code.api_request`) with a fundamentally different structure:

- `body` is a flat string (`"claude_code.api_request"`), not an object
- Token/cost fields are under a top-level `attributes` object (sibling of `body`), not `body.attributes`
- Field names differ (`cache_read_tokens` vs `cache_read_input_tokens`, `cost_usd` vs `total_cost_usd`)

## What

- Add 6 metric filters for `claude_code.api_request` schema (pattern: `{ $.body = "claude_code.api_request" }`, values: `$.attributes.*`)
- Retain 6 `lam_session_turn_completed` filters (renamed with `Lam` prefix)
- Both schema filters emit to the **same** CloudWatch metric names (dashboard works regardless of mode)
- Fix Logs Insights query to use `coalesce()` for both schemas
- **Tests**: 7 new tests validating filter patterns, value paths, namespace, and cross-schema metric name parity

## Backward Compatible

Existing deployments with lam-mode events continue to work. The additional filters for the api_request schema are purely additive — CloudWatch Metric Filters with new names don't conflict. Template is idempotent on redeploy.

## Verification

```bash
cd source && poetry run pytest tests/e2e/test_cowork_dashboard_filters.py -v
cfn-lint deployment/infrastructure/cowork-dashboard.yaml
```

---

Credit: @adiospeds captured the wire format and identified the schema mismatch (#541). Their corrected YAML was used as reference.

### DIFF
--- a/deployment/infrastructure/cowork-dashboard.yaml
+++ b/deployment/infrastructure/cowork-dashboard.yaml
@@ -16,21 +16,25 @@ Parameters:
 
 Resources:
   # CloudWatch Metric Filters — extract metrics from CoWork log events.
-  # CoWork emits lam_session_turn_completed events with token counts and cost.
-  # These filters create CloudWatch metrics in the ClaudeCoWork namespace.
+  # CoWork emits events in two schemas depending on mode:
+  #   1. Local Agent Mode: body.name="lam_session_turn_completed", body.attributes.*
+  #   2. Claude-Code-for-Desktop: body="claude_code.api_request", attributes.*
+  # We create filters for both schemas so the dashboard works regardless of mode.
   # Note: AWS limits MetricTransformations to 1 per filter, so each metric
   # requires its own filter resource.
+
+  # --- Claude-Code-for-Desktop schema (body="claude_code.api_request") ---
 
   InputTokenMetricFilter:
     Type: AWS::Logs::MetricFilter
     Properties:
       LogGroupName: /aws/claude-cowork/events
       FilterName: cowork-input-tokens
-      FilterPattern: '{ $.body.name = "lam_session_turn_completed" }'
+      FilterPattern: '{ $.body = "claude_code.api_request" }'
       MetricTransformations:
         - MetricNamespace: ClaudeCoWork
           MetricName: token.usage.input
-          MetricValue: $.body.attributes.input_tokens
+          MetricValue: $.attributes.input_tokens
           DefaultValue: 0
 
   OutputTokenMetricFilter:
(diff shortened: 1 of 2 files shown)


wirjo (commented):
## Self-Review

**Risk: Low** | **Tier:** 2 (cowork-dashboard.yaml) | **Size:** +214/-13, 2 files

**Automated checks:** ✅ cfn-lint passes, ✅ 7 tests pass locally

**Review:**
- **Blast radius:** CoWork dashboard only — no impact on quota, auth, or core collector
- **Backward compat:** ✅ Safe — existing `lam_*` filters retained (renamed resources with `Lam` prefix). New `claude_code.api_request` filters are purely additive. Both emit to the same metric names so the dashboard works for either mode.
- **Test coverage:** ✅ 7 new tests — schema coverage, value path correctness, namespace parity, cross-schema metric name alignment
- **Simplification:** ✅ Each filter is one resource — no shared logic to break. Duplication is intentional (AWS requires 1 MetricTransformation per filter).
- **One concern:** ✅ Single purpose (fix metric filters for real CoWork schema)

**Dashboard JSON check:**
- ✅ No embedded JSON in this template (dashboard body uses `!Sub` with CloudWatch metric references, not raw JSON)
- ✅ Metric names match between both schema variants (verified by test)
- ✅ Logs Insights query uses `coalesce()` to handle both schemas in one view

**CFN resource rename note:** The original resources (`InputTokenMetricFilter`, etc.) now target the `api_request` schema. The `lam_*` schema gets new `Lam`-prefixed resource names (`LamInputTokenMetricFilter`, etc.) with different `FilterName` values. On redeploy, CFN will create the new resources and update the existing ones — no orphaned filters.

**Verdict: ✅ Ship it**

---

@adiospeds — this is the one where your wire captures are most valuable. Could you review:
1. Do the `$.attributes.cache_read_tokens` and `$.attributes.cache_creation_tokens` field names match what you saw? (vs `cache_read_input_tokens` / `cache_creation_input_tokens` in the lam schema)
2. Is `$.attributes.cost_usd` correct, or did you see `$.attributes.total_cost_usd`?
3. Does the `coalesce()` approach in the Logs Insights query work for your dashboard view?

PR #573, opened 2026-06-16
fix(monitoring): remove hardcoded names that break stack updates

spawnia:
## Why

Hardcoded resource names in `otel-collector.yaml` break stack updates when CloudFormation needs to replace resources — the old and new resources collide on the same name. Additionally, `ServiceName` on conditional ECS services causes `AlreadyExists` errors because CloudFormation creates the new resource before deleting the old.

## What

**Template fixes (`otel-collector.yaml`):**
- Remove `Name` from ALB — let CloudFormation auto-generate a unique name
- Remove `ServiceName` from both ECS services (conditional resources)
- Simplify HTTP listener to always forward (no conditional redirect needed without fixed naming)

**Deploy code (`deploy.py`):**
- Replace hardcoded `service="otel-collector-service"` in the force-redeploy step with dynamic service discovery via `list_services()`
- Handles first-deploy gracefully (no service exists yet)

**Regression test:**
- `test_cfn_naming_limits.py`: validates otel-collector.yaml has no hardcoded ALB Name or ECS ServiceName
- Updated `test_deploy_ecs_redeploy.py` to verify `list_services()` discovery pattern

## Migration

Existing deployments: delete the monitoring stack, then redeploy. The ALB name removal triggers replacement, which requires the old ALB to be gone first (target groups can only attach to one ALB).

```bash
aws cloudformation delete-stack --stack-name <profile>-monitoring
# wait for deletion
ccwb deploy monitoring
```

## Test plan

- [x] `test_otel_collector_no_hardcoded_names` passes
- [x] `test_deploy_ecs_redeploy.py` — 4 tests pass (discovery, first-deploy, failure scenarios)
- [x] Deployed to development environment (stack deleted + redeployed successfully)

🤖 Generated with Claude Code

### DIFF
--- a/deployment/infrastructure/otel-collector.yaml
+++ b/deployment/infrastructure/otel-collector.yaml
@@ -192,10 +192,6 @@ Resources:
   LoadBalancer:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
-      Name: !If
-        - HasCustomDomain
-        - otel-collector-https-alb
-        - otel-collector-alb
       Type: application
       IpAddressType: ipv4
       Scheme: internet-facing
@@ -230,8 +226,6 @@ Resources:
         HttpCode: '200,404'
 
   # Listeners
-  # IMPORTANT: Enabling HTTPS on an existing HTTP-only stack requires deletion and recreation
-  # due to CloudFormation's inability to atomically switch target groups between listeners
   HTTPListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
     Properties:
@@ -239,25 +233,12 @@ Resources:
       Port: 80
       Protocol: HTTP
       DefaultActions:
-        - Type: !If
-            - HasCustomDomain
-            - redirect
-            - forward
-          RedirectConfig: !If
-            - HasCustomDomain
-            - Protocol: HTTPS
-              Port: '443'
-              StatusCode: HTTP_301
-            - !Ref AWS::NoValue
-          TargetGroupArn: !If
-            - HasCustomDomain
-            - !Ref AWS::NoValue
-            - !Ref HTTPTargetGroup
+        - Type: forward
+          TargetGroupArn: !Ref HTTPTargetGroup
 
   HTTPSListener:
     Type: AWS::ElasticLoadBalancingV2::Listener
     Condition: HasCustomDomain
-    DependsOn: HTTPListener  # Ensure HTTP listener updates first to release target group
     Properties:
       LoadBalancerArn: !Ref LoadBalancer
       Port: 443
@@ -664,7 +645,7 @@ Resources:
     DependsOn:
       - HTTPListener
     Properties:
-      ServiceName: otel-collector-service
+
       Cluster: !Ref ECSCluster
       TaskDefinition: !Ref TaskDefinition
       LaunchType: FARGATE
@@ -690,7 +671,7 @@ Resources:
       - HTTPListener
       - HTTPSListener
     Properties:
-      ServiceName: otel-collector-service
+
       Cluster: !Ref ECSCluster
       TaskDefinition: !Ref TaskDefinition
       LaunchType: FARGATE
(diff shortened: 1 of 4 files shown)


wirjo:
Thanks @spawnia — the underlying problem (hardcoded names + conditional logical IDs causing `AlreadyExists` on stack update) is a real and known CloudFormation footgun. A few observations:

### Concern: ALB name change triggers replacement for existing HTTPS stacks

Changing `Name` from `otel-collector-https-alb` → `otel-collector-alb` on an already-deployed ALB will force CloudFormation to **replace** it (ALB names are create-and-replace). This gives existing HTTPS users the exact same cascade failure you're trying to fix — just from a different direction.

**Suggestion:** Remove the `Name` property entirely instead of changing it. CloudFormation will auto-generate a unique name and never hit this again. Existing stacks keep their current ALB (no replacement triggered by omitting a property that was previously set — CloudFormation only replaces when the value *changes*).

Actually, wait — removing `Name` on an existing stack *also* counts as a change. The safest approach is:
1. Remove `Name` only for **new** stacks (via a condition like `IsFirstDeploy`), OR  
2. Document this as a **one-time migration** that requires a brief ALB replacement window, OR
3. Use `DeletionPolicy: Retain` + a custom resource to handle the rename gracefully

### HTTP redirect removal

Agree the redirect isn't critical for programmatic OTEL clients, but worth noting in docs that HTTP telemetry endpoints are now forwarded unencrypted rather than redirected to HTTPS.

### Missing test coverage

The untested scenarios (fresh stack, HTTP-only → HTTPS toggle) are the highest-risk paths. Could you verify one of these before merge?

### Priority call

This is a real issue but affects a narrow window of users (upgrading from pre-split stacks). For current `beta` users who deployed recently, the ECS split is already in place. I'd suggest:
- Low priority for merge as-is
- Worth revisiting once the ALB replacement concern is addressed (either drop the Name entirely or document the one-time migration)

spawnia:
Thanks for the detailed review @wirjo — you're right that changing the Name triggers the same replacement for existing HTTPS users.

Revised approach: **removed `Name` entirely** instead of renaming. Both paths require exactly one ALB replacement for existing users, but removing `Name` is the permanently correct fix — no future condition change can ever trigger a replacement again.

Changes in force-push:
- ALB: `Name` property removed (was conditional rename → now omitted)
- Rebased on latest `main`
- Added regression test (`test_otel_collector_no_hardcoded_names`) that statically asserts no `Name` on ALB or `ServiceName` on ECS in the template — addresses the "missing test coverage" concern without requiring a live stack
- Documented the one-time migration in the PR description

We'll verify the fresh-stack and HTTP→HTTPS toggle paths in our development environment and can report back if useful.

spawnia:
**Update: tested fresh stack + HTTPS deployment in development.**

Verified:
- [x] Fresh stack from scratch (deleted existing, deployed clean) — all resources created successfully, no naming conflicts
- [x] OTEL endpoint reachable via HTTPS — returns 401 (JWT validation active, correct)
- [x] ECS service running with auto-generated name (`claude-code-auth-monitoring-ECSServiceHTTPS-<suffix>`)
- [x] Target group healthy
- [x] All 8 stacks deploy cleanly

Migration notes for existing users:
1. **In-place update won't work** — removing `Name` triggers ALB replacement, which fails because the target group can't be associated with two load balancers simultaneously during the replacement window
2. **Workaround**: delete the monitoring stack, then redeploy fresh. This is a one-time operation.
3. **Dependent stacks** (cowork-dashboard): if metric filters reference the log group from the old monitoring stack, they'll fail on update. Delete and recreate the dependent stack too.

The `ccwb deploy` force-redeploy step also has a pre-existing issue: it hardcodes `otel-collector-service` as the ECS service name. With auto-generated names, this lookup fails (non-blocking warning). I'll open a separate PR for that.

wirjo:
Merged via #603 — rebased onto latest `beta` to resolve conflicts (the `HasCustomDomain` → `EnableHttps` condition rename).

Thanks @spawnia for the fix and the thorough verification notes. Co-author credit preserved in the commit.

PR #575, opened 2026-06-16
feat(auth): make OIDC prompt parameter configurable

spawnia:
## Why

Enterprise deployments with a single Azure AD tenant get an unnecessary account picker on every browser-based auth flow. The `prompt=select_account` parameter forces the picker even when only one account is signed in — adding a pointless click every time a session refreshes via browser (when refresh_token is unavailable or expired).

## What

Add an `oidc_prompt` configuration field that controls the `prompt` parameter sent during Azure AD authorization requests:

- **Default**: `"select_account"` — backward compatible, same behavior as today
- **`""`** (empty string): omit the prompt parameter entirely, letting Azure reuse the existing session silently
- **Any other value**: sent as-is (e.g. `"login"`, `"consent"`, `"none"`)

Changes across both credential-process implementations:
- **Go**: `ProfileConfig.OIDCPrompt` field (`*string`, nil = default), threaded through `Authenticate()`
- **Python**: reads `oidc_prompt` from config dict, falls back to `"select_account"`
- **Config**: `Profile.oidc_prompt` field for serialization into config.json

Also adds a `per-file-ignores` rule for pre-existing E501/F841 issues in `credential_provider/__main__.py` that block the ruff hook.

## Verified

- [x] Go tests pass (`go test ./internal/config/ ./internal/oidc/ ./internal/provider/`) — all PASS
- [x] Config deserialization test added: absent field → nil (default), explicit `""` → skip prompt, explicit value → use as-is
- [x] Historical config fixtures still load correctly (backward compatible)
- [x] Non-Azure providers unaffected (prompt logic gated by `providerType == "azure"`)
- [ ] Default behavior unchanged: Azure auth shows account picker without config change — requires interactive browser test
- [ ] Set `"oidc_prompt": ""` → Azure reuses session (no picker) — requires interactive browser test
- [ ] Set `"oidc_prompt": "login"` → Azure forces re-authentication — requires interactive browser test

🤖 Generated with Claude Code

### DIFF
--- a/source/claude_code_with_bedrock/config.py
+++ b/source/claude_code_with_bedrock/config.py
@@ -39,7 +39,9 @@ class Profile:
     updated_at: str = field(default_factory=lambda: datetime.utcnow().isoformat())
     provider_type: str | None = None  # Auto-detected: "okta", "auth0", "azure", "cognito", "google", "generic"
     cognito_user_pool_id: str | None = None  # Only for Cognito User Pool providers
-    okta_auth_server: str = ""  # Okta authorization server ID ("default" for dev/free plans, empty for Org server on paid plans)
+    okta_auth_server: str = (
+        ""  # Okta authorization server ID ("default" for dev/free plans, empty for Org server on paid plans)
+    )
 
     # Generic OIDC provider configuration (provider_type == "generic")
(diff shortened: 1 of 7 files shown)


wirjo:
Closing to recreate from maintainer fork (rebased to latest beta, resolved conflict). Thank you @spawnia! Credit preserved in the new PR.

PR #576, opened 2026-06-16
feat(settings): support managed-settings.json deployment for org enforcement

### LINKED ISSUE #538: Add option to deploy enforcement settings to managed-settings.json instead of user-scope settings.json
## Summary

The install scripts currently write Bedrock enforcement and permission configuration to the user-scope `~/.claude/settings.json`. For organization-wide enforcement, these should be deployable to the OS-level **managed settings** location instead, which has the highest precedence and cannot be overridden by user or project settings.

## Problem

User-scope `~/.claude/settings.json` sits at the **lowest** precedence in Claude Code's settings hierarchy (managed → CLI args → local project → shared project → user). Any policy written there can be trivially overridden — or simply edited/deleted — by the end user. For a controlled rollout (~100 users in our case), the whole point of pushing `deny` rules, Bedrock-only egress controls, `forceLoginMethod`, etc. is that they're non-overridable.

Managed settings are the documented mechanism for exactly this: *"Security policies that must be enforced organization-wide"* and *"Compliance requirements that can't be overridden."*

## Proposed change

Add an option (flag or config toggle) so the install scripts write enforcement config to the managed-settings location for the target OS:

| OS | Path |
|---|---|
| macOS | `/Library/Application Support/ClaudeCode/managed-settings.json` |
| Linux / WSL | `/etc/claude-code/managed-settings.json` |
| Windows | `C:\Program Files\ClaudeCode\managed-settings.json` |

Same JSON format as `settings.json`, so the existing generated config should drop in with minimal changes.

## Implementation notes

- **Privileges:** all three paths require elevated/admin rights to write (`sudo` on macOS/Linux, Administrator on Windows). The installer should detect this and prompt or fail with a clear message rather than silently writing nothing.
- **Keep user-scope as the default** for the existing scout/individual-developer flow; managed-settings deployment should be opt-in for org rollout.
- **Managed-only fields become available** in this mode and are worth surfacing as options, e.g. `forceLoginMethod`, `forceLoginOrgUUID`, `allowManagedPermissionRulesOnly`, `allowManagedMcpServersOnly`, `disableBypassPermissionsMode`. These are ignored when set in user scope.
- **MDM alternative:** for fleets already managed via Jamf/Kandji/Intune/Group Policy, the same config can be delivered as a plist (`com.anthropic.claudecode`) / registry policy (`HKLM\SOFTWARE\Policies\ClaudeCode`) instead of a file. Worth documenting as an alternative path even if the script only handles the file-based approach initially. Anthropic publishes starter MDM templates at `anthropics/claude-code/tree/main/examples/mdm`.
- **Windows path note:** the legacy `C:\ProgramData\ClaudeCode\` path is no longer supported as of v2.1.75 — use `C:\Program Files\ClaudeCode\`.
- **Validation:** managed settings parse tolerantly (invalid entries are stripped with a warning rather than failing the whole file), so suggest documenting `claude doctor` / `/status` for verifying the deployed policy is active and shows source `(file)`.

## References

- Settings files & precedence: https://code.claude.com/docs/en/settings#settings-files

wirjo:
## Why

Users deploying to ~100+ seats need enforcement settings (Bedrock routing, permissions deny-rules, `forceLoginMethod`) to be **non-overridable**. The current installer writes everything to `~/.claude/settings.json` (user-scope, lowest precedence), which users can trivially edit or delete.

Additionally, `install.sh` currently **overwrites** existing `settings.json` entirely — clobbering users' custom preferences (reported by @jorgevazquezsanchez in #538).

## What

Add opt-in support for deploying settings to the OS-level **managed-settings.json** location, which has the highest precedence in Claude Code's settings hierarchy.

### Changes

1. **New `settings_target` profile field** — `'user'` (default) or `'managed'`
2. **Filename selection** — `_create_claude_settings` outputs to `managed-settings.json` when target is managed
3. **install.sh** — detects `managed-settings.json`, requires sudo, writes to OS path:
   - macOS: `/Library/Application Support/ClaudeCode/managed-settings.json`
   - Linux/WSL: `/etc/claude-code/managed-settings.json`
4. **install.bat** — detects `managed-settings.json`, requires Administrator, writes to:
   - Windows: `C:\Program Files\ClaudeCode\managed-settings.json`
5. **Merge logic** for user-scope settings — deep-merges incoming config into existing `settings.json` instead of overwriting (preserves user customizations)
6. **`--managed` flag on `ccwb init`** — opt-in to managed deployment mode
7. **13 new tests** covering backward compat, filename selection, installer scripts, and merge behavior

### Backward Compatibility

- Default is `settings_target='user'` — zero behavior change for existing users
- Profiles without the field work correctly (dataclass default)
- Managed mode is strictly opt-in via `ccwb init --managed`

### Simplification Notes

- Same settings dict generated regardless of target — only the output filename and installer path differ
- Deep merge replaces the old overwrite-with-prompt flow, solving @jorgevazquezsanchez's complaint even for non-managed users

## Testing

```
993 passed, 22 skipped, 1 xfailed, 1 xpassed (full suite)
13 new tests in test_package_managed_settings.py
```

Closes #538

Credit: @adiospeds for the detailed issue write-up and implementation notes.

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -76,6 +76,7 @@ poetry run ccwb init [options]
 **Options:**
 
 - `--profile <name>` - Configuration profile name (optional, will prompt if not specified)
+- `--managed` - Deploy settings to OS-level managed-settings.json (highest precedence, non-overridable by users)
 
 **What it does:**
 
@@ -254,7 +255,7 @@ poetry run ccwb package [options]
 
 1. Cross-compiles native Go binaries for all selected platforms (with `--go`)
 2. Creates `config.json` with federation config read from the admin profile
-3. Creates `claude-settings/settings.json` with Bedrock model and OTel endpoint
+3. Creates `claude-settings/settings.json` with Bedrock model and OTel endpoint (or `managed-settings.json` if `--managed` was used during init)
 4. Creates installer scripts (`install.sh`, `install.bat`, `ccwb-install.ps1`)
 5. Outputs to `dist/{profile}/{timestamp}/`
 
--- a/assets/docs/DEPLOYMENT.md
+++ b/assets/docs/DEPLOYMENT.md
@@ -159,6 +159,25 @@ Windows binary builds use AWS CodeBuild with Nuitka for optimal performance. Win
 - Windows builds take 20+ minutes
 - To enable Windows builds after initial setup, re-run `poetry run ccwb init`
 
+### Organization-Wide Enforcement (Optional)
+
+For large deployments (50+ users) where settings must be non-overridable, use managed settings:
+
+```bash
+poetry run ccwb init --managed
+poetry run ccwb package --go
+```
+
+This writes settings to the OS-level `managed-settings.json` path instead of user-scope `~/.claude/settings.json`. Managed settings have the **highest precedence** in Claude Code's settings hierarchy and cannot be edited or overridden by end users.
+
+| OS | Managed path | Requires |
+|---|---|---|
+| macOS | `/Library/Application Support/ClaudeCode/managed-settings.json` | `sudo` |
+| Linux/WSL | `/etc/claude-code/managed-settings.json` | `sudo` |
+| Windows | `C:\Program Files\ClaudeCode\managed-settings.json` | Administrator |
+
+The installer will detect managed settings in the package and prompt for elevated privileges. For MDM-managed fleets (Jamf, Intune, Group Policy), see Anthropic's [MDM templates](https://github.com/anthropics/claude-code/tree/main/examples/mdm) for alternative delivery.
+
 ## Phase 4: Testing Your Deployment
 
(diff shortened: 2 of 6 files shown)


wirjo:
Hey @adiospeds @jorgevazquezsanchez — this PR implements the managed-settings feature you requested in #538. Would appreciate your review when you get a chance.

**Key points:**
- `ccwb init` now asks whether to target user-scope or managed (org enforcement) settings
- Managed mode writes to the OS-level path (`/Library/Application Support/ClaudeCode/`, `/etc/claude-code/`, or `C:\Program Files\ClaudeCode\`) with elevation checks
- Also fixes the overwrite issue @jorgevazquezsanchez raised — user-scope installs now **merge** into existing `settings.json` instead of clobbering it

Let me know if this covers your use case or if anything needs adjusting.

adiospeds:
Will be testing this today.

wirjo:
@adiospeds Let me know if you have any updates

adiospeds:
I got some other things on my plate which have some priority, but I will definitely test this and post the results here by Monday.

adiospeds:
Tested managed settings for macos and it is working as expected.. Good to go into main branch via #634

PR #581, opened 2026-06-17
fix(otel): validate monitoring token expiry before use

### LINKED ISSUE #561: OTEL telemetry: env-var monitoring token attached without expiry (exp) validation
## 🟡 Maintainer decision requested — proposal, close freely

This is a **judgment call**, not a clear-cut bug, so I'm opening it as a proposal rather than a PR. There's an explicit **Option A vs Option B** below; if you'd rather not change anything here, please close — no attachment.

## Describe the bug

A monitoring token supplied via the `CLAUDE_CODE_MONITORING_TOKEN` environment variable is attached as the OTEL `Authorization: Bearer …` header on **truthiness alone** — its `exp` (expiry) claim is never checked. The file and keyring token paths *do* validate expiry before use, so the env-var path is the lone exception.

The practical effect: if an **expired** token sits in that env var, it gets spliced onto telemetry requests, the collector/ALB rejects it with **401**, and the telemetry is dropped **silently** — no log explains why.

**Honest scope note:** dropping the expired token does **not** make telemetry succeed (the token is expired regardless). It converts a *silent* 401 into a *diagnosable* "emitting headers without authorization" log by falling through to the existing no-token branch. The benefit is **diagnosability**, not recovery. This is **hardening**, not a hot-path outage — the normal trigger is a user *manually* exporting a stale token; credential-process itself only ever writes a fresh token.

## Reproducible steps

1. Export a syntactically valid but **expired** JWT: `export CLAUDE_CODE_MONITORING_TOKEN=<expired-jwt>`
2. Run a flow that emits OTEL telemetry through the helper.
3. Observe the request is sent with the expired `Bearer` token and is rejected **401** by the collector/ALB — with no diagnostic indicating the token was stale.

(For contrast, an equally-stale token in the cache file or keyring is correctly dropped before send.)

## Root cause (file:line, on `beta` @ `f2afa73`)

The env-var branch returns the token with **no `exp` check**, while the file/keyring branch in the *same function* does check it:

- **Go** — `source/go/internal/storage/monitoring.go:13-15` (env branch, no check) vs `:29-33` (file/keyring branch, `Expires-now <= 600` → drop):
  ```go
  // Check environment first
  if token := os.Getenv("CLAUDE_CODE_MONITORING_TOKEN"); token != "" {
      return token, nil          // <-- no exp validation
  }
  ...
  if data.Expires-now <= 600 { return "", nil }   // file/keyring path IS guarded
  ```
- **Python** — `source/credential_provider/__main__.py:858-860` (env branch, no check) vs `:892` (`exp_time - now > 60` → use):
  ```python
  env_token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN")
  if env_token:
      return env_token            # <-- no exp validation
  ```

The Go `attachBearer` helper documents the intended contract — and this env branch is exactly what violates it:

- `source/go/cmd/otel-helper/main.go:48-52`:
  ```go
  // attachBearer sets the Authorization header. Intentionally exp-AGNOSTIC:
  // ... validation lives at the token source (see storage.GetMonitoringToken), NOT here.
  ```
  i.e. the splice site deliberately delegates expiry checking "to the token source," but the token source skips it for the env-var case.

The otel-helper also reads the env var **directly** at several sites, bypassing the guarded function entirely and feeding straight into the splice: `source/otel_helper/__main__.py:642` (proxy), `:839` (Layer-1), `:857` (Layer-2); Go `source/go/cmd/otel-helper/main.go:92`, `:109`. The existing no-token fallback that emits headers without `authorization` already lives at `otel_helper/__main__.py:842-849`.

## Impact / Evidence

- **Who:** only users who *manually* set a stale `CLAUDE_CODE_MONITORING_TOKEN`. Normal use never hits this — credential-process writes a fresh token at the same moment it validates it (`credential_provider/__main__.py:845`).
- **Severity:** Medium / hardening. No data is recoverable when the token is expired; the fix only improves observability of the failure.
- **Secondary parity note (out of scope here):** the expiry buffers diverge across the codebase — Go `monitoring.go:31` = **600s**, Python `:892` = **60s**, `otel_helper/__main__.py` `read_cached_headers` = **60s**, `otel-helper.sh` = **60s**. Reconciling that is a separate, file-path-affecting change and should not be bundled with this.

## Version / environment

- Confirmed on `beta` @ `f2afa73`.
- **Also present on `main`** (`e2d4057`): `internal/storage/monitoring.go` is byte-identical, and the Python env branch is present at `credential_provider/__main__.py:849`. So a fix would apply to both.

## Why CI missed it

Tests exercise the file/keyring expiry guard, but there's no test that feeds an **expired token via the env var** and asserts it's dropped. The env-var branch's "return on truthiness" passes every existing test because no test sets an expired env token.

---

## Decision requested — Option A vs Option B

| | **Option A — code fix** | **Option B — doc note** |
|---|---|---|
| **What** | Add an `exp` gate to the env-var branch at the token source (Go `GetMonitoringToken` + Python `get_monitoring_token`), and route the otel-helper's direct env reads through that guarded resolver so the gate is single-sourced. Expired / no-`exp` / malformed → drop and fall through. | Document that `CLAUDE_CODE_MONITORING_TOKEN` is trusted as-is and must not be set to a stale token manually. |
| **Effort** | ~30-40 lines + a Go↔Python **parity test** (mandatory — these are Tier-1 files). | One line of docs. |
| **Pro** | Single-sourced expiry contract; converts the silent 401 into a diagnosable log even if a stale token is set. | Zero code risk; matches the reality that the only trigger is manual misuse. |
| **Con / ripple** | The "drop no-`exp` tokens too" rule flips Go's `GetMonitoringToken` to return `""` for a no-`exp` env token, which downstream (`credential-process/main.go` ~196) **can trigger a browser re-auth**. Behavior change a maintainer may not want. | The gap remains; a manually-set stale token still 401s silently. |

**My lean: Option B (doc note)** — the trigger is manual-only, the fix recovers no data, and Option A carries a re-auth ripple that may be undesirable. But this is your call. Happy to implement Option A (with the parity test) if you prefer it.

_Discovered while reviewing the otel-helper Bearer follow-ups (the choke point added in #526)._

wirjo:
## Why

When `CLAUDE_CODE_MONITORING_TOKEN` env var contains an expired token, otel-helper blindly attaches it to telemetry requests. The result depends on the deployment mode:

| Mode | What happens with expired token | User impact |
|---|---|---|
| **Central/proxy** (ECS) | ALB rejects with 401 | Telemetry **silently dropped** — no metrics, no diagnostic |
| **Sidecar** (local) | Token used for attribution headers | User shows as **anonymous** instead of real identity |

Both are silent failures — the user sees no error, just missing or incorrect data.

Fixes #561. Credit: @peepeepopapapeepeepo (issue report + design proposal).

## Who is affected

- **Central mode users** where the ECS task has `CLAUDE_CODE_MONITORING_TOKEN` injected via environment (e.g., Secrets Manager) and it expired without rotation → silent telemetry loss
- **Sidecar mode users** with long-running terminal sessions where the env var was set by credential-process and the token expired while the shell stayed open → anonymous attribution
- **Normal users** (credential-process refreshes tokens automatically) → **not affected** — this is an edge case for stale env vars

## Change

Added `is_token_expired(token, buffer_seconds=60)` and applied at all 3 env-var read sites:

1. **Line 663** (sidecar/headers mode) — expired → skip env var, try credential-process
2. **Line 860** (proxy mode Bearer attachment) — expired → skip env var, try credential-process
3. **Line 878** (main entry) — expired → skip env var, fall through to credential-process

In all cases: expired token is discarded and the code falls through to `get_token_via_credential_process()` which handles refresh. If that also fails, the existing anonymous/no-token paths handle it gracefully.

## Risk & Backwards Compatibility

| Concern | Assessment |
|---|---|
| **Valid token (normal flow)** | Unchanged — `is_token_expired` returns False, token used as before |
| **Expired token (bug case)** | Improved — falls through to refresh instead of silent 401 |
| **No token at all** | Unchanged — same anonymous/fallback path |
| **Performance** | Negligible (~0.1ms: base64 decode + JSON parse + time comparison) |
| **Fail-safe** | Unparseable tokens treated as expired → triggers fallback, never crash |

**Regression risk: Very low.** The only behavior change is for expired tokens, which previously failed silently anyway. Valid tokens are completely unaffected.

## Testing (6 regression tests + 1 existing test fix)

New tests (`test_env_token_expiry.py`):
- `test_valid_token_not_expired` — 1h future token passes
- `test_expired_token_detected` — 1h past token caught
- `test_token_within_buffer_is_expired` — 30s remaining (within 60s buffer) → expired
- `test_token_beyond_buffer_is_valid` — 120s remaining → valid
- `test_no_exp_claim_treated_as_expired` — fail-safe
- `test_malformed_token_treated_as_expired` — fail-safe

Existing test fix (`test_otel_latency_fixes.py`):
- `test_exception_emits_empty_json_and_exit_zero` — patched to bypass expiry check on dummy token (test targets decode error path, not expiry path)

## Files

- `source/otel_helper/__main__.py` — `is_token_expired()` + 3 call sites modified
- `source/tests/test_env_token_expiry.py` — 6 new tests
- `source/tests/test_otel_latency_fixes.py` — 1 line added (monkeypatch for new function)

### DIFF
--- a/source/otel_helper/__main__.py
+++ b/source/otel_helper/__main__.py
@@ -106,6 +106,27 @@ def parse_args():
 # This prevents macOS keychain permission prompts for the OTEL helper
 
 
+def is_token_expired(token, buffer_seconds=60):
+    """Check if a JWT token's exp claim has passed.
+
+    Returns True if the token is expired or unparseable (fail-safe: treat
+    bad tokens as expired so they fall through to re-authentication).
+    """
+    try:
+        _, payload_b64, _ = token.split(".")
+        padding_needed = len(payload_b64) % 4
+        if padding_needed:
+            payload_b64 += "=" * (4 - padding_needed)
+        payload_b64 = payload_b64.replace("-", "+").replace("_", "/")
+        payload = json.loads(base64.b64decode(payload_b64))
+        exp = payload.get("exp")
+        if exp is None:
+            return True  # No exp claim = treat as expired
+        return time.time() > (exp - buffer_seconds)
+    except Exception:
+        return True  # Unparseable = treat as expired
+
+
 def decode_jwt_payload(token):
     """Decode the payload portion of a JWT token"""
     try:
@@ -639,7 +660,10 @@ def build_proxy_user_headers() -> dict:
     """
     token = None
     if not ANONYMOUS_MODE:
-        token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN") or get_token_via_credential_process()
+        token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN")
+        if token and is_token_expired(token):
+            token = None
+        token = token or get_token_via_credential_process()
 
     if token:
         payload = decode_jwt_payload(token)
@@ -836,7 +860,10 @@ def main():
         if cached_headers is not None:
             # Resolve Bearer fresh — the cache stores attribution headers only,
             # never the token itself. Try env var (free) then credential-process (~20ms).
-            _bearer_token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN") or get_token_via_credential_process()
+            _bearer_token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN")
+            if _bearer_token and is_token_expired(_bearer_token):
+                _bearer_token = None
+            _bearer_token = _bearer_token or get_token_via_credential_process()
             if _bearer_token:
                 _attach_bearer(cached_headers, _bearer_token)
             else:
@@ -855,6 +882,9 @@ def main():
     token = None
     if not ANONYMOUS_MODE:
         token = os.environ.get("CLAUDE_CODE_MONITORING_TOKEN")
+        if token and is_token_expired(token):
+            logger.info("Environment token CLAUDE_CODE_MONITORING_TOKEN is expired, falling through to credential-process")
+            token = None
         if token:
             logger.info("Using token from environment variable CLAUDE_CODE_MONITORING_TOKEN")
         else:
--- a/source/tests/test_env_token_expiry.py
+++ b/source/tests/test_env_token_expiry.py
@@ -1,0 +1,63 @@
+# ABOUTME: Tests for env-var token expiry validation in otel-helper
+# ABOUTME: Ensures expired CLAUDE_CODE_MONITORING_TOKEN is not used blindly
+
+"""Regression tests for monitoring token expiry check (issue #561)."""
+
+import base64
+import json
+import time
+import importlib.util
+from pathlib import Path
+
+# Load otel_helper module
+_spec = importlib.util.spec_from_file_location(
+    "otel_helper_main",
+    Path(__file__).resolve().parents[1] / "otel_helper" / "__main__.py",
+)
+_module = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_module)
+is_token_expired = _module.is_token_expired
+
+
+def _make_jwt(exp_offset_seconds):
+    """Create a minimal JWT with exp claim at now + offset."""
+    payload = {"sub": "test", "email": "test@co.com", "exp": int(time.time()) + exp_offset_seconds}
+    payload_b64 = base64.urlsafe_b64encode(json.dumps(payload).encode()).rstrip(b"=").decode()
+    return f"eyJhbGciOiJSUzI1NiJ9.{payload_b64}.fake_signature"
+
+
+class TestIsTokenExpired:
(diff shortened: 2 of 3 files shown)


wirjo:
@peepeepopapapeepeepo — this implements your proposal from #561. Would appreciate your review, particularly:

1. The `is_token_expired` fail-safe behavior (malformed/missing exp → treat as expired)
2. The 60s buffer matching credential-process
3. Whether the fallthrough to `get_token_via_credential_process()` matches your Option A

Thanks for the clear issue writeup — made implementation straightforward. 🙏

PR #624, opened 2026-06-20
fix(init): preserve all quota settings when re-running init

qing-at-work:
## Why

Re-running `ccwb init` over an existing profile silently reset `quota_check_interval` (and several sibling quota fields) to their prompt defaults — e.g. a saved 5-minute re-check interval reverted to 30. A customer's configured value was lost on the next init run.

## Root cause

On a re-run, `_check_existing_deployment` rebuilds the in-memory config dict from the saved Profile, but its quota block only restored **4 of the 10** quota fields that `_save_configuration` persists. The omitted fields were absent from the rebuilt dict, so the wizard prompt defaults (and the subsequent save) fell back to hardcoded defaults — round-trip drift between the save and load paths.

Omitted fields: `daily_limit`, `burst_buffer_percent`, `daily_enforcement_mode`, `monthly_enforcement_mode`, `check_interval`, `enable_bypass_detection`.

## What

- Restore the 6 missing quota fields in the rebuild block, reversing the `quota_check_interval -> check_interval` key mapping that `_save_configuration` applies.
- Add a round-trip regression test asserting all quota fields survive profile → config reconstruction. Verified it fails without the fix and passes with it.

## Testing

- `poetry run pytest tests/cli/commands/test_init.py tests/cli/commands/test_init_quota.py tests/cli/commands/test_init_quota_roundtrip.py -q` → 42 passed.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -2804,13 +2804,24 @@ class InitCommand(Command):
                     "idp_userinfo_endpoint": getattr(profile, "distribution_idp_userinfo_endpoint", None),
                 }
 
-            # Add quota monitoring configuration if present
+            # Add quota monitoring configuration if present.
+            # Must mirror every quota field that _save_configuration persists, or a
+            # re-run of `ccwb init` silently resets the omitted fields to their prompt
+            # defaults (e.g. quota_check_interval -> 30). Note the key-name mapping:
+            # the config dict uses "check_interval" while the Profile attribute is
+            # "quota_check_interval" (see _save_configuration), so reverse it here.
             if hasattr(profile, "quota_monitoring_enabled"):
                 existing_config["quota"] = {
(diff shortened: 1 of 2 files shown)


wirjo (approved):
Excellent fix @qing-at-work — another init re-run bug squashed! 🎯

**Review:**
- **Risk: Very Low** — purely additive (6 missing `getattr()` calls restored to match the save path)
- **Root cause clear:** config rebuild only restored 4/10 quota fields; omitted ones reverted to defaults silently
- **Test quality:** Round-trip test creates a profile with non-default values, rebuilds via `_check_existing_deployment`, asserts all 10 fields survive. Fails without fix, passes with it.
- **Backward compat:** ✅ — defaults match the existing Profile defaults, so fresh installs see no change

Two PRs in one day (#619, #624) — both clean, well-tested, and fixing real user pain. Thank you for the consistently high-quality contributions! 🙏

**Verdict: ✅ Ship it**

PR #632, opened 2026-06-20
fix(dashboard): remove DefaultValue from dimensioned CoWork metric filters

adiospeds:
## Why
\`ccwb deploy\` failed on the \`cowork-dashboard\` stack with:
\`Invalid metric transformation: dimensions and default value are mutually exclusive properties\`.

The six \`claude_code.api_request\` metric filters declared both \`DefaultValue: 0\` and a
\`Dimensions\` block in the same \`MetricTransformation\`, which AWS::Logs::MetricFilter rejects.

## What
Removed \`DefaultValue: 0\` from those filters (Dimensions retained for per-user
\`user_email\` attribution). LAM filters are unchanged. Added a regression test that
fails if any transformation sets both properties."

The error during deployment: 
image

### DIFF
--- a/deployment/infrastructure/cowork-dashboard.yaml
+++ b/deployment/infrastructure/cowork-dashboard.yaml
@@ -35,7 +35,9 @@ Resources:
         - MetricNamespace: ClaudeCoWork
           MetricName: token.usage.input
           MetricValue: $.attributes.input_tokens
-          DefaultValue: 0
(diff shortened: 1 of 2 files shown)


wirjo (approved):
Clean fix — verified locally: all 1143 tests pass (including the new regression test), CFN structure correct. DefaultValue and Dimensions are indeed mutually exclusive per CloudWatch Logs docs. Thanks for the well-scoped PR with the regression test!

PR #638, opened 2026-06-21
feat(auth): support multiple profiles in one AWS account/region (#528)

### LINKED ISSUE #528: Support multiple profile deployments in the same AWS account and region
## Problem

When deploying two profiles (e.g., one for R&D and one for non-R&D employees) in the same AWS account and region using the same Okta domain, the second deployment fails because `AWS::IAM::OIDCProvider` is keyed by issuer URL. IAM only allows one OIDC provider per unique issuer URL per account, so the second stack cannot create a duplicate provider for the same Okta domain.

The conflict occurs in `bedrock-auth-okta.yaml` at the `OktaOIDCProvider` resource — both stacks attempt to create an OIDC provider with `Url: https://${OktaDomain}`, but only one can exist.

## Use Case

I want to deploy two separate profiles with **different usage limits/quotas** for different employee groups:
- **Profile 1 (R&D):** Higher limits for engineering teams
- **Profile 2 (Non-R&D):** Lower limits for other employees

Both profiles use the same Okta organization (same domain) but different Okta application client IDs. Both should be deployed in the same AWS account and region.

## Motivation — Why Same Account and Region?

1. **Bedrock prompt caching:** Prompt cache is regional. Splitting users across accounts or regions may reduce cache hit ratio, leading to increased latency and cost. Keeping all users in the same account/region maximizes the chance of cache reuse across the organization.
2. **Operational simplicity:** Single account means unified billing, quota management, CloudWatch monitoring, and cost attribution.
3. **Quota management:** Bedrock service quotas (tokens-per-minute, requests-per-minute) are per-account per-region. A single account allows the organization to manage one set of quotas rather than coordinating across multiple accounts.

## Suggested Approaches

1. **Make OIDC provider creation conditional:** Accept an optional parameter for an existing OIDC provider ARN. If provided, skip creation and reference the existing one. If not provided, create it as today.
2. **Add ClientId to existing provider:** Instead of creating a new provider, add the new `ClientId` to the existing OIDC provider's `ClientIdList` (AWS supports multiple client IDs per provider).
3. **Import existing provider:** Use CloudFormation import or a custom resource to detect and reuse an existing OIDC provider for the same issuer URL.

Option 1 seems simplest — a new parameter like `ExistingOIDCProviderArn` that, when provided, skips the `OktaOIDCProvider` resource creation and uses the supplied ARN for role trust policies.

## Current Workaround

None that preserves proper CloudFormation lifecycle management. Manual creation of the OIDC provider outside both stacks is fragile.

## Environment

- Auth provider: Okta (same domain, two separate app registrations)
- Federation type: direct
- Deployment target: single AWS account, single region

peepeepopapapeepeepo:
## Why

Deploying a second profile that shares an OIDC issuer in the same AWS account fails the auth stack with `EntityAlreadyExists: Provider with url ... already exists` — IAM allows only one OIDC provider per issuer URL per account. Google always hits this (the issuer is always `accounts.google.com`); okta/azure/auth0/generic hit it whenever two profiles share an issuer; cognito-pool hits it when two direct-mode stacks point at the same user pool.

This is the exact scenario in #528: two profiles (e.g. R&D vs non-R&D) with different quotas, same IdP, same account/region (to preserve regional Bedrock prompt-cache locality and unified quota/billing).

Once the provider collision is resolved, a second profile of the **same provider type** then collides on the IAM **role name** (the federation role name is a template default, e.g. `BedrockAzureFederatedRole`, and `deploy.py` never overrode it on the standard auth path). Both same-account collisions are fixed here so the multi-profile case actually works end to end.

Fixes #528

## What

Implements the maintainer-recommended **Option 1** from the issue thread: an optional `ExistingOIDCProviderArn` parameter on the auth templates. When supplied, the stack reuses the existing provider instead of creating a duplicate; when empty (the default), behavior is byte-for-byte unchanged.

- **CFN (all 6 `bedrock-auth-*.yaml`):** new `ExistingOIDCProviderArn` parameter (`Default: ''`, partition-aware ARN `AllowedPattern`) + `CreateOIDCProvider` condition gating the provider resource. Every provider-ARN reference (trust policy, Cognito identity pool, output) is wrapped in `!If [CreateOIDCProvider, !GetAtt …, !Ref ExistingOIDCProviderArn]`. cognito-pool uses a combined `!And [UseDirectIAM, !Equals [arn, '']]` since its provider only exists in direct mode.
- **Role name:** per-profile `FederatedRoleName`, derived for **new direct** profiles (`Bedrock<Provider>FederatedRole-<identity_pool_name>`, ≤64 chars), mirroring the existing IDC path. Existing profiles keep their value (unset → template default), so a re-run never renames a deployed role.
- **Python:** optional `existing_oidc_provider_arn` and `federated_role_name` Profile fields (both default `None`, backward-compatible); `deploy.py` passes each param only when set; `init.py` adds an optional OIDC-provider-ARN wizard prompt. No Go change (the credential-process reads `federated_role_arn`, never the provider ARN or role name).

## Operator note

When reusing a provider, its `ClientIdList` must already include the new profile's client ID, or `sts:AssumeRoleWithWebIdentity` is denied at runtime. In **direct** federation mode the role trusts the provider without an `aud` condition, so a shared provider whose `ClientIdList` includes multiple client IDs lets a token for any of them assume any direct-mode role trusting that provider — set the ARN only for providers you control. (The default Cognito-Identity path is unaffected: it scopes by the per-profile identity pool's `aud`.)

## How / Why this is safe

ARN set → `CreateOIDCProvider=false` → no provider resource, no `CreateOIDCProvider` API call, all references resolve to the supplied ARN. ARN empty (default, and for every existing config) → `CreateOIDCProvider=true` → provider created and all `!If` resolve to `!GetAtt` — identical to pre-change behavior, so existing single-profile deploys and stack updates are unaffected. `FederatedRoleName` is passed only when set, so existing profiles keep their role ARN stable.

## Testing

- [x] cfn-lint passes on all 6 templates (cfn-lint 1.51.4)
- [x] New regression tests: CFN structure (all 6 templates, no bare provider `!GetAtt` remains), config backward-compat (both new fields default to `None`, old configs load), deploy param-threading (real boto3 kwarg, passed-when-set / absent-when-None), init role-name derivation (pattern + ≤64, new derives / existing preserved) + init round-trip restore of `existing_oidc_provider_arn`
- [x] Falsified against beta: each regression test fails on beta's files, passes on the fix
- [x] Live-validated on Azure/Entra direct mode (two profiles, one account, ap-southeast-1): collision reproduced on the unfixed path, both stacks reach `CREATE_COMPLETE` with the fix — one shared provider, two distinct roles
- [x] Rebased on latest `beta`; full targeted suite green; mypy + ruff clean on changed lines

## Notes

This bundles two same-account collisions (OIDC provider + IAM role name) because fixing only the provider leaves the feature non-functional for same-provider-type direct-mode deploys. Happy to split into two PRs if preferred. The Lambda auto-append of the client ID to an existing provider's `ClientIdList` (the issue's "bonus") is intentionally deferred — the manual step is documented above.

### DIFF
--- a/deployment/infrastructure/bedrock-auth-auth0.yaml
+++ b/deployment/infrastructure/bedrock-auth-auth0.yaml
@@ -57,10 +57,25 @@ Parameters:
       - 'false'
     Description: Enable CloudWatch monitoring for Bedrock usage
 
+  ExistingOIDCProviderArn:
+    Type: String
+    Default: ''
+    Description: >-
+      ARN of a pre-existing IAM OIDC provider for this issuer URL. Leave empty
+      (default) to create a new provider. Provide an existing provider's ARN to
+      reuse it instead of creating a duplicate - useful when deploying multiple
+      profiles that share the same identity provider in one AWS account/region
+      (IAM allows only one OIDC provider per issuer URL per account). The
+      existing provider's ClientIdList must already include this profile's
+      client ID.
+    AllowedPattern: '^(arn:aws[a-zA-Z-]*:iam::[0-9]{12}:oidc-provider/.+)?$'
+    ConstraintDescription: Must be empty or a valid IAM OIDC provider ARN
+
 Conditions:
   UseDirectIAM: !Equals [!Ref FederationType, direct]
   UseCognitoIdentity: !Equals [!Ref FederationType, cognito]
   MonitoringEnabled: !Equals [!Ref EnableMonitoring, 'true']
+  CreateOIDCProvider: !Equals [!Ref ExistingOIDCProviderArn, '']
   # Partition-aware conditions for Cognito Identity service principals
   IsGovCloudWest: !Equals [!Ref 'AWS::Region', 'us-gov-west-1']
   IsGovCloudEast: !Equals [!Ref 'AWS::Region', 'us-gov-east-1']
@@ -72,6 +87,7 @@ Resources:
 
   Auth0OIDCProvider:
     Type: AWS::IAM::OIDCProvider
+    Condition: CreateOIDCProvider
     Properties:
       Url: !Sub 'https://${Auth0Domain}/'
       ClientIdList:
@@ -167,7 +183,10 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              Federated: !GetAtt Auth0OIDCProvider.Arn
+              Federated: !If
+                - CreateOIDCProvider
+                - !GetAtt Auth0OIDCProvider.Arn
+                - !Ref ExistingOIDCProviderArn
             Action:
               - 'sts:AssumeRoleWithWebIdentity'
               - 'sts:TagSession'
@@ -193,7 +212,10 @@ Resources:
       AllowUnauthenticatedIdentities: false
       AllowClassicFlow: false
       OpenIdConnectProviderARNs:
-        - !GetAtt Auth0OIDCProvider.Arn
+        - !If
+          - CreateOIDCProvider
+          - !GetAtt Auth0OIDCProvider.Arn
+          - !Ref ExistingOIDCProviderArn
 
   CognitoAuthenticatedRole:
     Type: AWS::IAM::Role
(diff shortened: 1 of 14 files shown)


wirjo:
Thank you for this contribution! After reviewing the current roadmap, we've decided not to merge this PR — however, we'll keep it open as a reference for forks and community members who may find the approach useful for their own deployments.

Appreciate the time and effort you put into this. 🙏

PR #639, opened 2026-06-22
fix(install): prevent root-owned files when install.sh run with sudo

adiospeds:
## Why

When `install.sh` is generated with `managed-settings.json` support (PR #576), the only way to write to the OS-level path (`/Library/Application Support/ClaudeCode/` or `/etc/claude-code/`) is elevated privilege — and the current installer exits with an error unless the **entire script** is run as `sudo ./install.sh`.

Running the whole script as root causes all user-space files to be created with root ownership:

```
-rw-r--r--  1 root  staff   880B  Jun 22 04:21  ~/claude-code-with-bedrock/config.json
-rwxr-xr-x  1 root  staff   9.0M  Jun 22 04:21  ~/claude-code-with-bedrock/credential-process
-rw-r--r--  1 root  staff   146B  Jun 22  ~/.claude/settings.json.backup-*
```

This breaks subsequent credential-process execution and Claude Code settings reads for the real user.

## What

- **Detect the real user at startup** — `ACTUAL_USER`/`ACTUAL_HOME` derived from `$SUDO_USER` when present, falling back to `$USER`/`$HOME` for non-sudo runs
- **Replace all `~/` and `$HOME/` user-space paths** with `$ACTUAL_HOME/` so files always land in the correct user's home regardless of who runs the script
- **`chown` user-space files back to `ACTUAL_USER`** after each write when running under sudo
- **Replace "exit 1 if not root" for managed-settings** with targeted inline `sudo mkdir` / `sudo tee` — users run `./install.sh` normally; only the OS-level directory write escalates, not the whole script
- **5 new regression tests** covering: `ACTUAL_HOME` detection from `SUDO_USER`, non-sudo `$HOME` fallback, no hardcoded `~/` paths, no exit-if-not-root, `chown` presence

## Backward Compatibility

- Non-sudo runs (`./install.sh`): zero behaviour change — `SUDO_USER` is unset, `ACTUAL_HOME=$HOME`
- Managed-settings deployments that previously required `sudo ./install.sh` now work with `./install.sh` (sudo prompt appears only for the OS-level write)
- User-scope only deployments: completely unaffected

## Testing

```
18 passed in test_package_managed_settings.py (13 existing + 5 new)
```

Closes #576 (root-ownership side effect introduced by the managed-settings installer)

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -2431,6 +2431,17 @@ set -e
 SCRIPT_DIR="$(cd "$(dirname "${{BASH_SOURCE[0]}}")" && pwd)"
 cd "$SCRIPT_DIR"
 
+# Resolve the real invoking user — safe whether or not run with sudo.
+# If run as: sudo ./install.sh  →  SUDO_USER is set; use that for home-dir writes.
+# If run as: ./install.sh       →  use $USER / $HOME directly.
+if [ -n "$SUDO_USER" ]; then
+    ACTUAL_USER="$SUDO_USER"
+    ACTUAL_HOME=$(eval echo "~$SUDO_USER")
+else
+    ACTUAL_USER="$USER"
+    ACTUAL_HOME="$HOME"
+fi
+
 echo "======================================"
 echo "Claude Code Authentication Installer"
 echo "======================================"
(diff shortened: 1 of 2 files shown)


wirjo (approved):
Great fix @adiospeds! Verified locally — all 1181 tests pass (including the 5 new regression tests), zero regressions. The SUDO_USER detection and targeted escalation approach is clean. Thanks for catching the root-ownership side effect and shipping with comprehensive tests.

Issue #86, opened 2025-12-17
Target group name cannot be longer than 32 characters


peterdeoliveira:
Hi there,

Thank you for the amazing work you guys are doing here.

We have started deployment and faced the `Target group name cannot be longer than 32 characters` error when deploying the CFN monitoring stack (for the monitoring target group name).

<img width="2816" height="684" alt="Image" src="https://github.com/user-attachments/assets/ef0d3ffb-36cf-4a85-ae5c-bd2fbdda0611" />

It would be a good idea if there is a warning for this before deployment, or some kind of validation. 
The only reason we get this error is becoming a string is concatenated into our CFN stack name and then it exceeds the characters limits.

Also, I believe it would be good idea to use CFN Nested Stacks, because upon failure, we need to delete the stacks one by one, while with Nested stacks, it would be deleted in one shot.

Keep up the good work!

Best,
Peter

mortizbey:
Hit this today on a fresh deploy. Confirming the root cause and sharing a workaround.

## Reproduction

- `ccwb init` produced stack name `claude-code-bedrock-monitoring` (30 chars).
- `deployment/infrastructure/otel-collector.yaml` line 179 computes the target group name as `!Sub '${AWS::StackName}-tg'`, so the result is `claude-code-bedrock-monitoring-tg` = **33 chars**, over the ELBv2 32-char cap by exactly one character.

```
Target group name 'claude-code-bedrock-monitoring-tg' cannot be longer than '32' characters
(Service: ElasticLoadBalancingV2, Status Code: 400)
```

## Workaround (user-side)

Shorten the monitoring stack name in the profile. Edit `~/.ccwb/profiles/<profile>.json`:

```json
"stack_names": {
  "monitoring": "claude-code-bedrock-otel"
}
```

`claude-code-bedrock-otel` is 24 chars, so the target group becomes 27 chars, safely under the limit.

Then delete the failed stack (which will be in `ROLLBACK_COMPLETE`) and re-run `ccwb deploy`. Worked cleanly for me.

## Suggested fixes (template-side)

Three reasonable options for `otel-collector.yaml`:

**Option 1 — drop the explicit name.** Let CloudFormation auto-generate a guaranteed-unique short name:

```yaml
HTTPTargetGroup:
  Type: AWS::ElasticLoadBalancingV2::TargetGroup
  Properties:
[...]

PR #643, opened 2026-06-22
fix(otel-collector): replace target group when load balancer is replaced

HristoStoyanovYotpo:
## Why

When a stack update forces `LoadBalancer` replacement (e.g. changing `Subnets`, `Scheme`, or `SecurityGroups`), CloudFormation creates the new LB but then tries to attach the existing `HTTPTargetGroup` to it while the old LB still owns it, producing:

```
Resource handler returned message: "The following target groups cannot be
associated with more than one load balancer: arn:aws:elasticloadbalancing:
us-east-1:...:targetgroup/claude-HTTPT-VTF44BENVVF2/..."
(Service: ElasticLoadBalancingV2, Status Code: 400)
```

Root cause: `HTTPTargetGroup` had no `Name`, so CloudFormation never replaced it — the logical ID was stable and the TG was reused across LB replacements. A target group can only be associated with one LB at a time.

## What

Derive the TG `Name` from the ALB's unique hex ID (last path segment of its ARN):

```yaml
Name: !Sub
  - 'tg-${AlbId}'
  - AlbId: !Select [3, !Split ['/', !Ref LoadBalancer]]
```

When the LB is replaced its hex ID changes → the TG Name changes → CloudFormation replaces the TG alongside the LB → no dual-LB conflict.

The name stays within the 32-char limit: `"tg-"` (3) + 16-char hex ID = 19 chars max. This avoids the >29-char stack name issue from #86 that previously made `!Sub '${AWS::StackName}-tg'` unusable.

## ⚠️ Upgrade note for existing stacks

**First `update-stack` after this change will replace the target group and ECS service (one-time).**

Adding `Name` to a TG that previously had no explicit name forces CloudFormation to replace it. Because the ECS service's `LoadBalancers` property references the TG ARN (which changes on replacement), the ECS service is also replaced. The replacement chain is:

`HTTPTargetGroup` replaced → `ECSServiceHTTP`/`ECSServiceHTTPS` replaced

This is a **graceful replacement**: CloudFormation creates the new resources before deleting the old ones (create-before-delete). There will be a brief overlap where old and new tasks run simultaneously during ECS task draining. Expect ~60 seconds of overlap while tasks deregister and re-register. No data loss occurs.

This one-time cost is the tradeoff for preventing the unrecoverable `400` error on all future LB replacements.

## Test plan

- [ ] Deploy a fresh stack — verify TG is created with name `tg-<hex-id>`
- [ ] Update the stack with a change that forces LB replacement (e.g. add/remove a subnet) — verify both the LB and TG are replaced together with no `400` error
- [ ] Verify ECS service health checks pass after update
- [ ] Update an **existing** stack (no LB replacement) — verify only TG+ECS service are replaced (one-time migration), no `400` error

### DIFF
--- a/deployment/infrastructure/otel-collector.yaml
+++ b/deployment/infrastructure/otel-collector.yaml
@@ -226,9 +226,18 @@ Resources:
   HTTPTargetGroup:
     Type: AWS::ElasticLoadBalancingV2::TargetGroup
     Properties:
-      # Name omitted intentionally — CloudFormation auto-generates a unique name
-      # that respects the 32-char limit. Previously used !Sub '${AWS::StackName}-tg'
-      # which failed when stack names exceeded 29 characters (issue #86).
+      # Name is derived from the ALB's unique hex ID (last segment of its ARN).
+      # When the LoadBalancer is replaced its unique ID changes, so this name
+      # changes too — CloudFormation replaces the target group alongside the LB.
+      # Without this coupling, CF tries to attach the old TG to the new LB while
+      # the old LB still owns it, producing:
+      #   "target group cannot be associated with more than one load balancer"
+      # Using the ALB ID keeps the name within the 32-char limit: "tg-" (3) +
+      # 16-char hex ID = 19 chars max. (issue #86 used !Sub '${AWS::StackName}-tg'
+      # which exceeded the limit for long stack names; this approach avoids that.)
+      Name: !Sub
+        - 'tg-${AlbId}'
+        - AlbId: !Select [3, !Split ['/', !Ref LoadBalancer]]
       Port: 4318
       Protocol: HTTP
       TargetType: ip
--- a/source/tests/test_cfn_naming_limits.py
+++ b/source/tests/test_cfn_naming_limits.py
@@ -87,11 +87,15 @@ class TestCFResourceNamingLimits:
             pytest.skip("No CF templates found")
         return templates
 
-    def test_no_explicit_target_group_name(self, templates):
-        """Target groups must NOT have explicit Name (32-char limit too easy to exceed).
+    def test_no_stack_name_derived_target_group_name(self, templates):
+        """Target groups must NOT derive Name from ${AWS::StackName} (32-char overflow).
 
         With stack names like 'claude-code-auth-otel-collector' (31 chars),
-        any suffix pushes past 32. Let CloudFormation auto-generate.
+        any suffix pushes past 32. Names derived from other sources (e.g. an
+        ALB's unique hex ID via !Sub) are acceptable when provably within limits.
+
+        Allowed: !Sub with resource-ID references (bounded length, no StackName).
+        Forbidden: plain strings or !Sub '${AWS::StackName}-*' patterns.
         """
         violations = []
         for template_path in templates:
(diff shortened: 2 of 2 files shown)


wirjo:
Fix looks correct — the `!Select [3, !Split ['/', !Ref LoadBalancer]]` approach is sound and stays within the 32-char TG name limit.

**One concern: backwards compatibility for existing stacks.**

Adding `Name` to an existing TG that had no explicit name will force CloudFormation to **replace** it on the next `update-stack`. This means:
- First update after this change triggers a TG replacement (one-time)
- Brief traffic disruption while ECS tasks deregister from old TG and register on new one

Could you:
1. Confirm this has been tested on an **existing** stack (not just fresh deploy)?
2. Add a note to the PR or changelog warning users about the one-time TG replacement on first update?

Otherwise LGTM 👍

HristoStoyanovYotpo:
Thanks for the review @wirjo!

**On existing-stack testing:** I haven't run this against a live existing stack in a controlled environment — the fix came from hitting the error in production and reasoning through the CF replacement chain. To be transparent: the mechanics are sound (adding `Name` to a TG that previously had none will force CF to replace it on next `update-stack`), but I can't claim a verified existing-stack test run.

**On the migration note:** Good call — I've updated the PR body to include an explicit upgrade note warning users about the one-time TG+ECS service replacement on first update after this change.

wirjo:
Thanks @HristoStoyanovYotpo — really appreciate this contribution! It's fantastic to see AWS customers who are running this in production identifying edge cases and contributing fixes back. Community contributions like yours make the project better for everyone in the ecosystem.

The analysis here is excellent: clean root cause identification, a minimal fix with proper justification for the naming strategy (staying within the 32-char TG name limit), and honest documentation of the upgrade impact. The upgrade note alone saves other operators from a nasty surprise.

Reviewing now — will get this merged shortly.

wirjo (changes requested):
Thanks for this fix @HristoStoyanovYotpo — the approach is sound and the CFN syntax is correct. The ALB-hex-ID coupling elegantly solves the dual-LB-association error.

However, there's a **blocking test failure**:

```
FAILED source/tests/test_cfn_naming_limits.py::TestCFResourceNamingLimits::test_no_explicit_target_group_name
```

This existing guard test (added to prevent issue #86 regression) asserts that **no** TargetGroup resource may have an explicit `Name` property. Your PR adds one — intentionally and safely (≤19 chars) — but the test doesn't distinguish between safe ALB-ID-derived names and unsafe stack-name-derived names.

**To fix:** Please update `source/tests/test_cfn_naming_limits.py` to accommodate this pattern. Options:

1. **Allowlist approach** — skip the violation if the Name value uses `!Select`/`!Split` (i.e., it's derived from a resource ARN, not `${AWS::StackName}`)
2. **Replace the blanket check** with a smarter assertion that the Name is guaranteed ≤32 chars (e.g., verify the pattern produces at most N chars)
3. **Remove that specific assertion** and rely on the `test_no_target_group_overflow_with_max_name` test (which still passes) plus the explanatory comments in the template

Option 1 or 2 would be ideal since they preserve the guard while allowing the new pattern.

Everything else looks great:
- ✅ cfn-lint passes
- ✅ 95/95 CFN contract tests pass (except the naming limits one)
- ✅ No circular dependencies
- ✅ All `!Ref HTTPTargetGroup` usages are ARN-based, unaffected
- ✅ No other files need changes

wirjo (approved):
✅ Approved — pushed a test fix to allow ALB-ID-derived TG names in the naming-limit guard. All tests pass (naming limits: 6/6, CFN contracts: 95/95, cfn-lint clean). The fix is minimal, correct, and well-documented. Ship it.

PR #650, opened 2026-06-23
feat(init): add IAM Identity Center configuration prompts to init wizard

wirjo:
## Problem

When a user selects "IAM Identity Center" in the `ccwb init` auth method step, no further configuration is collected. The user is left to manually configure `~/.aws/config` with SSO settings. This is a regression from commit `a777677` which reverted the IDC-specific init prompts originally introduced in PR #462.

## Solution

Add a guided IDC configuration section to the init wizard that collects:
- **SSO start URL** (e.g., `https://company.awsapps.com/start`)
- **SSO region** (auto-suggested from start URL domain)
- **AWS account ID** (validated as 12-digit)
- **Permission set name** (defaults to `BedrockDeveloperAccess`)

These values are persisted to the profile config for use by `deploy` and `credential-process`.

## Changes

- `source/claude_code_with_bedrock/cli/commands/init.py` — Add IDC config wizard block between auth selection and OIDC config
- `source/claude_code_with_bedrock/config.py` — Add `sso_region` field to Profile dataclass
- Save IDC fields (`auth_type`, `idc_start_url`, `idc_account_id`, `idc_permission_set_name`, `sso_region`) in wizard_fields

## Risk

**Low** — Only triggers when `auth_type == "idc"` is selected. No changes to OIDC or None paths. Config fields already existed in dataclass (except `sso_region` which is additive).

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -401,9 +401,7 @@ class InitCommand(Command):
 
             # SSO region (auto-suggest from start URL if possible)
             suggested_region = "us-east-1"
-            import re as _re
-
-            _region_match = _re.search(r"\.(us|eu|ap|sa|ca|me|af|il)-[a-z]+-\d+\.", idc_start_url)
+            _region_match = re.search(r"\.(us|eu|ap|sa|ca|me|af|il)-[a-z]+-\d+\.", idc_start_url)
             if _region_match:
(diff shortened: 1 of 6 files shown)


wirjo:
@qing-at-work FYI — this adds the IDC init wizard prompts (start URL, SSO region, account ID, permission set) that were missing after the revert in the feature branch. Now merged to beta. Users selecting 'IAM Identity Center' in `ccwb init` will get guided config collection instead of having to manually set up `~/.aws/config`.

PR #641, opened 2026-06-22
feat(websearch): wire AgentCore web search gateway into ccwb deploy/destroy

chiararelandini:
## Why

Second PR of the Claude Cowork web search feature. PR #607 ships the AgentCore Gateway + Web Search connector CloudFormation template; this PR wires it into the `ccwb` CLI so an administrator can opt in and have `ccwb deploy` / `ccwb destroy` manage the gateway, reusing the existing SSO/IdP.

> ⚠️ **Depends on #607 — please merge that first.** This PR's deploy code references `deployment/infrastructure/bedrock-agentcore-gateway.yaml` (added in #607) at deploy time. The diff here is intentionally code-only (the template is not duplicated). Unit tests don't require the template; a real `ccwb deploy websearch` does, which is why #607 must land first.

## What

**`config.py`** *(Tier 1)*
- New opt-in `Profile` fields, all defaulting off → fully backward compatible: `cowork_websearch_enabled`, `websearch_region`, `websearch_jwt_audience` (Entra ID only), `websearch_domain_denylist`.
- `WEBSEARCH_SUPPORTED_REGIONS = ["us-east-1"]` allow-list (extend as the connector's regional availability grows).

**`deploy.py`** *(Tier 1)*
- Register a new `websearch` stack type (both `ccwb deploy websearch` and deploy-all).
- `websearch_preflight()` guard: only `cognito`/`azure`, region must be in the allow-list, Entra ID requires an audience — clear error, no partial deploy.
- Authorizer params derived from the existing IdP config: `JwtValidationMode=client_id` + `AllowedClients` (Cognito), `audience` + `AllowedAudience` (Entra ID); discovery URL built per provider.
- Deployed **cross-region into `us-east-1`** via a region-bound CloudFormation client (mirrors the existing CodeBuild cross-region pattern). Orphaned-stack detection covers the (possibly cross-region) gateway.

**`destroy.py`**
- `websearch` added to teardown, deleted in its own region when it differs from the main region.

### Test Environment
- **Python**: 3.10
- **OS**: macOS (local), CI matrix covers Linux/macOS/Windows

### Test Results
- New `tests/test_websearch_gateway.py` (15 tests): backward-compat profile load, provider/region guards (cognito/azure pass; okta blocked; azure-without-audience blocked; bad region blocked), discovery-URL derivation, and CFN param derivation.
[...]

PR #640, opened 2026-06-22
feat(websearch): hosted web search via AgentCore gateway (provider-agnostic id_token)

antkawam:
## Why

Enterprises want Claude Code to answer "what happened today" questions, but there's
no built-in web search on Bedrock. This adds an optional hosted web-search tool via
an Amazon Bedrock AgentCore gateway, authorized by the **same OIDC identity the
deployment already mints** — no proxy, no extra IAM grant, no per-user setup.

It works for **any OIDC provider** (Cognito, Okta, Auth0, Entra, Google) with zero
customer-side IdP config, because the gateway validates the uniform OIDC **id_token**
by matching its `aud` claim (which every provider sets to the client_id).

## What

- **CFN template** (`agentcore-websearch.yaml`): AgentCore gateway (MCP, CUSTOM_JWT
  authorizer matching `AllowedAudience`), web-search connector target, least-privilege
  service role. Pinned to us-east-1 (the managed connector's only region).
- **credential-process** (`--get-mcp-auth-header`, Go + Python, identical output):
  emits `{"Authorization":"Bearer <id_token>"}` for the MCP headersHelper; never opens
  a browser.
- **deploy / package / init / config**: region-pinned deploy with READY polling +
  gateway-URL persistence; `mcpServers.agentcore-websearch` block emitted on package;
  optional init prompt (discloses ~$7/1k-query cost + us-east-1); two new Profile fields
  (backward compatible).
- **Docs**: `WEB_SEARCH.md` (cost, region, acceptable-use citation retention, install
  caveat) + cross-links.

## Verification

- **Live, end-to-end on two providers:**
  - **Cognito** via the Go binary: `tools/list` + `tools/call` → 200 with cited results.
  - **Google** (real OAuth id_token): `tools/list` + `tools/call` → 200 with cited
    results. Google issues opaque access_tokens, so this case only works via the
    id_token + audience-match path.
- Reviewed by both same-model personas and a cross-model pass (GPT-5.5); both surfaced
  one real defect (deploy exit-code on unreadable outputs), now fixed with a regression
  test.
- Gates: Python 1254 passed; Go build/test/vet clean; cfn-lint clean.

## Known follow-ups (not blocking)

- **Python `--get-mcp-auth-header` has no silent refresh** (Go does). Output-byte
  parity holds; the shipped binary is Go on all platforms. Documented; deferred.
- **Google dual-issuer-form**: a manually minted bare-`iss` token (gcloud) 403s; the
[...]

PR #607, opened 2026-06-19
feat(websearch): AgentCore Gateway + Web Search connector CloudFormation template

chiararelandini:
## Why

Claude Cowork on Amazon Bedrock cannot use the native Anthropic web search tool (Bedrock does not expose it), so Cowork users have no way to ground answers in current web data. Amazon Bedrock AgentCore now offers a fully managed, MCP‑compliant [Web Search connector](https://aws.amazon.com/blogs/aws/announcing-web-search-on-amazon-bedrock-agentcore-ground-your-ai-agents-in-current-accurate-web-knowledge/) that keeps queries inside AWS with no third‑party search API or outbound credentials. This is the maintainer‑endorsed workaround from #363, and #374/#514 already added the custom‑MDM‑key mechanism needed to surface it to Cowork.

This is the **first PR** of a multi‑PR feature that lets administrators optionally add web search to Claude Cowork, reusing the solution's existing SSO. It adds only the infrastructure template — no CLI wiring yet — so it is self‑contained and safe to review/merge on its own.

## What

Adds `deployment/infrastructure/bedrock-agentcore-gateway.yaml`, an opt‑in stack that provisions:

- **`AWS::BedrockAgentCore::Gateway`** — MCP protocol, `CUSTOM_JWT` authorizer whose `DiscoveryUrl` + `AllowedClients`/`AllowedAudience` are supplied as parameters so it reuses the **same OIDC IdP** as the rest of the deployment.
- **`AWS::BedrockAgentCore::GatewayTarget`** — the managed `web-search` connector, with an optional server‑side domain denylist.
- **Gateway execution IAM role** — least‑privilege: `GetGateway`, `GetConfigurationBundleVersion`, `InvokeWebSearch`, with `SourceAccount`/`SourceArn` confused‑deputy guards on the trust policy.

Key design points:
- `ProviderType` condition selects `AllowedClients` (Cognito) vs `AllowedAudience` (Entra ID).
- `WebSearchRegion` is constrained via `AllowedValues` to where the connector is GA (**`us-east-1`** today); the list is meant to grow as availability expands.
- No hardcoded resource names (`!Sub '${AWS::StackName}-*'`), `${AWS::Partition}`/`${AWS::Region}`/`${AWS::AccountId}` in ARNs.

### Note for reviewers — cfn-lint suppression
[...]

PR #651, opened 2026-06-23
fix(websearch): MCP auth-header helper — Go/Python refresh parity + expired-token handling

antkawam:
## Why

The `--get-mcp-auth-header` mode of the credential-process is the `headersHelper` for the AgentCore web-search MCP server — Claude Code calls it to supply `{"Authorization":"Bearer <id_token>"}` to the gateway's CUSTOM_JWT authorizer. Two gaps from review:

- **Go discarded a valid refreshed token on unrelated failures.** `getMCPAuthHeader` refreshed an expired id_token via `tryRefreshToken`, which returns `nil` unless it *also* exchanges the token for AWS credentials. A successful id_token refresh could be thrown away on an unrelated Cognito/STS/IAM error — even though the gateway only needs the id_token.
- **Python gave no signal on an expired token.** Python has no refresh_token store, so a present-but-expired token failed identically to "never authenticated", with no hint which case the user was in.

## What

- **Go:** add `refreshIDTokenOnly()` — the same browserless refresh_token exchange and monitoring-token persistence as `tryRefreshToken`, but returns the fresh id_token *without* requiring AWS credential exchange. `getMCPAuthHeader` uses it; the main credential path keeps `tryRefreshToken` unchanged. (Found by cross-model Codex review.)
- **Python:** factor the raw token-blob load into `_load_monitoring_token_data` (no expiry check) so the header path can tell "expired" from "absent". `get_mcp_auth_header` now emits a one-line stderr hint when the cached token is present-but-expired, then still returns `None` (no browser, no hang). The next full credential-process run — which Claude Code invokes routinely — re-mints and re-caches. Full silent-refresh parity is a larger cross-cutting change, documented in the docstring.

## Tests

- Go: refresh succeeds against a mock token endpoint with no STS configured → header emitted, exit 0 (red under the old coupling); env-var token uses a valid JWT.
- Python: expired-but-present → `None` + "expired" hint; absent → `None`, no hint; end-to-end session-storage expiry-boundary tests against the real storage code.
- Output contract unchanged (`{"Authorization":"Bearer <id_token>"}`); Go↔Python parity test green.

## Scope

First of a stacked series carving the web-search feature (#640) down to the Claude Code surface, atop the merged #607 template. This PR is the credential-helper layer — independent of the deploy path (#641). Config/wizard, package emit, and docs follow as separate PRs.

Python suite: 1206 passed. Go build/vet/test + gofmt clean.

### DIFF
--- a/source/credential_provider/__main__.py
+++ b/source/credential_provider/__main__.py
@@ -899,6 +899,19 @@ class MultiProviderAuth:
         except Exception:
             return None
 
+    def get_mcp_auth_header(self):
+        """Return the MCP Authorization header dict from the cached id_token, or None.
+
+        {"Authorization": "Bearer <id_token>"} for the AgentCore web-search MCP
+        headersHelper. Reads only the cached/silently-refreshed token via
+        get_monitoring_token() — never opens a browser. Returns None when no
+        valid token is available so the caller fails cleanly (no hang, no prompt).
+        """
+        token = self.get_monitoring_token()
+        if not token:
+            return None
+        return {"Authorization": f"Bearer {token}"}
+
     def save_to_credentials_file(self, credentials, profile="ClaudeCode"):
         """Save credentials to ~/.aws/credentials file
 
@@ -2471,6 +2484,14 @@ def main():
         "--get-monitoring-token", action="store_true", help="Get cached monitoring token instead of AWS credentials"
     )
     parser.add_argument(
+        "--get-mcp-auth-header",
+        action="store_true",
+        help=(
+            'Print {"Authorization":"Bearer <id_token>"} from the cached token for an MCP '
+            "headersHelper (never opens a browser)"
+        ),
+    )
+    parser.add_argument(
         "--clear-cache", action="store_true", help="Clear cached credentials and force re-authentication"
     )
     parser.add_argument(
@@ -2559,6 +2580,27 @@ def main():
                 # Return failure exit code so OTEL helper knows auth failed
                 # This prevents OTEL helper from using default/unknown values
                 sys.exit(1)
+
+    # Handle MCP auth-header request.
+    # Used as the headersHelper for the AgentCore web-search MCP server, whose
+    # gateway runs a CUSTOM_JWT authorizer validating the same OIDC id_token the
+    # solution already mints. MUST be fast and MUST NOT open a browser — an MCP
+    # headersHelper can never drive an interactive login. So, unlike
+    # --get-monitoring-token, this never falls through to authentication: on a
+    # cache miss it fails cleanly with a non-zero exit.
+    if args.get_mcp_auth_header:
+        header = auth.get_mcp_auth_header()
+        if header:
+            # Compact separators so Go (encoding/json) and Python emit byte-identical output.
+            print(json.dumps(header, separators=(",", ":")))
+            sys.exit(0)
+        else:
+            print(
+                f"Error: no valid cached token for profile '{args.profile}'; "
+                "run the credential process once to authenticate.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
 
     # Handle check-expiration request
     if args.check_expiration:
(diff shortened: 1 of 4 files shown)


wirjo:
## Heads up: Overlap with PR #666

PR #666 (fix: quota recheck + Azure refresh 404) extracts a shared `provider.TokenEndpointURL()` builder that centralizes domain normalization (e.g. stripping Azure's trailing `/v2.0`) so the auth flow and refresh flow can't drift apart.

The new `refreshIDTokenOnly()` method in this PR builds the token URL inline:

```go
provCfg := provider.ConfigFor(a.providerType, a.cfg.OktaAuthServerID)
if strings.HasPrefix(provCfg.TokenEndpoint, "https://") {
    tokenURL = provCfg.TokenEndpoint
} else {
    tokenURL = "https://" + a.cfg.ProviderDomain + provCfg.TokenEndpoint
}
```

This has the same Azure `/v2.0` doubling bug that #666 fixes — Azure deployments would get HTTP 404 on the MCP auth-header refresh path.

**Once #666 merges**, this can be simplified to:

```go
tokenURL = provider.TokenEndpointURL(a.providerType, a.cfg.OktaAuthServerID, a.cfg.ProviderDomain)
```

Not blocking — just flagging so this doesn't re-introduce the bug when rebased.

wirjo:
Rebased onto latest beta (includes #666) and replaced the inline token URL construction in `refreshIDTokenOnly()` with the shared `provider.TokenEndpointURL()` builder. This fixes the Azure `/v2.0` doubling bug that was flagged earlier — the MCP auth-header refresh path now uses the same URL resolution as `tryRefreshToken()`.

Changes in the force-push:
- Clean rebase (4 commits, no conflicts)
- `refreshIDTokenOnly()`: replaced 7 lines of inline URL logic with 1 call to `provider.TokenEndpointURL()`
- Net: +3/-7 lines in that function

wirjo:
## Code Path Trace — Backward Compatibility Assessment

### New Code Paths (fully isolated from existing flows):
- **Go:** `getMCPAuthHeader()` — only reachable via `--get-mcp-auth-header` flag, dispatched in flag section before `run()` is ever called
- **Go:** `refreshIDTokenOnly()` — only called from `getMCPAuthHeader()`, never from the main credential path (`run()`, `trySilentRefresh`, `tryRefreshToken`, `authenticate`)
- **Python:** `get_mcp_auth_header()` — new method on `MultiProviderAuth`, only via `--get-mcp-auth-header` arg
- **Python:** `_load_monitoring_token_data()` — private helper extracted from existing `get_monitoring_token()` internals

### Modified Existing Code:
- **Python `get_monitoring_token()`** — refactored (not behaviorally changed). Storage-reading logic extracted into `_load_monitoring_token_data()`. Same env-var precedence, same 60s expiry buffer, same exception handling. Return contract identical.
- **Go `main.go`** — one cosmetic whitespace removal. No functional change.

### Shared Code Risk:
`refreshIDTokenOnly()` shares storage reads/writes with `tryRefreshToken()` (`LoadRefreshToken`, `SaveRefreshToken`, `ClearRefreshToken`, `saveMonitoringTokenAndHeaders`). However:
- Mutually exclusive at runtime (flag dispatch ensures only one runs per invocation)
- All shared calls are idempotent
- KEY DIFFERENCE: `refreshIDTokenOnly` does NOT call `getAWSCredentials()` or `saveCredentials()` — intentionally decouples id_token refresh from STS exchange

### Can Existing Users Be Affected? **NO**
1. `--get-mcp-auth-header` flag never activated unless explicitly passed
2. Python refactoring preserves identical semantics (same try/except flow)
3. No existing function signatures changed in Go or Python
4. `refreshIDTokenOnly()` is never called from any existing path

### Verdict: ✅ Safe to merge (high confidence)

Minor observations:
- `refreshIDTokenOnly` triggers `ensureProxyRunning` via `saveMonitoringTokenAndHeaders` — harmless (proxy would start on next regular credential call anyway)
- Python refactoring splits exception handling into two layers — same end result, slightly different debug stack traces

wirjo:
Thanks @antkawam — clean implementation, well-isolated from existing flows. Merged to beta.

Issue #515, opened 2026-06-10
Terminal warnings are not visible after quota threshold hits.


adiospeds:
I local branch is at git commit "ccc46ea7bea7c7873702a382a6a96334ac3896bb" and I am using the OIDC path with Azure AD for quota enforcement. I was able to get the quota enforcement working by setting a 500 Daily and 1K monthly token limit on my user

Also the browser shows quota enforced when I hit the limit: 

<img width="1022" height="822" alt="Image" src="https://github.com/user-attachments/assets/7535b02e-dd09-4dd6-a70b-91baeddce2cb" />

But there were no terminal warnings as shown in below doc:

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/ccc46ea7bea7c7873702a382a6a96334ac3896bb/assets/docs/QUOTA_MONITORING.md#terminal-output

Using macOS Tahoe 26.3.1 (a) (25D771280a) - and iterm2 as my terminal.

wirjo:
This may be fixed in #515. 

@claude Review this issue.

wirjo:
Fixed by #516 (merged to beta). Quota warning now outputs to stderr when usage exceeds 80% threshold, with usage details and policy info.

PR #657, opened 2026-06-23
fix(go): emit quota warning on all 4 auth paths (fixes #655)

### LINKED ISSUE #655: Bug: Go credential-process quota warnings rarely fire and are invisible under Claude Code
## Summary

Users on the Go credential-process binaries do not see quota warnings as they approach or
exceed their limits, even though the equivalent PyInstaller (Python) binary did. A user on
Azure OIDC + central collector + Go binaries (enforcement = `block`) ran their **daily token
quota to 451.9%** without any warning surfacing, and only discovered it when an admin ran
`ccwb quota usage`:

```
Policy: default:default
Enforcement: block

Metric         Current   Limit   Used %
Tokens            9.1M   40.0M    22.7%
Daily Tokens      9.0M    2.0M   451.9%     <-- 451% with no warning shown
Cost             $1.81       —        —
```

With the old PyInstaller binary the same situation popped a browser "⚠️ Quota Warning" page
with progress bars. After switching to the Go binaries that notification disappeared.

---

## Problem

PR #516 (`feat(go): add quota warning output for 80%+ usage (Python parity)`, commit `0d75fb8`)
added `printQuotaWarning()` to the Go credential-process. In current code there are three
compounding gaps that keep that warning from ever reaching a Claude Code user.

### 1. The warning is wired into only 1 of 4 quota-check paths

`run()` checks quota in four places. The Python provider calls `_handle_quota_warning()` in all
four; the Go binary calls `printQuotaWarning()` only in the periodic-recheck path.

| Quota-check path | Python warns? | Go warns? |
|---|---|---|
| Periodic recheck on cached creds (`shouldRecheckQuota`) | ✅ `__main__.py:2320` | ✅ `main.go:770` |
| After silent refresh (id_token still valid) | ✅ `__main__.py:2382` | ❌ **missing** — `main.go:425-435` |
| After refresh_token exchange | ✅ (warn path) | ❌ **missing** — `main.go:444-454` |
| After fresh browser auth | ✅ `__main__.py:2407` | ❌ **missing** — `main.go:468-475` |

The three Go sites only handle the *blocked* case and fall through to `outputJSON(creds)` when
quota is allowed-but-over-threshold:

```go
// main.go:468-475 — fresh auth path (silent-refresh and refresh-token paths are identical)
if a.cfg.QuotaAPIEndpoint != "" {
    qr := quota.Check(a.cfg.QuotaAPIEndpoint, authResult.IDToken, a.cfg.QuotaCheckTimeout, a.cfg.QuotaFailMode)
    if !qr.Allowed {
        printQuotaBlocked(qr)
        return 1
    }
    _ = storage.SaveQuotaState(a.profile)
    // <-- no printQuotaWarning(qr) here
}
```

### 2. The one path that warns is sampled at most every 30 minutes

The only warning path is the periodic recheck, gated by `shouldRecheckQuota()`
(`main.go:740-752`), which fires only when `time.Since(lastCheck) >= QuotaCheckInterval`. The
default interval is **30 minutes** (`source/go/internal/config/config.go:207`).

STS credentials are valid for hours, so during an active session most credential-process
invocations hit the warm-cache fast path (`main.go:385`) with the interval not yet elapsed → no
quota check. When credentials do expire mid-session, the binary takes the silent-refresh /
refresh-token paths — exactly the paths that don't warn (gap 1). A user can therefore run far
over a **daily** limit (which a 30-minute sample barely tracks) without crossing the one code
path that emits a warning.

### 3. Even when it fires, stderr is invisible under Claude Code (ref #515)

`printQuotaWarning()` (`main.go:854`) writes a text banner to the credential-process **stderr**.
The AWS SDK inside Claude Code spawns credential-process, consumes stdout (the JSON
credentials), and the subprocess **stderr is not surfaced in the Claude Code TUI** — the concern
raised in #515. The Python binary avoided this by *also* opening a browser page
(`_show_quota_browser_notification`, `__main__.py:1948`), which does not depend on a visible
terminal. The Go binary has **no browser-notification equivalent** — `internal/oidc/callback.go`
serves HTML for the OAuth callback only, with no quota display.

---

## Expected vs. actual

- **Expected:** Crossing 80% (monthly or daily) produces a visible warning, as the PyInstaller
  binary did via the browser page.
- **Actual (Go):** No visible warning on any normal session path. The only warning is a stderr
  banner emitted at most every 30 minutes on the periodic-recheck path, and that stderr is not
  shown in Claude Code.

---

## Steps to reproduce

1. Deploy with quota enforcement and the Go credential-process binaries.
2. As a user, drive token usage past 80% (and beyond 100%) of the daily limit within a session.
3. Observe: no warning surfaces in Claude Code; admin `ccwb quota usage` confirms the overage.
4. Repeat with the old PyInstaller binary → browser "⚠️ Quota Warning" page appears.

---

## Proposed fix

**Minimum (parity):** Call `printQuotaWarning(qr)` in the three Go paths that currently handle
only the blocked case — `main.go:425-435`, `:444-454`, `:468-475` — so a warning is evaluated on
every refresh, not only on the 30-minute periodic recheck.

**Real fix (user-visible):** Port the Python `_show_quota_browser_notification` to Go — serve the
warning HTML from a transient local HTTP server (Python uses `redirectPort+1`) and open it via
the existing `internal/browser` `OpenURL()` (`source/go/internal/browser/open.go:13`). This is
what made the warning visible before. Throttle it (e.g. once per threshold crossing, keyed off
the persisted quota state) so it does not pop a browser on every credential refresh.

---

## Testing / review notes

`main.go` is a **Tier 1 critical** file (`.claude/rules/review-tiers.md`). A fix should include:

- A regression test for each of the 4 quota-check paths asserting a warning is emitted when
  `allowed=true` and `monthly_percent >= 80` or `daily_percent >= 80`.
- A Go ↔ Python parity test (`.claude/rules/credential-helper-parity.md`) — identical usage
  payload must produce a warning in both binaries.
- Coverage across auth types: oidc, idc, none.
- stdout stays sacred — the warning / HTML server must never write to stdout
  (`.claude/rules/credential-flow.md`).

---

## Affected files

| File | Status |
|------|--------|
| `source/go/cmd/credential-process/main.go` — `:425-435`, `:444-454`, `:468-475` | Missing `printQuotaWarning` call |
| `source/go/cmd/credential-process/main.go` — `printQuotaWarning()` `:854` | stderr only; not visible in Claude Code |
| `source/go/internal/browser/open.go` — `OpenURL()` `:13` | Available; no quota page served yet |
| `source/credential_provider/__main__.py` — `_handle_quota_warning()` `:2182`, `_show_quota_browser_notification()` `:1948` | Reference implementation |

---

## References

- PR #516 — `feat(go): add quota warning output for 80%+ usage (Python parity)` (commit `0d75fb8`)
- Issue #515 — stderr-only warning is invisible under Claude Code
- Default recheck interval (30 min): `source/go/internal/config/config.go:207`

wirjo:
## Why

Go credential-process users get **no visible quota warning** when approaching or exceeding limits. A user on Azure OIDC + Go binaries ran their daily quota to 451.9% without any notification — only discovered by admin running `ccwb quota usage`.

Root cause: `printQuotaWarning()` was only called in 1 of 4 quota-check paths (the periodic 30-minute recheck). The three other paths — silent refresh, refresh_token exchange, and fresh browser auth — only handled the `blocked` case and fell through to `outputJSON(creds)` without warning.

### Go fix (`main.go`)
Added `printQuotaWarning(qr)` to the three missing paths:
- **Silent refresh path** (line ~432): After `trySilentRefresh()` succeeds
- **Refresh token path** (line ~450): After `tryRefreshToken()` succeeds  
- **Fresh browser auth path** (line ~472): After `authenticate()` succeeds

### Tests
- **`quota_warning_test.go`** (Go): Verifies `printQuotaWarning` emits to stderr at 80%+ threshold and stays silent below. Tests nil/empty usage for safety.
- **`test_quota_warning_parity.py`** (Python): 6 parity tests confirming the Python reference implementation behavior matches expectations (baseline for Go).

## What this does NOT fix

Gap 3 from #655: stderr is still invisible under Claude Code. The browser notification port (Python's `_show_quota_browser_notification`) is not yet implemented in Go. That's a separate enhancement tracked in #515.

## Testing

- Python tests: 6/6 passing locally
- Go tests: require Go 1.24 (CI only — local env has 1.22)
- No behavioral change for users under 80% — only adds warning output

Fixes #655
Credit: @adiospeds for the excellent bug report and root cause analysis

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -435,7 +435,33 @@ class DeployCommand(Command):
 
             # Deploy based on stack type
             if stack_type == "auth":
-                # Select template based on provider type
+                # IAM Identity Center uses a dedicated template and different parameters
+                if profile.effective_auth_type == "idc":
+                    template = project_root / "deployment" / "infrastructure" / "bedrock-auth-idc.yaml"
+                    stack_name = profile.stack_names.get("auth", f"{profile.identity_pool_name}-stack")
+
+                    from claude_code_with_bedrock.models import get_all_bedrock_regions
+
+                    bedrock_regions = profile.allowed_bedrock_regions
+                    if not bedrock_regions:
+                        bedrock_regions = [r for r in get_all_bedrock_regions() if "gov" not in r]
+
+                    idc_role_name = getattr(profile, "idc_permission_set_name", None) or "BedrockIDCFederatedRole"
+                    params = [
+                        f"FederatedRoleName={idc_role_name}",
+                        f"IdentityPoolName={profile.identity_pool_name}",
+                        f"AllowedBedrockRegions={','.join(bedrock_regions)}",
+                        f"EnableMonitoring={str(profile.monitoring_enabled).lower()}",
+                    ]
+                    return deploy_with_cf(
+                        template,
+                        stack_name,
+                        params,
+                        capabilities=["CAPABILITY_NAMED_IAM"],
+                        task_description="Deploying IAM Identity Center auth stack...",
+                    )
+
+                # OIDC providers
                 provider_type = profile.provider_type or "okta"
                 template_map = {
                     "okta": "bedrock-auth-okta.yaml",
@@ -949,7 +975,11 @@ class DeployCommand(Command):
 
                     # Deploy the packaged template
                     result = deploy_with_cf(
-                        packaged_template_path, stack_name, params, task_description="Deploying quota monitoring..."
+                        packaged_template_path,
+                        stack_name,
+                        params,
+                        capabilities=["CAPABILITY_NAMED_IAM"],
+                        task_description="Deploying quota monitoring...",
                     )
 
                     # Seed default quota policy on successful deploy
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -435,33 +435,7 @@ class DeployCommand(Command):
 
             # Deploy based on stack type
             if stack_type == "auth":
-                # IAM Identity Center uses a dedicated template and different parameters
-                if profile.effective_auth_type == "idc":
-                    template = project_root / "deployment" / "infrastructure" / "bedrock-auth-idc.yaml"
-                    stack_name = profile.stack_names.get("auth", f"{profile.identity_pool_name}-stack")
-
-                    from claude_code_with_bedrock.models import get_all_bedrock_regions
-
-                    bedrock_regions = profile.allowed_bedrock_regions
-                    if not bedrock_regions:
-                        bedrock_regions = [r for r in get_all_bedrock_regions() if "gov" not in r]
-
-                    idc_role_name = getattr(profile, "idc_permission_set_name", None) or "BedrockIDCFederatedRole"
-                    params = [
-                        f"FederatedRoleName={idc_role_name}",
-                        f"IdentityPoolName={profile.identity_pool_name}",
-                        f"AllowedBedrockRegions={','.join(bedrock_regions)}",
-                        f"EnableMonitoring={str(profile.monitoring_enabled).lower()}",
-                    ]
-                    return deploy_with_cf(
-                        template,
-                        stack_name,
-                        params,
-                        capabilities=["CAPABILITY_NAMED_IAM"],
-                        task_description="Deploying IAM Identity Center auth stack...",
-                    )
-
-                # OIDC providers
+                # Select template based on provider type
                 provider_type = profile.provider_type or "okta"
                 template_map = {
                     "okta": "bedrock-auth-okta.yaml",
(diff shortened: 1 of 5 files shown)


adiospeds:
Thanks @wirjo for the quick turn around. Can we also implement the `_show_quota_browser_notification` for golang so users get a browser notifications. The terminal warning is still invisible to users while using Claude code as it goes to stderr and users continue to run over their daily quota without any notifications. Basically #515 still remains incomplete without a visual warning to users.

wirjo:
Good catch @adiospeds — you're right, the stderr warning is invisible to Claude Code users.

Created PR #658 to port  to Go: opens a browser page with visual progress bars on quota warning/block, matching the Python credential-provider's behavior.

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/658

PR #658, opened 2026-06-24
feat(go): add browser-based quota notification (Python parity)

wirjo:
## Problem

Terminal/stderr quota warnings are invisible to Claude Code users (reported by @adiospeds in [PR #657 comment](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/pull/657#issuecomment-4784657032)). The credential-process binary writes to stderr, but the AWS SDK consumes stdout and discards stderr — users exceed their quota without any visual notification.

The Python credential-provider has `_show_quota_browser_notification()` which opens a browser page with progress bars. The Go binary lacked this.

## Why This Was Missed

The Go rewrite focused on credential-flow parity (OIDC auth, token caching, STS federation). The browser notification was a Python-only UX enhancement added later and was never tracked as a Go parity requirement in `.claude/rules/` or the rewrite checklist.

## Solution

Added `quota_notification.go` to the Go credential-process that:
1. Serves an HTML page on `localhost:8401` (matching Python port) with visual progress bars
2. Opens the user's default browser
3. Shows monthly + daily usage with color-coded progress (green → yellow → orange → red)
4. Auto-closes the HTTP server after page is served (or 5s timeout)

Called from both `printQuotaWarning` (80%+ threshold) and `printQuotaBlocked` (access denied).

## Behavior

- **Additive:** Browser notification fires AFTER stderr output (doesn't replace terminal warnings)
- **Non-blocking:** Runs with timeout — if browser can't connect, credential flow continues normally
- **Platform-aware:** Uses `open` (macOS), `xdg-open` (Linux), `rundll32` (Windows)
- **Port fallback:** Tries 8401, falls through to any available port if busy

## Risk

**Low** — additive feature, wrapped in error handling. If browser notification fails for any reason, credential flow continues unaffected (stderr warning still works as before).

Addresses: #515, feedback on #657

### DIFF
--- a/.claude/rules/credential-helper-parity.md
+++ b/.claude/rules/credential-helper-parity.md
@@ -34,3 +34,20 @@ When adding a new OIDC provider:
 3. Same hostname patterns and domain matching in both
 
 Example: PR #563 added Google to Go after it was already in Python — this gap should never happen.
+
+## UX Feature Parity
+
+When adding user-facing features (notifications, warnings, interactive flows)
+to either the Python or Go credential-process:
+
+- **Both variants must have the feature** — users may get either binary depending
+  on how they installed (Go via `ccwb package`, Python via pip/legacy)
+- Key UX features that must stay in sync:
+  - Quota warning display (stderr + browser notification)
+  - Quota blocked display (stderr + browser notification)  
+  - Auth flow (OIDC browser, IDC device-auth)
+  - Token refresh (silent vs interactive)
+- If a feature is Python-only or Go-only, add a comment explaining why
+  (e.g., platform constraint) and create a tracking issue
+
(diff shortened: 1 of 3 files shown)


wirjo:
@adiospeds This is now merged to beta. The Go binary will open a browser notification with progress bars on quota warning (80%+) or block — matching the Python behavior.

Key details:
- Non-blocking (goroutine) — does not delay credential output
- Auto-skips in headless environments (SSH, Docker, no DISPLAY)
- Opt-out via `CCWB_NO_BROWSER_NOTIFICATION=1` env var
- Falls back gracefully if browser cannot be opened

Let us know if this addresses your feedback from #515/#657!

PR #660, opened 2026-06-24
feat(monitoring): enable per-user identity for Cowork in central + sidecar modes

wirjo:
## Summary

Enables **per-user identity** for CoWork (Claude Desktop) telemetry by auto-spawning an otel-helper proxy from `credential-process`. The proxy injects user attribution headers (`x-user-email`, `x-department`, `x-team-id`) into OTLP requests, enabling per-user dashboards for CoWork deployments.

## Why

Currently, CoWork telemetry is device-level only — all OTLP metrics from a machine are anonymous. Claude Code CLI has per-user identity via `otelHeadersHelper` (the credential-process binary with `--get-monitoring-token`), but CoWork cannot use external helper scripts. This PR bridges that gap with a transparent proxy that Claude Desktop sends telemetry through.

## Architecture

```
Central:  CoWork → localhost:4318 (otel-helper proxy) → Remote ALB → ECS Collector → CloudWatch
Sidecar:  CoWork → localhost:4319 (otel-helper proxy) → localhost:4318 (otelcol) → CloudWatch
CLI:      Claude Code CLI → otel-helper (direct, unchanged) → CloudWatch
```

### otel-helper Role by Mode

| Mode | Surface | otel-helper role | Port | Upstream | Auth method |
|------|---------|-----------------|------|----------|-------------|
| Central | CoWork (Claude Desktop) | Identity proxy (passthrough) | 4318 | Remote ALB | Service token (X-Cowork-Token header) |
| Central | Claude Code CLI | JWT bearer + identity headers | N/A (piped via otlpHeadersHelper) | Remote ALB | JWT (monitoring_token via Authorization) |
| Sidecar | CoWork (Claude Desktop) | Identity proxy (passthrough) | 4319 | localhost:4318 (otelcol) | None (otelcol handles SigV4 to CloudWatch) |
| Sidecar | Claude Code CLI | SigV4 signer + identity headers | N/A (piped via otlpHeadersHelper) | CloudWatch OTLP direct | SigV4 |

### Proxy Lifecycle (Self-Healing)

The proxy is managed via `credential-process` — no LaunchAgent, systemd, or manual startup:

1. AWS SDK calls `credential-process` for Bedrock credentials
2. `credential-process` checks if proxy is listening (TCP port dial, 200ms timeout)
3. If not listening → spawns `otel-helper --proxy --proxy-upstream <url> --proxy-port <port>` as detached process
4. Proxy runs until killed or machine restarts
[...]

PR #671, opened 2026-06-25
fix(idc): reliable IDC auth, credential refresh, and Windows telemetry

qing-at-work:
## Why

IAM Identity Center (IDC) deployments — especially on headless / EC2 hosts — had several breakages that left Claude Code stuck, silently mis-attributed, or sending no telemetry. These were found and fixed iteratively while standing up an IDC + quota + monitoring deployment on EC2 (Linux and Windows):

- **Hourly credential expiry → instance-role fallback.** IDC role credentials expire at the permission set's session duration (~1h). Nothing re-invoked the credential process, so after ~1h the AWS SDK chain fell through to the EC2 **instance role** (wrong identity, broken attribution/quota) and Claude Code looped on `API error · Retrying`.
- **Rotating refresh-token race.** Claude Code fires concurrent credential-process calls (e.g. session-title + main message). They raced the **rotating** SSO refresh token; the loser presented a consumed token → `InvalidGrantException`, bricking the cache until manual re-login.
- **Silent hang / confusing error on expired session.** A genuinely expired SSO session caused a silent multi-minute hang (headless, incl. **non-SSH Windows**) or a jargon-heavy message.
- **Windows telemetry sent nothing.** `settings.json` pointed `otelHeadersHelper` at `otel-helper.cmd`, which could be absent on the host and was referenced with a cmd-unsafe forward-slash path → "is not recognized" → Claude Code dropped **all** metric exports.
- **No per-user dashboard attribution for IDC.** The helper that surfaces `x-user-email` was disabled for IDC even when the binary could resolve the email from the IAM ARN.

## What

**Credential refresh (Go + settings)**
- Wire `awsCredentialExport` (silent refresh, auto re-invoked ~5 min before `Expiration`) **and** `awsAuthRefresh --login` (the only hook whose output is displayed, so it surfaces the sign-in message) for IDC profiles.
- Gate device-auth on `tokenValid` **or** `tokenRefreshable`, so an expired-but-refreshable access token is refreshed silently via the SDK rather than forcing interactive login.
- Serialize the rotating-token refresh under an exclusive `portlock`, so concurrent credential-process invocations can't consume each other's refresh token.
- Set `AWS_EC2_METADATA_DISABLED=true` for IDC so a refresh failure surfaces a clear credentials error instead of silently using the instance role.

**Headless UX**
- Fail fast (don't hang) when stderr is not a terminal — the correct signal for "Claude Code invoked us non-interactively", covering **headless Windows without SSH**, not just SSH/Linux.
- Plain-language, no-jargon message with absolute paths, directing users to relaunch via the generated `claude-bedrock` launcher (which signs in, showing the URL live, then starts Claude Code).

**Windows telemetry**
- Point `otelHeadersHelper` at `otel-helper.cmd` (the AV-resilient wrapper: tries `.exe`, falls back to `.ps1`) using an **escaped-backslash** path that `cmd.exe` handles correctly.
- Fail the install **loudly** if the required `.cmd`/`.ps1` are missing when monitoring is enabled (was a silent skip → broken telemetry).

**Per-user OTEL attribution**
- Wire `otelHeadersHelper` for IDC-with-binary so the ARN-derived email reaches the collector; `otel-helper --test` now reads the cache so it reflects IDC attribution.

## Tests
- **Go:** token validity/refreshability gates, refresh-lock serialization, headless fail-fast (SSH and non-SSH), `isHeadless` matrix, SSO cache round-trip, otel-helper test-mode cache read.
- **Python:** IDC credential-hook wiring, IMDS guard, `otelHeadersHelper` → `.cmd` with backslashes, missing-helper-is-fatal, launcher generation + generated-bash validity.

All Go and new Python tests pass; rebased on latest `beta` (reconciled a duplicate `isHeadless` with #658 by consolidating into one shared, superset implementation).

## Notes / follow-ups
- A separate PR covers the distribution fix for including all Windows-required files (kept out of this PR per one-concern-per-PR).
- Windows behaviors were verified by inspecting generated `install.bat`/settings and cross-compiling; final on-box confirmation is recommended (multi-layer batch/PowerShell/JSON escaping).
- Tracked TODO: add an `ALBIdleTimeoutSeconds` parameter (default 300) to the OTEL collector ALB to reduce benign "socket hang up" export noise from idle-connection reaping.

### DIFF
--- a/assets/docs/LOCAL_TESTING.md
+++ b/assets/docs/LOCAL_TESTING.md
@@ -192,3 +192,88 @@ unset AWS_SESSION_TOKEN
 export AWS_PROFILE=ClaudeCode
 aws sts get-caller-identity
 ```
+
+### IAM Identity Center (IDC) on headless / SSH hosts
+
+IDC profiles (`auth_type: idc`) sign in with the browser-based device-authorization
+flow. The credential process runs this automatically the first time it needs
+credentials, printing a verification URL and code. On a desktop it also opens
+your browser; on a headless box or over SSH it detects the lack of a local
+browser (via `SSH_CONNECTION`/`SSH_TTY` or no `DISPLAY`) and instead shows a URL
+you open on any other device.
+
+**Use the `claude-bedrock` launcher instead of `claude` directly.** The installer
+generates a launcher next to the binaries
+(`~/claude-code-with-bedrock/claude-bedrock`, or `claude-bedrock.cmd` on Windows).
+It signs you in (showing the verification URL live in your terminal — a no-op if
+you already have a valid session), then runs Claude Code:
+
+```bash
+~/claude-code-with-bedrock/claude-bedrock
+# First run: a verification URL + code print to your terminal.
+# Open the URL on any device with a browser, approve — then Claude Code starts.
+
+# Optional: add the folder to PATH so you can just type `claude-bedrock`:
+export PATH="$HOME/claude-code-with-bedrock:$PATH"
+
+# Use a non-default profile:
+AWS_PROFILE=ClaudeCode ~/claude-code-with-bedrock/claude-bedrock
+```
+
+You can also sign in by hand without launching Claude Code (e.g. to pre-warm the
+cache):
+
+```bash
+credential-process --login --profile ClaudeCode
+export AWS_PROFILE=ClaudeCode
+aws sts get-caller-identity
+```
+
+`--login` performs only the sign-in (it never prints credentials) and is a no-op
+when a valid session is already cached. No AWS CLI is required — the
+device-authorization login is built into the credential process binary, and the
+SSO token is cached in `~/.aws/sso/cache/` (refreshed automatically until the
+session fully expires).
+
+> **Why a launcher instead of plain `claude`?** The IDC sign-in is interactive
+> (browser approval), but Claude Code runs the credential helper
+> non-interactively and cannot display its prompt reliably — so it can't be
+> driven from inside a running session. If you start `claude` without an active
+> sign-in, the credential process fails fast and Claude Code flashes a brief
+> "Cloud authentication" error before retrying — easy to miss, leaving `claude`
+> apparently stuck. The `claude-bedrock` launcher avoids this by running the
+> sign-in first, in your terminal, where the URL is shown and persists.
+>
+> If you need to see a credential error that already scrolled away, run Claude
+> Code with debug logging: `CLAUDE_CODE_DEBUG_LOGS_DIR=~/.claude/debug claude
+> --debug` captures credential-resolution output to a file.
+
+### IDC credential refresh during a session
+
+IDC role credentials are short-lived — their lifetime is set by the IAM Identity
+Center **permission set's session duration** (AWS default: 1 hour, max 12). For
+IDC the generated `settings.json` wires **two** Claude Code credential hooks,
+which work together:
+
+- **`awsCredentialExport`** — the primary resolver. Its output is captured
+  silently, and Claude Code re-invokes it automatically ~5 minutes before the
+  credentials' `Expiration`. This drives the *silent* hourly refresh: while the
+  longer-lived SSO session is still valid, the credential process re-mints role
+  credentials via STS with no browser. (Requires Claude Code **v2.1.176+**;
+  the credential-process JSON schema is accepted in **v2.1.181+**.)
+- **`awsAuthRefresh`** (`--login`) — fires on the first credential failure and
+  its output **is displayed to the user**. This is the only channel that can
+  surface the sign-in message, because `awsCredentialExport` discards stderr.
+  When there's no valid SSO session, this is what tells the user to relaunch
+  with `claude-bedrock` instead of leaving `claude` silently retrying.
+
+Without `awsCredentialExport`, credentials are resolved only once at startup, so
+after the session duration elapses Claude Code retries expired credentials
+(`API error · Retrying…`) and, on EC2, the SDK chain falls back to the instance
+role. To prevent that silent wrong-identity fallback, IDC settings also set
+`AWS_EC2_METADATA_DISABLED=true`, so a refresh failure surfaces as a clear
+credentials error instead.
+
+The separate ~8-hour SSO **session** expiry still requires an interactive
+re-login — relaunch with `claude-bedrock`, which runs `--login` (URL shown live)
+before starting Claude Code.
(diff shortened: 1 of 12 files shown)


wirjo:
@scouturier Tagging you for awareness — this PR has significant Windows-specific changes (otelHeadersHelper .cmd paths, escaped backslashes, headless detection for non-SSH Windows). Your Windows AV/deployment expertise would be valuable here.

wirjo:
## Backward Compatibility Assessment

### Scope: IDC-only vs Cross-cutting

**IDC-only (cannot affect OIDC users):**
- All new idc.go functions (`runIDCLogin`, `ensureIDCToken`, `refreshIDCTokenLocked`, `doIDCRefresh`, `runDeviceAuthorization`)
- Portlock serialization (port 8401) — IDC refresh only
- `AWS_EC2_METADATA_DISABLED=true` — gated by `effective_auth_type == "idc"`
- `awsCredentialExport` hook — IDC only
- `--login` flag — rejects non-IDC profiles

**Cross-cutting (all auth types):**
- `isHeadless()` consolidation in `quota_notification.go`
- `otelHeadersHelper` wiring logic
- Windows otel backslash path fix
- Installer messaging (launcher reference)
- otel-helper.cmd missing = fatal when monitoring enabled

### Changes to OIDC `run()`: NONE

### isHeadless() vs #658:
- Adds `$BROWSER` override (less restrictive — more users get notifications)
- Adds `SSH_CLIENT` check (more restrictive for SSH edge cases)
- Net: correct refinements, not regressions

### Risk to existing OIDC/CoWork: LOW
- OIDC credential flow untouched
- Windows backslash fix is a bug fix (existing fwd-slash was broken in cmd.exe)
- Installer messaging mentions launcher for all auth types, but launcher `--login` exits 1 for OIDC (UX issue, not regression)

### Concerns:
1. **Port 8401 sharing** between IDC portlock and quota notification HTTP server (graceful fallback, low probability)
2. **Launcher generation unconditional** but fails for OIDC users who try it (new feature, not regression)

### Verdict: ✅ Safe to merge (90% confidence)

---

## Test Coverage Review

### What IS tested (11 Go tests + 10 Python tests):
| Function | Tested? | Coverage |
|----------|---------|----------|
| `tokenValid()` | ✅ | Valid/expired/missing/malformed |
| `tokenRefreshable()` | ✅ | Expired+refreshable, not-refreshable |
| `writeSSOTokenCache()` | ✅ | SDK format, file permissions |
| `ensureIDCToken()` (refreshable path) | ✅ | Skips device-auth when refreshable |
| `refreshIDCTokenLocked()` (concurrent) | ✅ | Portlock wait behavior |
| `runIDCLogin()` | ✅ | Valid token, refreshable via concurrent winner |
| `runDeviceAuthorization()` (headless) | ✅ | Fail-fast in SSH, fail-fast without TTY |
| `isHeadless()` | ✅ | Matrix of SSH/DISPLAY/BROWSER combos |
| `otelHeadersHelper` wiring | ✅ | Backslash paths, missing=fatal, IDC gating |
| Launcher generation | ✅ | Valid bash, login-before-claude, balanced quotes |
| IDC credential hooks | ✅ | auth_refresh+login, credential_export, IMDS disable |

### What is NOT tested:
| Function | Gap | Risk |
|----------|-----|------|
| `resolveIDCSettings()` | No test | Config resolution — could fail on edge-case profiles |
| `runIDC()` (full flow) | No integration test | Only unit tests of sub-functions |
| `doIDCRefresh()` | No test | Actual SSO OIDC client refresh (needs mock) |
| `pollForToken()` | No test | Device-auth polling loop (timing, retry logic) |
| `stderrIsTerminal()` | No test | Platform-specific TTY detection |
| `launcherPath()` | No test | Path resolution |

### Suggestions to Simplify

1. **Extract `isHeadless()` to a shared package.** It is now used in both `quota_notification.go` and `idc.go` (via `stderrIsTerminal` + headless checks). A single `internal/env` or `internal/tty` package with `IsHeadless()` + `StderrIsTerminal()` would prevent drift and dedup.

2. **Consolidate portlock usage.** Port 8401 is used for both IDC refresh serialization AND the quota notification HTTP server (#658). Consider offsetting one (e.g., IDC refresh → 8402, or use a file lock instead of port for serialization). The port collision is currently graceful but accidental.

3. **`doIDCRefresh()` + `pollForToken()` could benefit from table-driven tests** with a mock SSOOIDC client. The AWS SDK v2 has interface-based mocking patterns — injecting a `ssooidcClient` interface would allow testing the refresh logic without network calls.

4. **`runIDC()` end-to-end test:** The function orchestrates resolveIDCSettings → ensureIDCToken → getRoleCredentials → outputJSON. A single integration test with mocked AWS clients would cover the full flow and catch wiring errors that unit tests miss.

5. **The installer bash generation** is tested for validity (`bash -n`) but not for idempotency. Running the installer twice should be safe — consider a test that generates, "installs" to a temp dir, then re-generates to verify no corruption.

wirjo:
## Merge Ordering Note: #671 vs #660

These two PRs have a hard conflict on `idc.go` — #671 rewrites `runIDC()` entirely, while #660 adds an `ensureProxyRunning(a.profile)` call at the end of the current `runIDC()`.

**Recommended order: merge #671 first, then rebase #660.**

- If #671 merges first → #660 rebase is trivial (add 1 line: `ensureProxyRunning(a.profile)` after `getRoleCredentials` in the new structure)
- If #660 merges first → #671 rebase is painful (full `idc.go` rewrite conflicts with proxy call placement)

Other shared files (`main.go`, `package.py`, `otel-helper/main.go`) touch different sections and should auto-resolve on rebase.

cc @wirjo — I will rebase #660 after this merges.

wirjo:
## Follow-up PRs (in progress)

Building two follow-up PRs to reduce regression risk and simplify:

**PR A: `refactor: IDC reliability hardening`**
- Offset IDC portlock to port 8402 (avoids collision with quota notification HTTP server on 8401)
- Gate launcher generation on IDC auth type only (OIDC users don't need it and it fails for them)
- Fix Ruff lint error in test_package_idc_otel_helper.py
- Skip bash-validation tests on Windows (heredocs with Windows paths break bash -n)

**PR B: `test: IDC integration and coverage gaps`**
- Integration test for full `runIDC()` flow with mocked AWS clients
- Unit tests for `pollForToken()` (retry, backoff, timeout, access_denied)
- Unit tests for `doIDCRefresh()` (success, expired refresh_token, network error)

Both will target beta and be low-risk (PR A is minimal refactoring, PR B is test-only). Will link here when ready.

PR #673, opened 2026-06-25
refactor: IDC reliability hardening (portlock offset, launcher gating, lint fixes)

wirjo:
## Summary

Follow-up to #671. Addresses reliability issues and CI failures introduced by the IDC launcher and portlock changes.

## Changes

1. **Offset IDC portlock to port 8402** — avoids collision with `quota_notification.go`'s HTTP server which tries port 8401 first. The IDC refresh lock now uses 8402, distinct from both the OIDC redirect port (8400) and the quota notification server (8401).

2. **Gate launcher generation on IDC auth only** — the `claude-bedrock` launcher is only meaningful for IDC (where interactive sign-in must be shown before Claude starts). OIDC users get a simplified 'use claude directly' message instead. This prevents confusion when `--login` exits 1 for non-IDC profiles.

3. **Fix Ruff C408 lint error** — `test_package_idc_otel_helper.py` had `dict()` call that Ruff wants as a dict literal.

4. **Fix Windows bash test failures** — `test_installer_is_valid_bash` and `test_generated_launcher_is_valid_bash` are skipped on Windows (`sys.platform == 'win32'`) since bash heredoc validation isn't meaningful there.

## Dependencies

Builds on #671 which is already merged to beta. No other dependencies.

## Testing

- `ruff check source/tests/` passes
- `go build ./...` succeeds
- `go test ./cmd/credential-process/...` passes
- Python AST parse of package.py succeeds

### DIFF
--- a/source/tests/cli/commands/test_installer_launcher.py
+++ b/source/tests/cli/commands/test_installer_launcher.py
@@ -63,7 +63,7 @@ class TestBashInstallerLauncher:
         assert "exec claude" in content
         assert "|| exit 1" in content
 
-    @pytest.mark.skipif(sys.platform == 'win32', reason='bash heredoc validation not meaningful on Windows')
+    @pytest.mark.skipif(sys.platform == "win32", reason="bash heredoc validation not meaningful on Windows")
     def test_installer_is_valid_bash(self):
         if shutil.which("bash") is None:
             pytest.skip("bash not available")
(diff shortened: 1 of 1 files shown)


wirjo:
Thanks @qing-at-work for the solid IDC implementation in #671! This PR follows up with a few hardening tweaks — would appreciate your review since you know the IDC flow best:

1. Portlock offset (8401→8402) to avoid collision with quota notification server
2. Launcher gated to IDC-only (OIDC users don't need it)
3. Ruff lint fix + Windows test skip for the bash validation tests

All minimal changes, no new features.

wirjo:
Merged — thanks again @qing-at-work for the solid IDC work in #671. These hardening improvements build directly on your implementation. The portlock offset and launcher gating were your first PR's only minor gaps, now closed.

PR #675, opened 2026-06-25
fix(distribution): use CNAME for custom-domain landing page (OIDC callback mismatch)

qing-at-work:
## Why

When the landing-page distribution is deployed with a **custom domain backed by Route53**, it created an **A-alias** record pointing at the ALB. An A alias resolves the hostname to the ALB's IP addresses, so OIDC callbacks ended up carrying the resolved IP instead of the configured custom domain. The IdP then rejected sign-in with a **"Callback URL mismatch"** error, because the callback didn't match the custom-domain URL registered with the provider.

## What

Change the Route53 record from an **A alias** to a **CNAME** pointing at the ALB's DNS name, so the custom domain name (which matches the registered callback URL) stays in the request:

```diff
-      Type: A
-      AliasTarget:
-        HostedZoneId: !GetAtt DistributionALB.CanonicalHostedZoneID
-        DNSName: !GetAtt DistributionALB.DNSName
-        EvaluateTargetHealth: false
+      Type: CNAME
+      TTL: '300'
+      ResourceRecords:
+        - !GetAtt DistributionALB.DNSName
```

Also renamed the logical ID `DomainARecord` → `CustomDomainRecord` so the resource name no longer misrepresents its type. Still conditional on `HasRoute53Zone`.

## ⚠️ Upgrade impact (existing custom-domain deployments)

Renaming the logical ID makes CloudFormation **create the new CNAME before deleting the old A record**. Route53 rejects a CNAME that coexists with a same-name A record (`conflicting RRSet`), so an in-place stack update would fail and roll back.

**Before upgrading an existing custom-domain deployment**, do one of:
- Delete the existing Route53 **A record** for the custom domain, then run the update; or
- Delete and recreate the landing-page stack.

**Fresh deployments are unaffected.** (Per the `AWS::Route53::RecordSet` spec, keeping the logical ID would have allowed an in-place A→CNAME UPSERT — "No interruption" — but it would leave the resource misleadingly named `DomainARecord` while holding a CNAME. We chose the accurate name + documented migration over the wrong name.)

This caveat is also documented in a comment on the resource in the template.

## Tests

New `TestLandingPageCustomDomainRecord` (parses the template with a CFN-aware YAML loader; asserts by resource Type, not logical-ID name):
- the record `Type` is `CNAME`, not `A`
- it uses `ResourceRecords` + `TTL`, not `AliasTarget`
- the target is the ALB DNS name
- it remains conditional on `HasRoute53Zone`

`cfn-lint`: no new findings — warning count identical to `beta`, none touch the changed `RecordSet`.

## Compatibility
Subdomain custom domains only (e.g. `claude.example.com`); a CNAME at the zone apex is not permitted by DNS. This matches the documented setup.

## Related
Found alongside the Windows distribution fix (#672) and the IDC reliability fixes (#671); kept as a separate PR per one-concern-per-PR.

### DIFF
--- a/deployment/infrastructure/landing-page-distribution.yaml
+++ b/deployment/infrastructure/landing-page-distribution.yaml
@@ -1009,18 +1009,29 @@ Resources:
           Order: 2
           TargetGroupArn: !Ref LambdaTargetGroup
 
-  # Route53 A Record (if custom domain with Route53)
-  DomainARecord:
+  # Route53 record for the custom domain (if custom domain with Route53).
+  # Use a CNAME (not an A alias) so the host resolves through the custom
+  # domain name rather than the ALB's IP addresses. With an A alias, OIDC
+  # callbacks can be sent to/from the resolved IP instead of the configured
+  # custom domain, producing a "Callback URL mismatch" error at the IdP.
+  #
+  # UPGRADE NOTE: earlier versions created an A-alias record named DomainARecord.
+  # Renaming the logical ID here makes CloudFormation create this CNAME before
+  # deleting the old A record, and Route53 rejects a CNAME that coexists with a
+  # same-name A record ("conflicting RRSet"). Existing custom-domain deployments
+  # must therefore delete the old Route53 A record for CustomDomainName (or
+  # delete/recreate the landing-page stack) before applying this update. Fresh
+  # deployments are unaffected. See the PR / docs for the migration steps.
+  CustomDomainRecord:
     Type: AWS::Route53::RecordSet
     Condition: HasRoute53Zone
     Properties:
       HostedZoneId: !Ref HostedZoneId
       Name: !Ref CustomDomainName
-      Type: A
-      AliasTarget:
-        HostedZoneId: !GetAtt DistributionALB.CanonicalHostedZoneID
-        DNSName: !GetAtt DistributionALB.DNSName
-        EvaluateTargetHealth: false
+      Type: CNAME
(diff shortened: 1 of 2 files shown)


wirjo:
## Review

**Risk: Low (with documented migration caveat)**

**The Bug:** A-alias records resolve to ALB IP addresses. OIDC callbacks then carry the resolved IP instead of the custom domain → IdP rejects with "Callback URL mismatch".

**The Fix:** A-alias → CNAME preserves the domain name in requests, so callbacks match registered URLs.

**Impact Assessment:**

| Scenario | Impact |
|----------|--------|
| Fresh deployments | ✅ Zero impact |
| Existing without custom domain | ✅ Zero impact (conditional on `HasRoute53Zone`) |
| Existing WITH custom domain | ⚠️ Requires migration: delete old A record before stack update |
| Zone apex domains | ❌ CNAME not valid (matches documented setup — subdomains only) |

**Checklist:**
- ✅ Correct fix — CNAME is the right record type for hostname preservation
- ✅ Migration caveat documented in template comments and PR body
- ✅ Tests added — validates CNAME type, ResourceRecords, TTL, conditional
- ✅ CFN lint clean
- ✅ Logical ID rename is honest (`DomainARecord` → `CustomDomainRecord`)
- ✅ Backwards compatible for the majority (fresh deploys + no-custom-domain)

**Verdict:** ✅ Ship it. Well-scoped fix with thorough documentation of the edge case.

PR #677, opened 2026-06-25
fix(go): keep quota notification server alive until browser connects

adiospeds:
## Why

Users intermittently saw `ERR_CONNECTION_REFUSED` at `localhost:8401/quota-status` instead of the quota warning page when opening Claude Code. The quota data was correct — the local page server was being killed before the browser could connect.

The root cause is a **goroutine-vs-`os.Exit` race** in `quota_notification.go`. The server was started in a fire-and-forget goroutine with only a `time.Sleep(100ms)` before `showQuotaBrowserNotification` returned. The process then exits via `os.Exit`, which terminates every goroutine immediately — including the HTTP server — without honoring the intended 5s `select` wait inside the goroutine. Whether the browser connected in time was a pure race against how fast the OS launched Chrome, which is why it failed randomly and worked on warm runs.

This also diverged from the Python credential-provider, which is synchronous (blocks until the page is served via `server.timeout = 5`), violating `credential-helper-parity.md`. Introduced by PR #658.

## What

**Output-first, wait-at-exit** — credentials hit stdout with zero added latency; only process teardown waits for the browser:

- **Bind the listener synchronously** before opening the browser. The socket queues the browser's connection into the kernel backlog the instant `net.Listen` returns — this closes the connection race.
- `showQuotaBrowserNotification` registers the started server in a pending list and **returns immediately** (non-blocking). Credentials are already written to stdout by `outputJSON()` inside `run()`/`runIDC()`/`runPassthrough()` before those return.
- New `waitForQuotaNotification()` blocks on served-or-timeout (5s), then shuts the server down.
- New `exitAfterNotifications(code)` wrapper calls `waitForQuotaNotification()` then `os.Exit(code)` — used at all three auth entrypoints.
- `sync.Once` guards the `served` channel so a browser reload can't double-close it and panic.
- Test seams `openBrowserFunc` / `isHeadlessFunc` for testability without launching a real browser.

## Tests

New `quota_notification_test.go`:

- `TestNotificationServerReachableAfterReturn` — **the regression**: server is still reachable _after_ `showQuotaBrowserNotification` returns (this test fails on the old fire-and-forget implementation).
- `TestWaitTimesOutWhenBrowserNeverConnects` — exit path falls back to timeout and doesn't hang.
- `TestReloadDoesNotPanic` — browser reload doesn't double-close the channel.
- `TestOptOutSkipsServer`, `TestNilUsageSkipsServer` — skip conditions start no server.
- `TestWaitWithNoPendingIsNoop` — common case (no notification shown) is safe.
- `TestMain` installs a no-op browser opener binary-wide so existing `printQuotaWarning` tests no longer launch a real browser on developer machines.

Verified: `go build ./...`, `go vet`, `go test -race ./...` all clean.

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -102,11 +102,11 @@ func main() {
 
 	// Auth dispatch: IDC > legacy passthrough > OIDC
 	if cfg.IsIDC() {
-		os.Exit(app.runIDC())
+		exitAfterNotifications(app.runIDC())
 	}
 	if !cfg.IsSsoEnabled() {
 		// Legacy passthrough: no IDC fields, just ambient credential chain.
-		os.Exit(app.runPassthrough())
+		exitAfterNotifications(app.runPassthrough())
 	}
 
 	// OIDC path — resolve provider type and redirect port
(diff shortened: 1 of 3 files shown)


adiospeds:
retesting this will raise PR again.

wirjo:
Hey @adiospeds — noticed you closed this just now. Was this intentional (superseded by another PR?) or accidental? The fix looks solid and we'd be happy to review/merge if you want to reopen it. Let us know!

adiospeds:
Yes this can be merged, was performing quick tests and everything looks good... please review and merge this. reopening it ..

PR #338, opened 2026-05-04
feat: rewrite credential-process and otel-helper in Go (drop-in Python replacement)

pipeflo:
## Summary

Replaces the PyInstaller/Nuitka-based Python credential_provider and otel_helper with Go implementations. Python versions stay in the tree alongside; `ccwb package` still defaults to them. Two new flags opt into Go:

- `ccwb package --go` — cross-compiles Go binaries locally (requires Go 1.22+)
- `ccwb package --prebuilt` — copies pre-committed binaries from `source/go/prebuilt/v2.0.0/` (no build tools needed, recommended for most admins)

End users see zero protocol-level change: the Go `credential-process` produces `credential_process`-format JSON on stdout that AWS CLI parses identically to the Python version (verified byte-semantic-equal; whitespace differs — compact vs space-separated — both valid per spec).

## Motivation

1. **Windows AV false positives** — Nuitka-compiled Python binaries trigger Defender's Wacatac.B!ml heuristic in a sizable fraction of enterprise rollouts. The Go binaries ship with PE version info (go-winres .syso files) and are compiled with `-trimpath` to scrub builder paths; field testing on multiple Windows tenants shows clean execution.
2. **No Python 3 runtime dependency on end-user machines** — simplifies distribution, especially for non-developer audiences (data scientists, analysts).
3. **Cross-compilation from a single host** — replaces the PyInstaller-Docker-on-Linux + CodeBuild-for-Windows split with one `go build` matrix. Admins on any OS can produce all 5 binaries.

## What's in / out

**In:**
- `source/go/` full module: cmd/credential-process, cmd/otel-helper, internal/{browser,config,federation,jwt,oidc,otel,portlock,provider,quota,storage,version}
- `source/go/Makefile` + `source/go/prebuilt/v2.0.0/` (pre-built binaries for all 5 platforms + generic install scripts)
- `source/go/scripts/credential-process.ps1` (PowerShell alternative for zero-AV environments)
- `source/claude_code_with_bedrock/cli/commands/package.py` — two new option flags `--go` and `--prebuilt`, plus inline install.bat fix (`>/dev/null` → `>nul` — pre-existing upstream-compatible syntax)
- `assets/docs/WINDOWS_BUILD_SYSTEM.md` + `CLAUDE.md` — updated architecture notes
- Unit tests:
  - `source/go/internal/config/migration_test.go` (7 historical fixture configs)
  - `source/go/internal/storage/keyring_split_test.go` (8 Windows keyring scenarios)
[...]

PR #678, opened 2026-06-25
fix(package): restore otelcol build and sidecar config generation

scouturier:
## Why

PR #338 (Go rewrite, merged May 28) silently deleted `_build_otelcol()` and its call site from `package.py` as collateral damage. Since then, `ccwb package` for sidecar profiles has shipped packages missing three things:

- `otelcol` binary (the local collector that receives and forwards metrics)
- `collector-config.yaml` (tells otelcol where to send metrics)
- `<profile>-collector` AWS profile (SigV4 auth for CloudWatch OTLP)

Since sidecar is the **default and recommended mode** in `ccwb init`, this regression has been silently breaking telemetry for the majority of new deployments since May 28.

## What

**Restored `_build_otelcol()` and its call site** — the primary regression fix.

**Fixed three additional issues found during review:**

1. **Windows OCB binary not selected** — `_build_otelcol` fell through to the Linux OCB binary on Windows admin machines (`"darwin" if ... else "linux"`), downloading a non-executable binary and silently failing the sidecar build on Windows.

2. **IDC + quota + sidecar gap** — when `auth_type=idc` AND `quota_api_endpoint` is set, `is_idc_zero_binary=False`. Neither the zero-binary IDC block nor the OIDC block fired, so no `collector-config.yaml` was generated. Unified both paths into a new `_generate_collector_config()` method.

3. **Inline test logic replicas** — existing tests re-implemented the collector config substitution logic inline instead of calling `_generate_collector_config()`. Those tests would pass even if the production method was broken or deleted.

## Changes

- `package.py` — restored `_build_otelcol()` call site, extracted `_generate_collector_config()`, added `_GO_PLATFORM_MAP` and `_go_ldflags()` as shared module-level constants (used by both `_build_go_binaries` and `_build_otelcol` to prevent platform drift), fixed Windows OCB path detection
- `test_package_oidc_sidecar.py` — 23 new regression tests covering: sidecar endpoint override, collector config generation (OIDC + IDC + IDC+quota), installer otelcol block, shared platform map, Windows OCB branch, structural `inspect.getsource()` guards that catch future call site deletion

## Regression tests

The structural tests use `inspect.getsource()` to assert the call site exists in `handle()` — if `_build_otelcol` is deleted again, the test fails immediately:

```python
def test_build_otelcol_called_for_sidecar_profile(self):
    src = inspect.getsource(PackageCommand.handle)
    assert "_build_otelcol" in src  # fails if deleted like PR #338
    assert "sidecar" in src          # fails if guard is removed
```

## Testing

- 23/23 tests passing locally (macOS arm64)
- Verified central monitoring deployment end-to-end (metrics flowing to CloudWatch `ClaudeCode` namespace)
- Note: Windows E2E otelcol verification blocked by corp proxy preventing OCB build — the Windows OCB path fix is covered by structural test

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -3229,11 +3229,15 @@ if exist "otel-helper.cmd" (
         exit /b 1
     )
 ) else (
-    echo {'ERROR' if _otel_missing_is_fatal else 'INFO'}: otel-helper.cmd not found in package.
-{'''    echo        Claude Code needs it to send telemetry [otelHeadersHelper].
+    echo {"ERROR" if _otel_missing_is_fatal else "INFO"}: otel-helper.cmd not found in package.
+{
+            '''    echo        Claude Code needs it to send telemetry [otelHeadersHelper].
     echo        Re-extract the full package [including .cmd and .ps1 files] and retry.
     pause
-    exit /b 1''' if _otel_missing_is_fatal else '    REM Monitoring disabled - helper not required.'}
+    exit /b 1'''
+            if _otel_missing_is_fatal
+            else "    REM Monitoring disabled - helper not required."
+        }
 )
 if exist "otel-helper.ps1" (
     copy /Y "otel-helper.ps1" "%USERPROFILE%\\claude-code-with-bedrock\\otel-helper.ps1" >nul
@@ -3243,11 +3247,15 @@ if exist "otel-helper.ps1" (
         exit /b 1
     )
 ) else (
-    echo {'ERROR' if _otel_missing_is_fatal else 'INFO'}: otel-helper.ps1 not found in package.
-{'''    echo        It is the antivirus fallback for otel-helper.cmd and is required.
+    echo {"ERROR" if _otel_missing_is_fatal else "INFO"}: otel-helper.ps1 not found in package.
+{
+            '''    echo        It is the antivirus fallback for otel-helper.cmd and is required.
     echo        Re-extract the full package and run install.bat again.
     pause
-    exit /b 1''' if _otel_missing_is_fatal else '    REM Monitoring disabled - fallback not required.'}
+    exit /b 1'''
+            if _otel_missing_is_fatal
+            else "    REM Monitoring disabled - fallback not required."
+        }
 )
 
 REM Install OTEL Collector sidecar (sidecar-mode packages only). otelcol is built
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -30,6 +30,31 @@ _CREDENTIAL_PROVIDER_RUNTIME_DEPS = ["boto3", "requests", "PyJWT", "keyring", "c
 _OTEL_HELPER_RUNTIME_DEPS: list[str] = []  # otel_helper uses only stdlib
 _PYINSTALLER_PIN = "pyinstaller==6.*"
 
+# Single source of truth for Go cross-compilation targets, shared by the auth-binary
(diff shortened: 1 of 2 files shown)


wirjo:
## Review

**Risk: Low** | **Priority: High** (fixes critical silent regression since May 28)

**The Bug:** PR #338 accidentally deleted `_build_otelcol()`. Since then, all sidecar packages (the default mode) ship without the otelcol binary or collector config — silently breaking telemetry for new deployments.

**Fixes:**
1. ✅ Restores `_build_otelcol()` call site — primary regression
2. ✅ Windows OCB binary selection (was defaulting to Linux binary)
3. ✅ IDC + quota + sidecar config gap — unified into `_generate_collector_config()`
4. ✅ Structural regression tests via `inspect.getsource()` — catches future deletions

**Quality:**
- 23 new tests covering OIDC/IDC/IDC+quota sidecar paths + Windows OCB
- Shared `_GO_PLATFORM_MAP` and `_go_ldflags()` prevent platform drift
- Graceful degradation — failed collector build warns, does not fail packaging

**Verdict:** ✅ Ship it. Critical fix that should have been caught in the Go rewrite PR. Well-tested, clean implementation.

wirjo:
## Why This Wasn't Caught Earlier

**Root cause:** PR #338 (Go rewrite) was a large PR (+740 lines) that refactored the credential-process into Go. The deletion of `_build_otelcol()` was collateral damage — it wasn't an intentional removal, it was lost in the noise of a big diff.

**Why tests didn't catch it:**
1. No existing test asserted that `_build_otelcol` was *called* from `handle()` — tests only verified the method itself worked, not that it was wired up
2. No integration test actually ran `ccwb package` end-to-end and checked the output zip contained `otelcol-*` binaries
3. The failure was **silent** — packaging succeeded (exit 0), telemetry just didn't work at runtime

**Why CI didn't catch it:**
- The test suite covers unit behaviour (does this function produce correct output?) but not wiring (is this function called at the right time?)
- No smoke test validates "given a sidecar profile, the output package contains these N files"

## Suggested Improvements

1. **Structural/wiring tests** (done in this PR) — `inspect.getsource()` asserts that critical call sites exist. If `_build_otelcol` is deleted again, the test fails immediately.

2. **Package manifest test** — a test that runs the package logic in dry-run mode and asserts the output directory contains expected files:
   ```python
   def test_sidecar_package_contains_otelcol():
       # Given a sidecar profile, package outputs must include:
       assert (output / "otelcol-darwin-arm64").exists()
       assert (output / "collector-config.yaml").exists()
       assert (output / f"{profile_name}-collector").exists()  # AWS profile
   ```

3. **PR review rule** — `.claude/rules/package-completeness.md`:
   > When modifying package.py, verify that ALL platform builds (auth binary, otel-helper, otelcol) remain wired into `handle()`. Deletions in one code path (e.g., Go migration) must not remove call sites for orthogonal paths (e.g., collector sidecar).

4. **Large PR decomposition** — PR #338 was +740/-0 targeting a Go rewrite. It should have been split: (a) add Go binary path, (b) remove Python PyInstaller path. Step (b) would have been reviewed in isolation and the `_build_otelcol` deletion caught.

5. **CI package smoke test** — add a GitHub Actions job that runs `ccwb package --dry-run` with a sidecar profile fixture and asserts the output manifest. Fast (<30s), catches wiring regressions.

wirjo:
Follow-up PR implementing the robustness improvements from this review: #680

Adds:
- Structural wiring tests for `_generate_collector_config` and `_build_go_binaries` call sites
- Package manifest validation tests (17 tests covering sidecar/central/IDC output completeness)
- `.claude/rules/package-completeness.md` to prevent future wiring regressions in Claude Code sessions

PR #685, opened 2026-06-26
fix(cli): honor non-interactive mode in package and distribute + fix generic linux binary naming

wirjo:
## Summary
Fixes Bugs 2, 3, and 4 from #682 — all related to non-interactive/CI-driven deployment flows.

## Bug 2 & 3 — `package` and `distribute` ignore non-interactive mode

**Problem:** Both commands call `questionary` prompts unconditionally. When stdin is not a TTY (CI pipelines, agent-driven automation, piped commands), they crash with:
- `EOFError` (from prompt_toolkit under a PTY)
- `[Errno 22] Invalid argument` (non-TTY fd)

**Fix:** Added `_is_interactive()` helper (`sys.stdin.isatty()`) and guarded all questionary calls:
- Platform selection: defaults to all available platforms
- Co-authorship: defaults to `False` (opt-in)
- OTel customization: defaults to `False` (skip)
- IDC user email: uses auto-detected value or empty string
- Windows warning: proceeds without Windows (prints info message)
- Build selection wizard: auto-selects latest build

**Impact:** Unblocks all CI/scripted/agent-driven `ccwb package` and `ccwb distribute` flows.

## Bug 4 — Generic `--target-platform linux` silently drops binaries

**Problem:** `ccwb package --target-platform linux --go` produces `credential-process-linux` (no arch suffix), but `distribute` only recognizes `credential-process-linux-x64` / `credential-process-linux-arm64`, so:
1. `distribute` silently ships a config-only zip (no warning)
2. `install.sh` fails on target machines

**Fix (two-layer defense):**
1. **Normalize early:** In Go mode, `linux` → `linux-x64` and `macos` → `macos-arm64` before the build runs. Binary gets named correctly from the start.
2. **Accept generic names:** Added fallback entries in distribute's `platform_files` map that accept `credential-process-linux` and rename it to `credential-process-linux-x64` in the zip. This handles existing builds that used the old naming.

## Risk Assessment
- **Risk:** Low
- **Backward compat:** ✅ Existing interactive workflows unchanged (prompts still appear when TTY detected)
- **Test coverage:** All existing tests pass (31 distribute tests, 17 manifest tests)
- **Blast radius:** Only affects non-TTY execution paths (previously broken) and generic platform token resolution

## Testing
```bash
python3 -m pytest source/tests/cli/commands/test_package_manifest.py -x -q  # 17 passed
python3 -m pytest source/tests/test_distribute.py -x -q                      # 31 passed
```

Partially addresses #682
cc @scouturier for review

PR #686, opened 2026-06-26
test: add package ↔ distribute binary naming contract

wirjo:
## Why This Is the Highest-Priority Improvement

Issue #682 revealed that `--target-platform linux` silently produced a config-only distribution zip with **zero binaries and zero warnings**. The failure was invisible at build time, invisible at distribute time, and only surfaced when `install.sh` ran on the target machine — potentially days later, on a different machine, by a different person.

### Why wasn't this caught before?

The three components that must agree on binary names are tested **in isolation** but never as a **connected pipeline**:

| Component | What it does | Its view of names |
|-----------|-------------|-------------------|
| `package.py` | Builds binaries | Names via `f"-{plat}"` — accepts `linux` |
| `distribute.py` | Packs into zip | Hardcoded allowlist — only `linux-x64`, `linux-arm64` |
| `install.sh` | Installs on target | Constructs from `uname` — expects `linux-x64` |

Each passes its own tests. Together, they silently fail.

### What this PR adds

A contract test (`test_package_distribute_contract.py`) that **statically verifies naming agreement** between all three components:

1. **`test_every_go_platform_produces_recognized_binary`** — The core contract. Every platform key in `_GO_PLATFORM_MAP` must produce a binary name that distribute.py will include in its zip (either directly or via a known normalization mapping). If someone adds a new platform token, this fails immediately.

2. **`test_install_sh_suffixes_match_distribute`** — Every suffix install.sh can construct from `uname` must exist in distribute's allowlist.

3. **`test_otel_helper_naming_matches_credential_process`** — otel-helper must cover the same platform set as credential-process.

4. **`test_go_platform_map_has_no_ambiguous_generic_tokens`** — Flags generic tokens (`linux`, `macos`) that would produce non-standard names if not normalized.

5. **`test_platform_suffix_is_deterministic`** — No runtime state (uname, env) should influence binary naming in the Go build path.

### Why this test over other improvements?

- **Silent failures are worse than loud failures.** The non-interactive crash (Bugs 2/3) at least threw an error. Bug 4 succeeded everywhere except the last step.
- **Trivial to implement** — 235 lines, runs in <1s, no external dependencies.
- **Catches an entire class of future bugs** — any naming change in any of the three components breaks the contract immediately.
- **Shift-left guarantee** — fails in CI before the broken package ever ships.

### Test results
```
10 passed, 1 skipped in 0.63s
```

The skipped test (`test_go_platform_map_has_no_ambiguous_generic_tokens`) is informational — it documents that generic tokens exist and should be normalized.

Related: #682, PR #685 (the fix), PR #684 (quota route fix)
cc @scouturier for review

### DIFF
--- a/source/tests/test_package_distribute_contract.py
+++ b/source/tests/test_package_distribute_contract.py
@@ -1,0 +1,235 @@
+# ABOUTME: Contract test ensuring package.py binary naming agrees with distribute.py's allowlist.
+# ABOUTME: Catches silent failures where built binaries are dropped from distribution zips.
+
+"""Contract tests for package ↔ distribute binary naming agreement.
+
+These tests verify that every binary filename package.py can produce is recognized
+by distribute.py's platform_files allowlist. Without this contract, a naming mismatch
+causes distribute to silently ship a config-only zip (no binaries, no warning), and
+install.sh fails on the target machine.
+
+Motivation: Issue #682 Bug 4 — `--target-platform linux` produced `credential-process-linux`
+which distribute silently dropped (only recognized `credential-process-linux-x64`).
+The failure was invisible at build time and only surfaced at install time.
+
+This test ensures the three components (package naming, distribute allowlist, install.sh
+expectations) always agree on binary names.
+"""
+
+import sys
+from pathlib import Path
+
+import pytest
+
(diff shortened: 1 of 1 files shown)


wirjo:

wirjo:

PR #689, opened 2026-06-26
fix(credential-process): add --set-client-secret to Go binary

scouturier:
## Why

A user on a Go-packaged install using Azure EntraID with `azure_auth_mode=secret` was blocked: the credential-process told them to run `credential-process --set-client-secret --profile <name>`, but that flag **only existed in the Python provider**. Running it against the Go binary produced an unknown-flag error — the binary's own error message pointed at a command it didn't implement.

This is a Python→Go parity gap of the same class called out in `.claude/rules/credential-helper-parity.md` (provider detection, browser notifications). It's the second such gap found recently.

## What

- Add `storage.SaveClientSecret()` — writes the **same** keyring key `ReadClientSecret` already reads (`{profile}-client-secret` under service `claude-code-with-bedrock`), so it interoperates with secrets stored by `ccwb init`.
- Add the `--set-client-secret` flag to the Go credential-process:
  - Reads `CCWB_CLIENT_SECRET` for non-interactive/automation use, or prompts via terminal with hidden input.
  - Blank input clears the stored entry.
  - Runs **before** config load, so it works on a fresh machine where `config.json` doesn't exist yet.
- Update the "no client secret is stored" error message to point at the now-working Go command instead of the `ccwb init` hint.
- Add a regression test covering the save → read → clear round-trip.

## Testing

- `go test ./internal/storage/` and `./cmd/credential-process/` pass.
- `gofmt` / `go vet` clean.

## Notes

Scoped to one concern per `.claude/rules/pr-standards.md`. A broader Python↔Go parity audit surfaced additional gaps (Windows keyring chunking for the monitoring token, OTEL cache `schema_version` mismatch, monitoring-token expiry buffer); those will be addressed in separate follow-up PRs.

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -12,6 +12,7 @@ import (
 
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"golang.org/x/term"
 
 	"ccwb-go/internal/config"
 	"ccwb-go/internal/federation"
@@ -51,6 +52,7 @@ func main() {
 	showTags := flag.Bool("show-tags", false, "Print the https://aws.amazon.com/tags claim from the cached ID token (debug)")
 	getTag := flag.String("get-tag", "", "Print the value of a single principal tag from the cached ID token (e.g. --get-tag Zone). Exit codes: 0 hit, 2 absent, 4 expired.")
 	login := flag.Bool("login", false, "Interactively sign in (IDC: run device authorization and cache the SSO token), then exit. Use this once on headless/SSH hosts before Claude Code runs.")
+	setClientSecret := flag.Bool("set-client-secret", false, "Store Azure AD client secret in OS secure storage. Set CCWB_CLIENT_SECRET env var for non-interactive use, or enter it at the prompt.")
 	flag.Parse()
 
 	if *versionFlag || *shortVersion {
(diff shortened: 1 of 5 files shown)


wirjo:

### Impact on existing deployments
**None — fully additive.** The change adds a new `--set-client-secret` flag and a new `SaveClientSecret()` function. No existing flags, behaviours, or API surfaces are modified. The keyring key format (`{profile}-client-secret` under service `claude-code-with-bedrock`) is identical to what `ReadClientSecret` and `ccwb init` already use, so secrets stored by older versions are still read correctly, and secrets stored by this new flag are readable by the Python provider.

The only existing-code change is the error message text (line 639) — switching from "Run: ccwb init" to "Run: credential-process --set-client-secret". This is a UX improvement for Go-binary users who previously received instructions pointing at a command that didn't work.

### Regression risks
**Low.**

1. **`keyringRW` interface expansion** — Adding `Remove(key string) error` to the interface is the one structural change. Any external or internal implementors of `keyringRW` would need to add `Remove`. However, this interface is internal (`internal/storage`), the only implementations are the real `keyring.Keyring` (which already has `Remove`) and the test mock (updated here). No downstream breakage.

2. **`--set-client-secret` runs before config load** — Intentional design so it works on fresh machines. Since it calls `os.Exit()` after completion, it cannot interfere with the normal credential-process flow. Clean isolation.

3. **`golang.org/x/term` dependency** — New import for hidden terminal input. Standard, widely-used Go module with no known issues. Already transitively present in most Go projects.

4. **Import reordering** (`keyring_split.go`, `keyring.go`) — goimports-style cosmetic change, no functional impact.

### Test coverage
**Good.** A dedicated regression test (`client_secret_test.go`) covers the full round-trip: absent → save → read → clear → read-after-clear → clear-when-absent. Uses the mock keyring so it runs reliably on Linux CI without a Secret Service backend. The `mockKeyring.Remove` method is also added to support this.

The test explicitly guards against the original bug (Go binary lacking the save path) and edge cases (clearing an already-absent key swallows `ErrKeyNotFound`).

### Verdict
**Safe to merge.** Purely additive feature that closes a real Python→Go parity gap. No behavioural changes to existing code paths, clean test coverage, and the keyring interop is correct by construction (same key format).

Thank you @scouturier for the contribution! Solid fix — well-scoped, clearly motivated, and the regression test is exactly what's needed to prevent this class of parity gap from recurring.

@scouturier LGTM — can merge when ready! ✅

wirjo (approved):

wirjo:

wirjo:

PR #391, opened 2026-05-22
feat: sidecar monitoring mode fixes and Cowork telemetry guardrails

gcacace:
## Description

Hardens the sidecar monitoring mode by fixing backward compatibility for existing profiles, adding Cowork telemetry guardrails, and migrating the CoWork dashboard to PromQL. Ensures that customers upgrading without re-running `ccwb init` continue operating in central mode, and that sidecar mode clearly communicates (and enforces) that Claude Cowork desktop telemetry is not supported.

## Why is this change needed

The recently introduced "sidecar" vs "central" monitoring mode had several issues:
1. Existing profiles missing the `monitoring_mode` field would incorrectly default to "sidecar" due to the dataclass default, breaking their deployments
2. Claude Cowork (desktop app) cannot send OTLP to a localhost collector, but the MDM config was emitting `otlpEndpoint: http://localhost:4318` in sidecar mode
3. The CoWork dashboard could still be deployed in sidecar mode despite having no data source
4. Orphaned stack cleanup deleted stacks in deployment order instead of reverse, causing dependency failures

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Infrastructure/deployment change
- [ ] Dependency update
- [ ] CI/CD changes

## Changes Made

- **Config**: Default `monitoring_mode` to "central" for existing profiles (backward compat) with explicit migration in `Profile.from_dict()`
- **Init wizard**: Add Cowork telemetry limitation warning to sidecar description; add Cowork telemetry support as a PRO for central mode
- **CoWork MDM generation**: Skip OTLP config entirely in sidecar mode; central mode correctly resolves endpoint from CloudFormation outputs
- **Deploy**: Block cowork-dashboard deployment in sidecar mode; delete orphaned stacks in reverse order; add cowork-dashboard to orphaned stack detection; remove unused MetricsLogGroup parameter
- **CloudFormation**: Migrate CoWork dashboard to PromQL with service.name filtering; consolidate SSM parameter to single resource for safe stack updates
- **Changelog**: Tag as v2.4.0 with all changes documented

## Additional Notes

- Existing customers who upgrade without re-running `init` will continue operating in "central" mode with no behavior change
- New customers running `init` for the first time will be offered "sidecar" as the recommended default
- The init wizard now clearly communicates the Cowork telemetry limitation for sidecar mode

[...]

PR #690, opened 2026-06-26
fix(deploy): gate central infrastructure on monitoring_mode in sidecar mode

wantmys2000:
## Why
In sidecar monitoring mode the OTEL collector runs locally and ships metrics straight to CloudWatch, so no VPC/ECS/ALB or Athena pipeline is needed. But `ccwb deploy` scheduled the entire central stack regardless of mode, so **sidecar deployments attempted to create a VPC + ECS + ALB**. That fails outright in accounts that disallow new VPCs/subnets, and needlessly provisions costly infrastructure for everyone else running sidecar mode.

## Root cause
PR #338 (the Go rewrite) refactored deploy.py's "deploy all" stack selection and dropped the `monitoring_mode == "central"` gate that the original sidecar feature carried. PR #391 only restored the guard for the `cowork-dashboard` path, leaving `networking`/`monitoring`/`analytics` ungated — so the regression stayed live.

## What
- Extract full-deploy selection into a pure, unit-testable `_select_full_deploy_stacks()`.
- **Sidecar mode** now schedules only `auth`, `dashboard`, and (for OIDC/IDC quota) `s3bucket` + `quota` — never `networking`, `monitoring`, `cowork-dashboard`, or `analytics`. **Central mode** keeps the full stack unchanged.
- `ccwb deploy networking` and `ccwb deploy monitoring` now refuse in sidecar mode with a clear message (mirrors the existing `cowork-dashboard` guard).
- Add 9 regression tests pinning stack selection per mode: sidecar exclusions, dashboard-always, quota+s3bucket ordering, IDC sidecar quota, central full stack, existing-VPC skip, missing-`monitoring_mode` backward-compat, and sidecar + `none` auth.

## Pairs with
#678 (already on `beta`) restored the `otelcol` build and sidecar `collector-config` generation in `ccwb package` that the same #338 refactor dropped. This PR is the deploy-side half of that same regression.

## Testing
- `cd source && poetry run pytest tests/ -q` → 1308 passed, 22 skipped, 0 failures.
- `ruff check` clean; `ruff format` applied.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -122,17 +122,27 @@ class DeployCommand(Command):
                     return 1
                 stacks_to_deploy.append(("auth", "Authentication Stack (Cognito + IAM)"))
             elif stack_arg == "networking":
-                if profile.monitoring_enabled:
-                    stacks_to_deploy.append(("networking", "VPC Networking for OTEL Collector"))
-                else:
+                if not profile.monitoring_enabled:
                     console.print("[yellow]Monitoring is not enabled in your configuration.[/yellow]")
                     return 1
+                if getattr(profile, "monitoring_mode", "central") == "sidecar":
+                    console.print(
+                        "[yellow]Networking stack is not used in sidecar monitoring mode "
+                        "(the local OTEL collector needs no VPC/subnets).[/yellow]"
+                    )
+                    return 1
+                stacks_to_deploy.append(("networking", "VPC Networking for OTEL Collector"))
             elif stack_arg == "monitoring":
-                if profile.monitoring_enabled:
-                    stacks_to_deploy.append(("monitoring", "OpenTelemetry Collector"))
-                else:
+                if not profile.monitoring_enabled:
                     console.print("[yellow]Monitoring is not enabled in your configuration.[/yellow]")
                     return 1
+                if getattr(profile, "monitoring_mode", "central") == "sidecar":
+                    console.print(
+                        "[yellow]The central OTEL collector stack is not used in sidecar mode "
+                        "(telemetry is sent to CloudWatch by the local collector).[/yellow]"
+                    )
+                    return 1
+                stacks_to_deploy.append(("monitoring", "OpenTelemetry Collector"))
             elif stack_arg == "dashboard":
                 if profile.monitoring_enabled:
                     stacks_to_deploy.append(("dashboard", "CloudWatch Dashboard"))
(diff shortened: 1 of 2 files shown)


wirjo:

### 1. Could existing sidecar deployments that ALREADY have central infra deployed break on update?

**Low risk, but worth noting.** After this PR, running `ccwb deploy` on a sidecar profile will no longer *update* the networking, monitoring, cowork-dashboard, or analytics stacks — it simply won't schedule them. This means those stacks won't be deleted, but they also won't receive any future configuration updates via `ccwb deploy`. Existing stacks remain intact in CloudFormation (they are not torn down), so nothing "breaks" — but operators should be aware that those stacks become unmanaged going forward. A `ccwb destroy` or manual cleanup would be needed to remove orphans. This is acceptable behaviour for a bug fix (those stacks should never have been deployed in the first place).

### 2. Could this leave orphaned resources (VPC, ECS, ALB) that were already deployed?

**Yes, by design.** The PR correctly avoids touching stacks it shouldn't have deployed. It does not auto-delete them (which would be dangerous and destructive). Orphaned resources will remain until the operator explicitly removes them. A follow-up improvement could be a warning message during `ccwb deploy` like "Detected stacks X, Y from a previous central deployment — consider running `ccwb destroy --stack networking monitoring` to clean up." But that's enhancement territory, not a blocker.

### 3. Does the explicit stack deploy (`ccwb deploy monitoring`) still work for central mode?

**Yes.** The explicit stack path (lines 125-142) checks `getattr(profile, "monitoring_mode", "central")` — if the mode is `"central"` (or missing/defaulted), it appends the monitoring stack as before. Only sidecar profiles get the informational error message and early return. Central mode is unaffected.

### 4. Is the backwards-compat for missing `monitoring_mode` field handled?

**Yes, correctly handled.** Both the explicit stack path and `_select_full_deploy_stacks()` use `getattr(profile, "monitoring_mode", "central")` which defaults to `"central"` when the attribute is absent. The test `test_missing_monitoring_mode_defaults_to_central` explicitly validates this by deleting the attribute and asserting the full central stack is deployed. Old config files without this field will behave identically to before.

### Verdict

**Safe.** This is a well-structured fix with thorough test coverage (10 test cases covering sidecar, central, quota interactions, missing field backcompat, and auth variants). The extraction into `_select_full_deploy_stacks()` is clean and testable. The only minor gap is no runtime warning about orphaned resources from prior incorrect sidecar deploys, but that's a polish item, not a regression risk.

Thank you @wantmys2000 for the contribution! The test matrix is excellent — particularly the `test_missing_monitoring_mode_defaults_to_central` case which catches the exact backwards-compat scenario that would bite users upgrading from older configs. The docstring explaining the #338 regression context is also valuable for future maintainers.

wirjo:
@scouturier LGTM — backwards compat confirmed, test coverage is solid. Can merge when ready! ✅

wirjo (approved):

wirjo:

wirjo:

PR #691, opened 2026-06-26
fix(govcloud): use ${AWS::Partition} for all ARNs so GovCloud deploys work

wantmys2000:
## Why
GovCloud uses the `aws-us-gov` ARN partition (China uses `aws-cn`). **19 ARNs** across the CloudFormation templates hardcoded the commercial `aws` partition, so in GovCloud they either fail outright or silently never match:

- IAM roles referencing `arn:aws:iam::aws:policy/...` **cannot attach** the managed policy.
- The IDC trust-policy principals in `bedrock-auth-idc.yaml` (`...:root`, SSO `PrincipalArn`) **cannot resolve** → broken auth.
- `execute-api`, `secretsmanager`, `cognito-idp`, `elasticloadbalancing`, and `logs` resource ARNs silently never match → access denied at runtime.

The repo already uses `${AWS::Partition}` in 46 places; these were stragglers (three templates had zero partition substitution).

## What
- Replace every hardcoded `arn:aws:` with `arn:${AWS::Partition}:` across **7 templates** — 11 AWS managed-policy ARNs + 8 `!Sub` service/principal ARNs. The AWS-owned account alias in `::aws:policy/...` is intentionally left unchanged (it is not the partition).
- **`landing-page-distribution.yaml`**: add GovCloud ELB log-delivery account IDs (`us-gov-west-1`, `us-gov-east-1`). The ALB access-log bucket policy does an **unconditional `!FindInMap`** on the ELB account map, so without these keys the whole stack fails to deploy in GovCloud — despite the dual-principal design that intended to cover it.
- **`init.py`**: make the last-resort hand-built Secrets Manager ARN partition-aware via a new `utils.partition.aws_partition_for_region()` helper (Python mirror of `${AWS::Partition}`).
- **Regression guards**: a parametrized contract test that fails on any `arn:aws:` literal in any template, a GovCloud-region-presence check on the ELB account map, and unit tests for the partition helper.

## Not changed (intentionally)
- Service principals (`*.amazonaws.com`, e.g. `ecs-tasks.amazonaws.com`) — these do **not** change in GovCloud.
- The Go/Python STS credential path — it uses AWS SDK v2 `NewFromConfig`, which resolves the partition from the configured region automatically.

## Testing
- `cd source && poetry run pytest tests/ -q` → **1334 passed, 22 skipped, 0 failures**.
- `cfn-lint` on all 7 changed templates: **zero new findings** (identical warning counts before/after).
- `ruff check` + `ruff format` clean.

### DIFF
--- a/deployment/infrastructure/analytics-pipeline.yaml
+++ b/deployment/infrastructure/analytics-pipeline.yaml
@@ -166,7 +166,7 @@ Resources:
               Service: lambda.amazonaws.com
             Action: 'sts:AssumeRole'
       ManagedPolicyArns:
-        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+        - !Sub 'arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole'
 
   TransformLambda:
     Type: AWS::Lambda::Function
--- a/deployment/infrastructure/bedrock-auth-idc.yaml
+++ b/deployment/infrastructure/bedrock-auth-idc.yaml
@@ -67,7 +67,7 @@ Resources:
                 - 'logs:PutLogEvents'
                 - 'logs:DescribeLogGroups'
                 - 'logs:DescribeLogStreams'
-              Resource: !Sub 'arn:aws:logs:*:${AWS::AccountId}:log-group:/aws/claude-code/*'
+              Resource: !Sub 'arn:${AWS::Partition}:logs:*:${AWS::AccountId}:log-group:/aws/claude-code/*'
             - !Ref 'AWS::NoValue'
   
   # IAM role for IAM Identity Center federation
@@ -87,11 +87,11 @@ Resources:
           # Trust policy for SSO-assumed roles (role chaining from Permission Set)
           - Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
+              AWS: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:root'
             Action: 'sts:AssumeRole'
             Condition:
               ArnLike:
-                'aws:PrincipalArn': !Sub 'arn:aws:iam::${AWS::AccountId}:role/aws-reserved/sso.amazonaws.com/*'
+                'aws:PrincipalArn': !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:role/aws-reserved/sso.amazonaws.com/*'
       ManagedPolicyArns:
         - !Ref BedrockAccessPolicy
       # Optional CloudWatch access for monitoring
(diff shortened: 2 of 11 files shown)


wirjo:

### 1. Does `${AWS::Partition}` resolve to `aws` in commercial regions?

**Yes — zero change for existing deployments.** `${AWS::Partition}` is a CloudFormation pseudo-parameter that resolves to `aws` in all commercial regions, `aws-us-gov` in GovCloud, and `aws-cn` in China regions. For every existing commercial deployment, the rendered ARN is byte-for-byte identical to the previous hardcoded `arn:aws:...` value. CloudFormation will detect no diff and skip updating those resources.

### 2. Could the partition substitution break any existing CFN stacks on update?

**No replacement risk.** The change is from a static string (`arn:aws:iam::aws:policy/...`) to a `!Sub` expression that resolves to the exact same string in commercial regions. CloudFormation evaluates the template, sees the resolved value is unchanged, and triggers no update. The affected properties are all IAM `ManagedPolicyArns`, trust policy `Resource`/`Principal` fields, and Lambda permission `SourceArn` values — none of which cause resource replacement when the resolved value doesn't change. Even if CloudFormation did register a "modification", these are all in-place-updatable properties (not replacement triggers like `FunctionName` or `RoleName`).

### 3. Are service principals (e.g., `ecs-tasks.amazonaws.com`) correctly left unchanged?

**Yes, correctly handled.** Service principals like `ecs-tasks.amazonaws.com`, `lambda.amazonaws.com`, `codebuild.amazonaws.com`, `elasticloadbalancing.amazonaws.com`, and `apigateway.amazonaws.com` are NOT partition-scoped — they use the same DNS-style principal across all partitions. The PR correctly leaves all service principal strings untouched and only modifies the ARN strings (which have the `arn:<partition>:` prefix). Good precision here.

### 4. Does the ELB account map addition cause any issues for existing stacks?

**No issues.** The `ELBServiceAccounts` mapping in `landing-page-distribution.yaml` gains two new entries (`us-gov-west-1` and `us-gov-east-1`). Adding entries to a CloudFormation `Mappings` section has zero effect on existing stacks — the `!FindInMap` only resolves against the *current* region, so commercial region deployments will never reference these new keys. The values are only used when deploying in GovCloud. This is purely additive.

### Verdict

**Safe — zero regression risk for commercial deployments.** This is a textbook partition-portability fix: `${AWS::Partition}` resolves identically in commercial regions, all service principals are correctly left as-is, and the new mapping entries are only exercised in GovCloud. The addition of the `TestArnPartitionPortability` guard test (scanning all templates for hardcoded `arn:aws:`) is an excellent regression preventer that will catch future violations. The Python `aws_partition_for_region()` helper for hand-built ARNs in `init.py` is the correct approach for code paths that can't use CloudFormation pseudo-parameters.

Thank you @wantmys2000 for the contribution! The guard test that scans all templates for hardcoded partitions is particularly well done — it ensures this class of bug can't regress silently. The GovCloud ELB account map entries also show attention to the full deployment surface (not just ARN syntax but also runtime lookup correctness).

wirjo:
@scouturier LGTM — zero regression risk for commercial deployments, GovCloud fix is correct. Can merge when ready! ✅

wirjo (approved):

wirjo:

wirjo:

PR #692, opened 2026-06-26
fix(okta): accept Okta custom URL domains in OktaDomain validation

wantmys2000:
## Why
`bedrock-auth-okta.yaml` validated `OktaDomain` with `^[a-zA-Z0-9][a-zA-Z0-9-]*\.okta(-emea)?\.com$`, accepting only `*.okta.com` / `*.okta-emea.com`. That **rejects valid Okta issuers**, and blocked a real deployment:

- **Okta custom URL domains** — enterprises CNAME their Okta org to their own domain (e.g. `login.company.com`). This is a first-class, supported Okta feature.
- **Okta preview orgs** (`*.oktapreview.com`).
- **Okta for Government** orgs (`*.okta-gov.com`) — relevant for GovCloud.

The value is only substituted into `https://${OktaDomain}` for the IAM OIDC provider URL, so it needs to be a valid hostname — it does not need to be on `okta.com`.

## What
- Replace the okta.com-specific pattern with a fully-qualified-domain-name check:
  `^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$`
  Accepts org **and** custom domains while still forbidding a scheme, path, port, whitespace, or empty labels — so it stays safe when substituted into the `!Sub` URL.
- Clarify the parameter `Description` and `ConstraintDescription`.
- Add a regression test (`test_okta_domain_pattern.py`) asserting the pattern accepts org + custom + `okta-gov` domains and rejects `https://`, paths, ports, trailing dots, empty labels, and single-char TLDs.

The CLI-side validator (`validate_oidc_provider_domain`) is already a generic FQDN check, so no Python change is needed — the template was the only blocker.

## Testing
- `cd source && poetry run pytest tests/ -q` → 1353 passed, 22 skipped, 0 failures.
- `cfn-lint deployment/infrastructure/bedrock-auth-okta.yaml` → clean (exit 0).
- ruff check + format clean.

### DIFF
--- a/deployment/infrastructure/bedrock-auth-okta.yaml
+++ b/deployment/infrastructure/bedrock-auth-okta.yaml
@@ -12,9 +12,19 @@ Parameters:
 
   OktaDomain:
     Type: String
-    Description: Your Okta domain (e.g., company.okta.com)
-    AllowedPattern: '^[a-zA-Z0-9][a-zA-Z0-9-]*\.okta(-emea)?\.com$'
-    ConstraintDescription: Must be a valid Okta domain
+    Description: >-
+      Your Okta domain — an Okta org domain (e.g., company.okta.com,
+      company.okta-emea.com, company.oktapreview.com, company.okta-gov.com) or an
+      Okta custom URL domain hosted on your own domain (e.g., login.company.com).
+    # Validate as a fully-qualified domain name rather than hard-coding okta.com.
+    # Okta supports custom URL domains (CNAME'd to the org) and government orgs
+    # (okta-gov.com), so the previous '.okta(-emea).com' pattern rejected valid
+    # issuers. This still substitutes safely into https://${OktaDomain}: it forbids
+    # a scheme, path, port, whitespace, or empty labels. (Issue: custom Okta domains.)
+    AllowedPattern: '^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$'
+    ConstraintDescription: >-
+      Must be a valid fully-qualified domain name with no scheme, path, or port
+      (e.g., company.okta.com or login.company.com)
 
   OktaClientId:
     Type: String
--- a/source/tests/test_okta_domain_pattern.py
+++ b/source/tests/test_okta_domain_pattern.py
@@ -1,0 +1,86 @@
+"""Regression test: OktaDomain AllowedPattern must accept Okta custom URL domains.
+
+The original pattern '^[a-zA-Z0-9][a-zA-Z0-9-]*\\.okta(-emea)?\\.com$' only allowed
+*.okta.com / *.okta-emea.com, which blocked:
+  - Okta custom URL domains hosted on the customer's own domain (login.company.com)
+  - Okta preview orgs (*.oktapreview.com)
+  - Okta for Government orgs (*.okta-gov.com) — relevant for GovCloud
+The domain is substituted into `https://${OktaDomain}` for the IAM OIDC provider,
+so it must be a valid FQDN with no scheme/path/port, but need not be on okta.com.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+import yaml
+
+INFRA_DIR = Path(__file__).resolve().parents[2] / "deployment" / "infrastructure"
+
+
+class CfnLoader(yaml.SafeLoader):
+    pass
+
+
+def _cfn_tag_constructor(loader, tag_suffix, node):
+    if isinstance(node, yaml.ScalarNode):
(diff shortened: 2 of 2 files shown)


wirjo:
## Review

**Risk: Very Low** — backwards compatible regex change + regression tests.

Great catch @wantmys2000! This is a real deployment blocker for enterprises using Okta custom URL domains (which is a very common enterprise pattern). The fix is clean:

- ✅ FQDN regex is correct and well-tested (accepts valid domains, rejects scheme/path/port/whitespace)
- ✅ All previously-valid okta.com domains still match
- ✅ Regression test with parametrized cases ensures this doesn't regress
- ✅ Guard test prevents re-introducing okta.com hardcoding
- ✅ CLI-side validator was already correct (generic FQDN) — only the CFN template was blocking

Thank you for the contribution — this unblocks a real customer deployment. The test coverage is particularly appreciated. 🎉

**Verdict:** ✅ Ship it.

wirjo:
## Backwards Compatibility & Regression Risk Analysis

### Previously-valid domains still accepted?

**Yes** — all realistic previously-valid values pass the new pattern:

| Domain | Old | New |
|--------|-----|-----|
| `company.okta.com` | ✅ | ✅ |
| `my-company.okta.com` | ✅ | ✅ |
| `company.okta-emea.com` | ✅ | ✅ |
| `abc123.okta.com` | ✅ | ✅ |
| `a.okta.com` | ✅ | ✅ |

**One theoretical edge case:** The old pattern accepted labels ending in a hyphen (e.g. `a-.okta.com`) because `[a-zA-Z0-9-]*` has no end-of-label constraint. The new pattern correctly rejects these per RFC 952/1123. This is not a real regression — Okta never issues such domains, and they are DNS-invalid anyway. The new pattern is actually *more correct* here.

### Could the new pattern accept dangerous values?

**The pattern itself is safe** — it correctly rejects:
- Schemes (`https://evil.com`) ❌
- Paths (`evil.com/path`) ❌
- Ports (`evil.com:8080`) ❌
- Query strings (`evil.com?foo=bar`) ❌
- IP addresses (`127.0.0.1`, `169.254.169.254`) ❌
- Null bytes, spaces, special chars ❌
- Labels starting/ending with hyphens ❌
- Empty labels / double dots ❌
- Single-char TLDs ❌

Since the value is substituted into `https://${OktaDomain}` for the OIDC Provider URL, the critical question is: can it produce an unsafe URL? **No** — the pattern guarantees a syntactically valid hostname with no scheme, path, port, or special characters.

**However, it does expand the _domain scope_:**
- Old: **only** `*.okta.com` and `*.okta-emea.com`
- New: **any** syntactically valid FQDN (e.g. `evil.com`, `attacker.org`)

This is an intentional design choice (supporting custom URL domains, `*.oktapreview.com`, `*.okta-gov.com`). The risk is mitigated by:
1. Deploying this stack requires privileged AWS access (console/CLI)
2. The OIDC provider still requires a matching `ThumbprintList` + `ClientIdList` to function
3. The person setting `OktaDomain` is the stack deployer — if they control the AWS account, restricting them to `*.okta.com` provides no real security benefit
4. The parameter name + description still clearly indicate intended usage

**Verdict on attack surface:** Low risk. This is defense-in-depth relaxation, not a vulnerability introduction. The deployer is trusted.

### CloudFormation update impact on existing stacks?

**No impact.** Specifically:

1. **`AllowedPattern` is validated only at deploy/update time** when the parameter value is provided. It is not stored as a resource property.
2. **Existing stacks** with `OktaDomain=company.okta.com` will not experience drift — the parameter value hasn't changed, only the validation rule in the template.
3. **The `OktaOIDCProvider` resource** uses `!Sub 'https://${OktaDomain}'` — since the parameter value is unchanged, no physical resource change occurs.
4. **No resource replacement** will be triggered. The `AWS::IAM::OIDCProvider` URL property is immutable (replacement required if changed), but it won't change here.
5. **Stack updates** with the same parameter value will pass the new (broader) validation without issue.

### Verdict

✅ **Safe to merge.** 

- Full backwards compatibility for all realistic Okta domains
- No dangerous inputs accepted (pattern is a proper FQDN validator)
- No CloudFormation replacement/drift risk for existing stacks
- The domain-scope expansion is intentional and well-justified (custom URL domains, preview, GovCloud)
- Minor theoretical regression (hyphen-ending labels) is actually a correctness improvement

The PR is well-motivated by real deployment blockers and includes regression tests. Good to go.

wirjo:
@scouturier LGTM — backwards compatible FQDN pattern, well-tested. Can merge when ready! ✅

wirjo (approved):

wirjo:

wirjo:

PR #702, opened 2026-06-27
fix(package): configure localhost OTEL endpoint in sidecar mode

wantmys2000:
## Why
`ccwb package` left telemetry **unconfigured** for sidecar profiles. The sidecar override to `http://localhost:4318` was applied *after* the endpoint was resolved from the profile or CloudFormation — inside an `if endpoint:` block:

```python
endpoint = getattr(profile, "otel_collector_endpoint", None)   # None in sidecar
if not endpoint:
    ... CloudFormation lookup ...                              # no monitoring stack → still None
if not endpoint:
    ... "No OTel collector endpoint found" warning + prompt ...
if endpoint:                                                   # never true in real sidecar
    if monitoring_mode == "sidecar":
        endpoint = "http://localhost:4318"                     # dead code here
```

Real sidecar deploys have no central monitoring stack, so `otel_collector_endpoint` is never saved and the CloudFormation lookup finds nothing. `endpoint` stayed `None`, the code fell into the warning/prompt path, and `settings.json` was written **without** `OTEL_EXPORTER_OTLP_ENDPOINT` or `CLAUDE_CODE_ENABLE_TELEMETRY` → Claude Code emitted no telemetry to the local collector.

The existing sidecar test masked this — its fixture pre-set `otel_collector_endpoint="https://alb.example.com"`, so `endpoint` was non-None and the override ran. That never happens in a real sidecar deployment.

## What
- Resolve the sidecar endpoint to `http://localhost:4318` **up front**, before the profile/CloudFormation/prompt resolution (which only applies to central mode), so it no longer depends on a pre-existing endpoint.
- Skip the CloudFormation stack lookup entirely in sidecar mode (there's no monitoring stack to query).
- Remove the now-redundant post-resolution override.
- Add a regression test with a sidecar profile and **no** saved endpoint (`otel_collector_endpoint=None`) asserting `OTEL_EXPORTER_OTLP_ENDPOINT == http://localhost:4318` and `CLAUDE_CODE_ENABLE_TELEMETRY == "1"`. **Verified it fails without the fix and passes with it.**

## Relationship to #678
#678 restored the otelcol build + collector-config generation and *introduced* the localhost override — but placed it after endpoint resolution, where it can't run for the normal sidecar case. This moves it to where it actually takes effect.

## Testing
- `poetry run pytest tests/cli/commands/test_package_oidc_sidecar.py -q` → 26 passed.
- Package/sidecar suite (117 tests) green. ruff check + format clean.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -3772,10 +3772,25 @@ Available metrics include:
 
             # If monitoring is enabled, add telemetry configuration
             if profile.monitoring_enabled:
-                # Try profile first (saved by ccwb deploy), fall back to CloudFormation query
-                endpoint = getattr(profile, "otel_collector_endpoint", None)
+                _monitoring_mode = getattr(profile, "monitoring_mode", "central")
 
-                if not endpoint:
+                # Sidecar mode: Claude Code always sends to the local otelcol on
+                # localhost:4318. There is no central monitoring stack to read a
+                # CollectorEndpoint from, so resolve the endpoint up front and skip
+                # the profile/CloudFormation/prompt resolution below (which only
+                # applies to central mode). Doing this here — rather than as an
+                # override after resolution — is what actually configures telemetry
+                # for sidecar packages: real sidecar deploys have no saved
+                # otel_collector_endpoint, so the old post-resolution override never
+                # ran and telemetry was silently left unconfigured.
+                if _monitoring_mode == "sidecar":
+                    endpoint = "http://localhost:4318"
+                else:
+                    # Central mode: try profile first (saved by ccwb deploy), then
+                    # fall back to CloudFormation query.
+                    endpoint = getattr(profile, "otel_collector_endpoint", None)
+
+                if not endpoint and _monitoring_mode != "sidecar":
                     # Fall back to reading from CloudFormation stack outputs
                     # Try multiple possible stack name patterns
                     possible_stacks = [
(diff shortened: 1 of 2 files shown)


wirjo:
Thanks for catching this, @wantmys2000! 🎉 Great to have the community contributing and finding real bugs in production sidecar deployments. cc @scouturier

**Review summary:**

The fix is clean and well-reasoned. The core issue was that the sidecar endpoint override (`http://localhost:4318`) lived *inside* the `if endpoint:` block — but real sidecar deploys have no saved `otel_collector_endpoint`, so `endpoint` was `None`, the block was never entered, and telemetry was silently left unconfigured.

Moving the sidecar resolution *before* the CloudFormation/prompt path ensures the endpoint is always set for sidecar mode regardless of what's in the profile. The guard `if not endpoint and _monitoring_mode != "sidecar"` correctly prevents sidecar packages from falling into the CloudFormation stack lookup.

**Checks performed:**
- ✅ All 26 tests in `test_package_oidc_sidecar.py` pass (including the new regression test)
- ✅ Related manifest tests still pass (central mode, sidecar, IDC zero-binary)
- ✅ Ruff lint clean
- ✅ Central mode is unaffected — endpoint resolution via profile/CloudFormation is only entered when `_monitoring_mode != "sidecar"`
- ✅ The new regression test properly exercises the failure case (sidecar + `endpoint=None`) and asserts both the endpoint value and that telemetry is enabled

No issues found. Ship it!

wirjo (approved):
LGTM — clean fix for a real sidecar telemetry bug. Logic is sound, tests cover both the happy path and the regression case, no risk to central mode. Approving.

wirjo:

wirjo:

PR #709, opened 2026-06-28
fix(init): update wizard terminology — Claude Desktop, legacy Cognito, Historical Usage Analytics

wirjo:
## Summary

Updates the `ccwb init` setup wizard terminology to reflect current product naming and clarify feature positioning.

### Changes

| # | Area | Before | After |
|---|------|--------|-------|
| 1 | Product rename | "CoWork Support" / "Enable CoWork support?" | "Claude Desktop Support" / "Enable Claude Desktop support?" |
| 2 | Federation | "Cognito Identity Pool" (no qualifier) | "Cognito Identity Pool (legacy)" with description "use only for existing deployments" |
| 3 | Analytics | "Analytics Pipeline" / "Enable analytics?" | "Historical Usage Analytics" / "Enable historical usage analytics?" with description "Query historical per-user usage with SQL via AWS Athena (S3 data lake, ~$5/month)" |

### Details

**1. CoWork → Claude Desktop rename**
- User-facing prompt: "Enable CoWork support?" → "Enable Claude Desktop support?"
- Section heading: "CoWork Support" → "Claude Desktop Support"  
- Confirmation messages updated to reference "Claude Desktop"
- Internal comments updated for clarity
- Config keys (`cowork_3p`) and CF parameter names (`CoWorkServiceToken`) unchanged

**2. Cognito federation marked as legacy**
- Choice text: `"Cognito Identity Pool"` → `"Cognito Identity Pool (legacy)"`
- Added guidance: "use only for existing deployments"
- Direct STS remains marked as "(recommended)"

**3. Analytics → Historical Usage Analytics**
- Section heading: "Analytics Pipeline" → "Historical Usage Analytics"  
- Prompt: "Enable analytics?" → "Enable historical usage analytics?"
- Added description: "Query historical per-user usage with SQL via AWS Athena (S3 data lake, ~$5/month)"
- Confirmation: "Analytics pipeline will be deployed" → "Historical analytics (S3 + Athena) will be deployed"
- Sidecar comment: clarified "no historical analytics (PromQL dashboards still work)"
- Resource summary: "analytics data streaming/storage" → "historical usage data streaming/storage"

### Scope

- **File changed**: `source/claude_code_with_bedrock/cli/commands/init.py` only
- **Config keys unchanged**: `cowork_3p`, `analytics`, `federation_type` — no breaking changes to saved profiles
- **Review table**: "Athena SQL Pipeline" label was already correct (unchanged)
- **Tests**: All 36 tests in `test_init.py` pass ✓
[...]

PR #647, opened 2026-06-22
feat(websearch): inject AgentCore web search gateway into CoWork MDM config

chiararelandini:
## Why

Final code PR of the Claude Cowork web search series. #607 (template, merged) and #641 (deploy wiring) provision the AgentCore Gateway; this PR makes it **usable from Claude Cowork** by injecting the gateway as a managed MCP server into the generated Cowork MDM config.

> ⚠️ **Stacked on #641 — please merge that first.** This branch is based on `feat/websearch-deploy-wiring` (#641), so until #641 lands the diff here also shows #641's commits. The PR4-specific change is the last commit (`feat(websearch): inject AgentCore web search gateway into CoWork MDM config`).

## What

When `web_search_enabled`, `ccwb package` and `ccwb cowork generate` emit a `managedMcpServers` **`agentcore-websearch`** entry into the generated MDM config (`.json` / `.mobileconfig` / `.reg`):

- **`cowork_3p.py`** — `add_websearch_mcp_config()`:
  - Resolves the gateway MCP endpoint **profile-first** (`websearch_gateway_url`, saved by `ccwb deploy`) with a CloudFormation fallback (`GatewayMcpEndpoint` output); ensures exactly one `/mcp` suffix.
  - Builds a native **OAuth authorization-code** entry against the same IdP (`oauth.clientId` + `authorizationServer` = OIDC issuer + `scope` + `localhost` callback) — **no secret** is stored in the MDM config. Reuses the `localhost:<redirect_port>` callback already registered for Claude Code.
  - `_oidc_issuer_for_profile` reuses deploy.py's per-provider discovery-URL derivation (single source of truth) stripped of the well-known suffix, so the issuer matches what the gateway authorizer validates.
  - Preserves admin-defined `managedMcpServers` and de-dupes the web search entry by name.
- Wired into `package.py` (`_generate_cowork_3p_mdm_config`) and the standalone `cowork generate` command, right after `add_monitoring_config`.

## Relationship with #640

This is the **Cowork** counterpart to #640's **Claude Code** injection (`mcpServers.agentcore-websearch` in `settings.json` via the credential helper). The two **coexist** on the same #607 gateway — different surfaces, no file overlap. See the plan on #641.

## Testing Evidence

[...]

PR #717, opened 2026-06-29
feat(init): add web search opt-in to ccwb init wizard

wirjo:
## Why

The websearch feature series (#641, #647) tells users to "re-run ccwb init" to configure web search, but no PR actually adds the init wizard prompt. This closes that gap.

## What

Adds an **"Enable web search?"** confirm prompt to the `ccwb init` wizard:

```
Web Search
Give Claude a hosted web-search tool via Amazon Bedrock AgentCore Gateway.
Currently available in us-east-1 only.

? Enable web search? (y/N)
✓ Web search gateway will be deployed to us-east-1
  Run 'ccwb deploy websearch' after init to provision the gateway.
```

### Behaviour

| Auth Type | Prompt? | Default |
|-----------|---------|---------|
| OIDC (cognito, okta, azure, auth0, google, generic) | Yes | No |
| IDC | Skipped with explanation | Forced disabled |

### Config mapping

- Wizard: `config["web_search"]["enabled"]`
- Profile: `web_search_enabled` (bool, default False)
- Round-trips on re-init: profile → config → profile ✓

## Dependencies

> ⚠️ **Stacked on #709** (init wizard terminology). Merge #709 first, then this.
>
> Also depends on #641 landing (which adds `web_search_enabled` to the Profile dataclass). Without #641, the field silently gets `setattr`'d but isn't in the dataclass yet — harmless but technically incomplete until both merge.

## Credits

This prompt integrates the init wizard design from @antkawam's original PR #640 into the decomposed series.

Co-authored-by: antkawam <antkawam@users.noreply.github.com>
Co-authored-by: chiararelandini <chiararelandini@users.noreply.github.com>

## Review

@chiararelandini — would appreciate your review since this integrates with your deploy wiring (#641). The wizard sets `web_search_enabled=True` on the profile, which your `deploy-all` loop checks to include the websearch stack.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -1472,8 +1472,7 @@ class InitCommand(Command):
         ws_region = config.get("web_search", {}).get("region") or WEBSEARCH_SUPPORTED_REGIONS[0]
         console.print(
             "Give Claude a hosted web-search tool via Amazon Bedrock AgentCore Gateway.\n"
-            f"[dim]Deploys to {ws_region}. Requires an OIDC provider (not IDC). "
-            "See: assets/docs/WEB_SEARCH.md[/dim]"
+            f"[dim]Currently available in {ws_region} only.[/dim]"
         )
 
         # Only show the prompt for OIDC-compatible providers
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -29,7 +29,7 @@ from claude_code_with_bedrock.cli.utils.progress import WizardProgress
 from claude_code_with_bedrock.cli.utils.validators import (
     validate_oidc_provider_domain,
 )
-from claude_code_with_bedrock.config import Config, Profile
+from claude_code_with_bedrock.config import WEBSEARCH_SUPPORTED_REGIONS, Config, Profile
 
 
 def validate_identity_pool_name(value: str) -> bool | str:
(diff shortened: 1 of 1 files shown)


wirjo:
## PR Review — Backwards Compatibility & Regression Risk

### Backwards compatibility ✅

1. **Existing `ccwb init` flow works if web_search config is absent** — In `_check_existing_deployment`, legacy profiles without `web_search_enabled` are handled via `getattr(profile, "web_search_enabled", False)`, which correctly defaults to `False`. The `Profile` dataclass field also defaults to `False`.

2. **Deploy skips websearch when `web_search_enabled` is False/missing** — The deploy-all path guards with `if getattr(profile, "web_search_enabled", False)` before appending the websearch stack. Explicit `ccwb deploy websearch` also gates on the same check and prints a helpful error + redirect to re-run init.

3. **Existing profiles without `web_search` field handled gracefully** — `test_profile_without_websearch_fields_loads_with_defaults()` explicitly tests this: a legacy dict without any websearch keys loads with `web_search_enabled=False`, `websearch_region=None`, etc. No migration step required.

4. **Prompt only shows for OIDC providers (skip for IDC)** — The init wizard checks `auth_type == "idc"` and silently sets `enabled = False` with an explanatory message, skipping the interactive prompt entirely. OIDC providers get the confirm prompt.

### Regression risk ✅

1. **New prompt cannot break init flow** — Uses the same `questionary.confirm(...).ask()` pattern already used 5+ times in the same function. No new imports needed (questionary already imported). If user Ctrl+C aborts, `ask()` returns `None` which is falsy — same pre-existing behavior as all other confirm prompts in the file.

2. **No existing config fields modified** — The PR only *adds* `web_search` as a new nested dict in config and `web_search_enabled` as a new Profile field. No existing fields are renamed, moved, or re-typed.

3. **Cannot accidentally enable web search for existing re-inits** — On re-init, the prompt default is `config.get("web_search", {}).get("enabled", False)`. For users who never enabled it, this defaults to `False`. For users who previously enabled it, it preserves their choice as the default. Either way, the user must actively confirm.

4. **Default (No) means zero infrastructure change** — When `web_search_enabled` is `False`, no websearch stack is appended to the deploy list, destroy skips it, and package ignores it. Zero behavioral change for existing users.

### Code quality

**Good:**
- Comprehensive test coverage (249-line test file covering backward compat, provider guards, region validation, CFN param derivation, and deploy-loop failure semantics).
- The websearch stack failure is non-fatal in deploy-all (continues with a warning) — good UX for an optional feature.
- Provider preflight validation is thorough with per-provider checks (Cognito needs pool_id, Azure needs issuer, generic needs issuer_url).
- Cross-region deployment handled correctly (parallels CodeBuild cross-region pattern).

**Minor nits (non-blocking):**
- `provider_type = config.get("provider_type")` at L1295 of init.py is assigned but never used in the web search block — the gating is done via `auth_type`. Harmless dead code (perhaps reserved for future provider-specific init messaging), but could be removed for clarity.
- The `questionary.confirm().ask()` → `None` on abort is a pre-existing pattern across all prompts in init. Not introduced here, but worth noting that `config["web_search"]["enabled"] = None` is theoretically possible (though downstream `getattr(profile, "web_search_enabled", False)` treats it as falsy).

### Verdict

**Low risk, safe to merge.** The PR follows established patterns (dataclass field defaults, `getattr` guards, questionary prompts), adds comprehensive tests, and the opt-in default of `False` guarantees zero behavioral change for existing users. The feature is cleanly gated behind a config flag at every touchpoint (init, deploy, destroy, package).

PR #718, opened 2026-06-29
fix(websearch): validate Cognito tokens via client_id claim

chiararelandini:
## What

The merged AgentCore Gateway template (#607) hard-codes the CUSTOM_JWT authorizer to validate the inbound token's `aud` claim:

```yaml
CustomJWTAuthorizer:
  DiscoveryUrl: !Ref DiscoveryUrl
  AllowedAudience:
    - !Ref ClientId
```

on the assumption that every OIDC provider places the client id in `aud`. **This is not true for Amazon Cognito.** Cognito **access tokens** carry the app client id in the **`client_id`** claim and have **no `aud`** claim, so the gateway rejects them and web search fails to authorize.

This restores the per-IdP claim selection that existed in the original PR before it was simplified at merge:

- new parameter `JwtValidationMode` (`audience` | `client_id`)
- new condition `UseClientId`
- the authorizer now validates via **`AllowedClients`** in `client_id` mode (Cognito) and **`AllowedAudience`** in `audience` mode (Entra ID, Okta, Auth0, Google, generic OIDC).

`Default: audience` preserves the current behavior, so **existing OIDC deployments are unaffected**.

## Why this is correct

The official AWS examples configure the Cognito gateway authorizer with `allowedClients` (not `allowedAudience`):

```python
auth_config = { "customJWTAuthorizer": { "allowedClients": [cognito_client_id], "discoveryUrl": cognito_discovery_url } }
```

Per the AgentCore docs, `allowedClients` validates the `client_id` claim and `allowedAudience` validates the `aud` claim — so the right field depends on the IdP.

## Scope

- Template-only change. Backward compatible (default mode = `audience`).
- The deploy CLI sets `JwtValidationMode` automatically from `provider_type` (`cognito` → `client_id`, otherwise `audience`). That wiring lives in the deploy PR #641 (`build_websearch_params`) and is **not** part of this PR, since it depends on code not yet on `beta`. This template PR should merge first.

## Testing

- `cfn-lint` clean (only the pre-existing W3037 false positive on the newer `bedrock-agentcore:InvokeWebSearch` action, unrelated to this change).
- Verified the conditional renders `AllowedClients: [ClientId]` for `JwtValidationMode=client_id` and `AllowedAudience: [ClientId]` for `audience`.

### DIFF
--- a/deployment/infrastructure/bedrock-agentcore-gateway.yaml
+++ b/deployment/infrastructure/bedrock-agentcore-gateway.yaml
@@ -43,8 +43,23 @@ Parameters:
     Type: String
     Default: ''
     Description: >-
-      OIDC client_id (required when AuthType=oidc, ignored for idc). The
-      gateway validates the id_token's 'aud' claim against this value.
+      OIDC client_id (required when AuthType=oidc, ignored for idc). Validated
+      against the token's 'aud' or 'client_id' claim depending on
+      JwtValidationMode.
+
+  JwtValidationMode:
+    Type: String
+    Default: audience
+    AllowedValues:
+      - audience
+      - client_id
+    Description: >-
+      How the CUSTOM_JWT authorizer validates the inbound token (OIDC only).
+      'audience' validates the token 'aud' claim via AllowedAudience — use for
+      Entra ID, Okta, Auth0, Google and generic OIDC, which place the client id
+      in 'aud'. 'client_id' validates the 'client_id' claim via AllowedClients —
+      use for Amazon Cognito, whose access tokens carry 'client_id' and have no
+      'aud' claim. The deploy CLI sets this automatically from the provider type.
 
   DomainExcludeList:
     Type: CommaDelimitedList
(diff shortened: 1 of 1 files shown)


chiararelandini:
Closing — the premise of this PR is incorrect.

Live testing with Claude Cowork showed the bearer presented to the gateway is the user's Cognito **id_token** (`aud` = client_id, and **no** `client_id` claim). So `AllowedAudience` (the existing #607 default) is the correct validation mode, and it holds for **every** OIDC provider (an id_token's `aud` is always the client_id). The `client_id`/`AllowedClients` path — and the `JwtValidationMode` toggle this PR added — are unnecessary, and the cognito→client_id default is actually wrong for this flow.

The real fix is on the client side: the Cowork MDM `managedMcpServers` entry must be a remote MCP entry with a `headersHelper` that injects the id_token bearer (not `server:websearch/provider:custom`, and not the dev-only `oauth` object). That will be handled in the Cowork MDM PR. Apologies for the noise.

PR #721, opened 2026-06-29
fix(package): resolve Cognito identity pool ID from stack outputs, not profile name

### LINKED ISSUE #720: ccwb package writes Cognito identity pool NAME instead of ID — breaks auth for all distributed users
## Summary

When using **Cognito Identity Pool** federation, `ccwb package` writes the identity pool **name** (e.g. `claude-code-auth`) into the generated `config.json` as `identity_pool_id`, instead of the real pool **ID** (`<region>:<uuid>`). At runtime the bundled `credential-process` then calls Cognito `GetId` with an invalid pool identifier and fails:

```
An error occurred (ValidationException) when calling the GetId operation:
1 validation error detected: Value 'claude-code-auth' at 'identityPoolId'
failed to satisfy constraint: Member must satisfy regular expression pattern: [\w-]+:[0-9a-f-]+
```

This blocks AWS credential exchange (so Bedrock calls fail / fall back to expired creds) **and** prevents the monitoring/OTEL id-token from being cached — so per-user telemetry attribution silently degrades to anonymous. Every distributed package built for a Cognito deployment is affected.

## Environment

- ccwb version: `main` (reproduced at `36021b9`)
- Federation type: **Cognito Identity Pool**
- Auth provider: AWS Cognito User Pool
- OS: macOS (platform-independent — the bug is in package config generation)

## Steps to reproduce

1. `ccwb init` with Federation Type = **Cognito Identity Pool**
2. `ccwb deploy` (creates the Identity Pool; the auth stack exposes output `IdentityPoolId = <region>:<uuid>`)
3. `ccwb package`
4. Inspect the generated `dist/<profile>/<ts>/config.json`:

```json
"identity_pool_id": "claude-code-auth"   // <-- the NAME, not "<region>:<uuid>"
```

5. Install and run `claude` (or `ccwb test`) → `GetId` ValidationException; authentication fails.

## Root cause

`source/claude_code_with_bedrock/cli/commands/package.py` resolves the federation identifier with a branch that short-circuits for Cognito:

```python
elif federation_type != "direct" and getattr(profile, "identity_pool_name", None):
    identity_pool_id = profile.identity_pool_name      # BUG: name used as the pool ID
    ...
else:
    # correct path — reads the real ID from the stack output, but is never
    # reached for Cognito because the elif above always matches
    identity_pool_id = stack_outputs.get("IdentityPoolId")
```

`profile.identity_pool_name` is only a human-readable name. The real pool ID does not exist until `ccwb deploy` creates the pool and is exposed as the CloudFormation stack output `IdentityPoolId`. The `Profile` schema (`config.py`) intentionally has **no** `identity_pool_id` field, so the package step must read it from the stack — but the `elif` makes the correct stack-output read dead code for Cognito.

The same buggy pattern appears **twice**: in the main package path (`run()`) and in the `--regenerate-installers` path.

## Impact

- **Severity: high.** Authentication is completely broken for any Cognito Identity Pool deployment packaged with the affected code.
- Direct STS federation is **not** affected (the profile stores the real role ARN, so its shortcut is valid).
- Symptom is easy to misdiagnose because the *login/JWT decode succeeds* — only the subsequent `GetId` call fails.

## Proposed fix

For Cognito federation, always resolve `IdentityPoolId` from the CloudFormation stack outputs (the profile only holds the name). Keep the Direct STS profile shortcut. I have a PR that extracts a shared `_resolve_federation` helper used by both code paths and adds regression tests.

PR: will link from this issue.

joaogac:
Closes #720

## Problem

For **Cognito Identity Pool** federation, `ccwb package` wrote the identity pool **name** (e.g. `claude-code-auth`) into `config.json` as `identity_pool_id`, instead of the real pool **ID** (`<region>:<uuid>`). At runtime `credential-process` then calls Cognito `GetId` with an invalid identifier and fails:

```
ValidationException ... Value 'claude-code-auth' at 'identityPoolId' failed to
satisfy constraint: Member must satisfy regular expression pattern: [\w-]+:[0-9a-f-]+
```

This blocks AWS credential exchange (Bedrock auth) and prevents the monitoring/OTEL id-token from being cached, so per-user telemetry silently degrades to anonymous. Every Cognito package built from the affected code is broken.

## Root cause

The federation-identifier resolution short-circuited for Cognito:

```python
elif federation_type != "direct" and getattr(profile, "identity_pool_name", None):
    identity_pool_id = profile.identity_pool_name   # BUG: name used as the ID
else:
    identity_pool_id = stack_outputs.get("IdentityPoolId")  # correct, but dead code for Cognito
```

`profile.identity_pool_name` is only a name; the real pool ID is created by `ccwb deploy` and exposed as the CloudFormation stack output `IdentityPoolId`. The `Profile` schema has no `identity_pool_id` field by design, so the package step must read it from the stack. The same buggy pattern existed in **two** places: the package path (`run()`) and the `--regenerate-installers` path.

## Fix

- Extract a shared `_resolve_federation(profile, console)` helper used by both code paths. For Cognito federation the identity pool ID is **always** read from the stack output `IdentityPoolId`.
- **Direct STS is unchanged** — the profile role-ARN shortcut is preserved (the profile stores the real ARN), and SSO-disabled handling is preserved.
- De-duplicates the previously copy-pasted resolution logic.

## Tests

Added `TestResolveFederation` (4 cases):
- Cognito resolves the pool ID from the stack output (and explicitly **not** the name) — regression guard
- Direct STS uses the profile role ARN with no stack lookup
- Missing `IdentityPoolId` in stack outputs fails gracefully (no name fallback)
- SSO-disabled returns no identifier and performs no stack lookup

Full suite: **920 passed** (+4 new), no new failures. (One pre-existing unrelated e2e failure about `ap-southeast-2` in `test_context_show_displays_profile_details` is present on `main` independent of this change.)

## Backward compatibility

- Direct STS federation: identical behavior.
- Cognito federation: now produces a valid `identity_pool_id` (`<region>:<uuid>`) — previously broken.
- No schema/config changes; no changes to `_create_config` output shape.

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -273,45 +273,13 @@ class PackageCommand(Command):
             )
             return 1
 
-        # Get federation identifier — try profile first, fall back to CloudFormation
-        federation_type = profile.federation_type
-        identity_pool_id = None
-        federated_role_arn = None
-
-        if not getattr(profile, "sso_enabled", True):
-            # SSO disabled — no auth stack to query, no federation needed
-            console.print("[dim]SSO disabled — skipping auth stack lookup[/dim]")
-        elif federation_type == "direct" and getattr(profile, "federated_role_arn", None):
-            federated_role_arn = profile.federated_role_arn
-            console.print(f"[dim]Using role ARN from profile: {federated_role_arn}[/dim]")
-        elif federation_type != "direct" and getattr(profile, "identity_pool_name", None):
-            identity_pool_id = profile.identity_pool_name
-            console.print(f"[dim]Using identity pool from profile: {identity_pool_id}[/dim]")
-        else:
-            # Fall back to CloudFormation stack outputs
-            console.print("[yellow]Fetching deployment information from CloudFormation...[/yellow]")
-            stack_outputs = get_stack_outputs(
-                profile.stack_names.get("auth", f"{profile.identity_pool_name}-stack"), profile.aws_region
-            )
-
-            if not stack_outputs:
-                console.print("[red]Could not fetch stack outputs. Is the stack deployed?[/red]")
-                return 1
-
-            federation_type = stack_outputs.get("FederationType", profile.federation_type)
-
-            if federation_type == "direct":
-                federated_role_arn = stack_outputs.get("DirectSTSRoleArn")
-                if not federated_role_arn or federated_role_arn == "N/A":
-                    federated_role_arn = stack_outputs.get("FederatedRoleArn")
-                if not federated_role_arn or federated_role_arn == "N/A":
-                    console.print("[red]Direct STS Role ARN not found in stack outputs.[/red]")
-                    return 1
-            else:
-                identity_pool_id = stack_outputs.get("IdentityPoolId")
-                if not identity_pool_id:
-                    console.print("[red]Identity Pool ID not found in stack outputs.[/red]")
-                    return 1
+        # Resolve federation identifier (role ARN for direct STS, Identity Pool ID
+        # for Cognito). See _resolve_federation: Cognito ALWAYS reads the pool ID
+        # from CloudFormation stack outputs because identity_pool_name is only a
+        # name, not the "<region>:<uuid>" pool ID.
+        federation_type, identity_pool_id, federated_role_arn = self._resolve_federation(profile, console)
+        if getattr(profile, "sso_enabled", True) and not (identity_pool_id or federated_role_arn):
+            return 1
 
         # Welcome
         console.print(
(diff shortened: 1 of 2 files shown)


wirjo:

wirjo (approved):
**✅ Approved — Cognito Identity Pool fix only, other auth flows unaffected**

**What it fixes:**
`ccwb package` was writing `identity_pool_name` (e.g. `"claude-code-auth"`) into config.json as `identity_pool_id`. At runtime, credential-process calls Cognito `GetId` which fails regex validation (`[\w-]+:[0-9a-f-]+`). The fix extracts `_resolve_federation()` that ALWAYS reads the pool ID from CloudFormation stack outputs for Cognito.

**Scope: Cognito Identity Pool pathway ONLY**

The bug was in the build-time resolution shortcut: `identity_pool_id = profile.identity_pool_name` — matched only when `federation_type != "direct"` (i.e. Cognito). The fix ensures Cognito always reads the real UUID-format pool ID from stack outputs.

**Why other auth flows are unaffected:**

- **OIDC federation (Okta/Azure/Google/Generic):** Uses Direct STS (`AssumeRoleWithWebIdentity`). Routes to `get_aws_credentials_direct()` when `federation_type == "direct"`, using `federated_role_arn` — never touches `identity_pool_id`. `_resolve_federation()` exits early with the profile's cached ARN, zero stack lookups. Test confirms: `mock_stack.assert_not_called()`.

- **Direct STS federation:** Same path — profile stores the real IAM role ARN which is a valid identifier (not a name). Fix preserves this shortcut exactly.

- **IDC (IAM Identity Center):** Uses `_run_passthrough()` reading ambient AWS credentials. Never touches `identity_pool_id`. In `_resolve_federation()`, SSO-disabled profiles exit at the first guard returning `(type, None, None)`.

**Code quality:** Excellent — extracts duplicated logic into single helper, removes ~95 lines of copy-paste, adds 4 targeted regression tests. 920 tests passed.

PR #724, opened 2026-06-30
fix(package): IDC zero-binary settings.json must not reference credential-process

scouturier:
## Why

In IDC zero-binary mode (no quota configured), `ccwb package` ships no credential-process binary — auth is handled entirely via `aws sso login` + the `ClaudeCode` AWS SSO profile written by `install.sh`. However, `_create_claude_settings` still emitted three unresolvable `__CREDENTIAL_PROCESS_PATH__` placeholders into `settings.json`:

- `AWS_CREDENTIAL_PROCESS` — breaks every AWS SDK call (command not found)
- `awsCredentialExport` — Claude Code tries to invoke a non-existent binary for silent refresh
- `awsAuthRefresh` — same

The result: credentials silently fail on every invocation for any IDC zero-binary deployment.

## What

- Pass `is_idc_zero_binary` into `_create_claude_settings` and skip all three credential-process references when `True`
- The AWS SDK resolves credentials via `AWS_PROFILE` → `~/.aws/config` `ClaudeCode` SSO profile (written by `install.sh`)
- IDC+quota path (`is_idc_zero_binary=False`) is unchanged — it still packages the binary and keeps all three references

## Test plan

- [ ] 5 regression tests added to `TestIDCZeroBinarySettings` in `test_package_idc_zero_binary.py`
- [ ] Tests confirm `__CREDENTIAL_PROCESS_PATH__` absent, `AWS_CREDENTIAL_PROCESS` absent, `awsCredentialExport`/`awsAuthRefresh` absent in zero-binary settings
- [ ] Tests confirm `AWS_PROFILE=ClaudeCode` still present
- [ ] Tests confirm IDC+quota path still contains all credential-process references
- [ ] All 21 tests in `test_package_idc_zero_binary.py` pass

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -725,7 +725,14 @@ class PackageCommand(Command):
 
         # Always create Claude Code settings (required for Bedrock configuration)
         console.print("[cyan]Creating Claude Code settings...[/cyan]")
-        self._create_claude_settings(output_dir, profile, include_coauthored_by, profile_name, otel_resource_attributes)
+        self._create_claude_settings(
+            output_dir,
+            profile,
+            include_coauthored_by,
+            profile_name,
+            otel_resource_attributes,
+            is_idc_zero_binary=is_idc_zero_binary,
+        )
 
         # Generate CoWork 3P MDM configuration if enabled
         if profile.cowork_3p_enabled:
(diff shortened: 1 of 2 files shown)


wirjo (approved):
Code review: clean fix for IDC zero-binary settings. Correctly skips credential-process references when no binary is packaged. All tests pass. Very low risk — targeted conditional with no impact on OIDC/IDC+quota paths.

PR #726, opened 2026-06-30
fix(quota): reset idle users' daily counter in monitor alerts and CLI display

adiospeds:
## Why

An idle user whose daily token counter froze above the daily limit kept receiving a fresh **"Daily Token Quota EXCEEDED"** alert email **every new UTC day**, and `ccwb quota usage` kept showing the same stale over-limit percentage. Neither reset, even though the user had no recent activity.

The daily counter is only rolled over to a new day as a *side effect of new activity* — `update_quota_metrics()` is gated on a PromQL delta (`if delta <= 0: continue`), and idle users aren't returned by the query at all. So an idle user's `daily_tokens`/`daily_date` stay frozen until they use Claude Code again (or the month rolls over).

The **real-time enforcement** path (`quota_check`) already guards against this — it treats `daily_tokens` as `0` when the stored `daily_date` is not today — so users were **never wrongly blocked**. But two other readers lacked the same guard:

1. **quota_monitor threshold/alert step** read `daily_tokens` verbatim (and didn't even project `daily_date`), so a frozen value re-fired a daily `exceeded` alert against each new day's date-stamped dedup key.
2. **`ccwb quota usage` (`_get_user_usage`)** returned `daily_tokens` raw — it read `daily_date` but never compared it to today.

## What

Apply the same read-time stale-day guard `quota_check` already uses in both places:

- **quota_monitor** (`lambda_handler` Step 2): project `daily_date` in the scan and zero the daily counter via a new `_build_usage_entry()` helper when `daily_date != current_date`.
- **CLI** (`_get_user_usage`): zero `daily_tokens` on a stale day, and switch to tz-aware UTC (`datetime.now(timezone.utc)`) so the day boundary matches the Lambdas exactly (avoids an off-by-one at midnight).

After this change, all three surfaces — monitor (alerting), check (enforcement), CLI (display) — agree that *a daily counter whose `daily_date` is not today is `0`*.

Monthly counters are unaffected: they accumulate across the whole month and a new month uses a fresh sort key (`MONTH#YYYY-MM`). Genuine same-day over-limit alerts still fire (no over-correction).

## Tests

Per project rules (every bug fix includes a test that fails without the fix):

- `source/tests/test_quota_monitor_lambda.py` (new):
  - idle user with stale `daily_date` over the limit + no activity → **no** daily alert
  - active user, same-day over-limit → daily alert **is** generated (guards against over-correction)
  - `_build_usage_entry` zeros a stale daily counter while leaving the monthly total intact
- `source/tests/cli/commands/test_quota_usage_stale_day.py` (new):
  - stale `daily_date` → displayed `daily_tokens == 0`, monthly untouched
  - same-day → daily usage preserved
  - missing `daily_date` (legacy row) → treated as stale → `0`

Verified the regression tests fail on `beta` without the fix and pass with it. Full quota suite (91 tests) green.

## Affected files / tiers (per `.claude/rules/review-tiers.md`)

| File | Defect | Tier |
|------|--------|------|
| `deployment/infrastructure/lambda-functions/quota_monitor/index.py` | idle reset missing in scan/threshold step | Tier 2 |
| `source/claude_code_with_bedrock/cli/commands/quota.py` | `_get_user_usage` no stale-day guard | Tier 2 |
| `deployment/infrastructure/lambda-functions/quota_check/index.py` | reference (already correct) | — |

### DIFF
--- a/deployment/infrastructure/lambda-functions/quota_monitor/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_monitor/index.py
@@ -222,6 +222,25 @@ def update_quota_metrics(usage_data):
     print(f"Updated UserQuotaMetrics for {len(usage_data)} users")
 
 
+def _build_usage_entry(item, current_date):
+    """Build a usage_data entry for threshold checking, applying the stale-day guard.
+
+    Mirrors quota_check.get_user_usage: if the stored daily_date is not today
+    (UTC), the daily counter belongs to a prior day and must be treated as 0.
+    Otherwise an idle user whose daily_tokens froze above the limit gets a fresh
+    "daily exceeded" alert every new UTC day even though they had no activity.
+    Monthly (total_tokens) is unaffected — it accumulates across the whole month.
+    """
+    daily_tokens = float(item.get("daily_tokens", 0))
+    daily_date = item.get("daily_date")
+    if daily_date != current_date:
+        daily_tokens = 0
+    return {
+        "total_tokens": float(item.get("total_tokens", 0)),
+        "daily_tokens": daily_tokens,
+    }
+
+
 def lambda_handler(event, context):
     """Fetch usage from PromQL, update DynamoDB, check quotas, send alerts."""
     print(f"Starting quota monitoring at {datetime.now(timezone.utc).isoformat()}")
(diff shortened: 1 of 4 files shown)


wirjo (approved):
✅ **Approved — Low risk, well-tested fix**

**What it fixes:** Idle users with stale `daily_date` kept triggering false 'Daily Token Quota EXCEEDED' alerts and showing inflated usage in `ccwb quota usage`. The daily counter is only rolled over on new activity (`update_quota_metrics`), but the monitor and CLI read it verbatim without checking staleness.

**Why it's safe:**
- Mirrors the exact guard already in `quota_check/index.py` (lines 392-397): `if daily_date != current_date: daily_tokens = 0`
- Read-only fix — doesn't write to DynamoDB, only changes display/alert interpretation
- Monthly counters (`total_tokens`) correctly untouched
- Missing `daily_date` (legacy rows) correctly treated as stale (`None != '2026-07-01'`)
- CLI properly switches from naive `datetime.utcnow()` to tz-aware `datetime.now(timezone.utc)`

**Test coverage:** Excellent — 254 new lines across 2 test files covering the regression case, over-correction guard, and legacy row handling.

**No breaking changes.** DynamoDB schema unchanged, no new env vars, no API changes.

PR #730, opened 2026-07-01
fix(deploy): pass profile tags to CloudFormation stacks

lorenzstorm1:
## Why

`ccwb init` collects resource tags and stores them in `profile.tags`. `CloudFormationManager.deploy_stack()` accepts a `tags` parameter and correctly formats them for the CloudFormation API. But `_deploy_stack()` in `deploy.py` never passes them through — so all deployed stacks are missing user-configured tags.

## What

Add `tags=profile.tags if profile.tags else None` to the `manager.deploy_stack()` call.

## Test plan

- [x] Regression test: `test_deploy_tags.py` verifies tags reach `deploy_stack()`
- [x] Regression test: empty tags pass `None` (no empty dict sent to CF API)
- [x] Pre-commit hooks pass

🤖 Generated with Claude Code

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/deploy.py
+++ b/source/claude_code_with_bedrock/cli/commands/deploy.py
@@ -434,6 +434,7 @@ class DeployCommand(Command):
                         template_path=template_path,
                         parameters=boto3_params,
                         capabilities=capabilities or ["CAPABILITY_NAMED_IAM"],
+                        tags=profile.tags if profile.tags else None,
                         on_event=lambda e: progress.update(
                             task,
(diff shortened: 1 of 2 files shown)


wirjo (approved):
✅ **Approved — one-line bug fix, well-tested**

Tags configured in `ccwb init` were silently dropped during deploy. The `Profile.tags` dict was never passed through to `CloudFormationManager.deploy_stack()`. Fix adds `tags=profile.tags if profile.tags else None` to the single call site.

- Zero risk for users without tags (empty dict → None → unchanged behavior)
- Enterprise users with tag policies benefit immediately on next deploy
- No migration required
- 2 regression tests cover both positive and empty cases

PR #731, opened 2026-07-01
fix(idc): recover IDC sign-in in-session instead of failing fast

qing-at-work:
## Why

When an IDC user's SSO session expired (or on cold start with no session), Claude Code ran the credential process via the `awsAuthRefresh` hook — but the device-auth fail-fast gate keyed on `stderrIsTerminal()` alone. Because Claude Code captures the hook's stderr (a pipe, not a TTY), it **always** failed fast with "exit Claude Code and re-run the launcher", even though Claude Code streams the hook's stderr live and can surface the sign-in prompt. Result: IDC users could never complete sign-in from inside a running session — first-launch and mid-session expiry both dead-ended.

Related upstream issues: anthropics/claude-code#67529 (no recovery path for expired tokens), #67530 (re-auth without restart).

## What

- **`loginMode` gate.** `runDeviceAuthorization` now takes `loginMode`. The `--login` / `awsAuthRefresh` slot runs the device-auth poll (browser on desktop, verification URL + code on headless/SSH) instead of failing fast. The silent `awsCredentialExport` path (stderr discarded) still fails fast so it can never hang. Gate is now `stderrIsTerminal() || loginMode`.
- **165s timeout.** The `--login` timeout is capped just under Claude Code's hardcoded 180s `awsAuthRefresh` hook timeout, so our own actionable message prints before Claude Code kills the hook.
- **Plain-language messages.** Timeout and generic-failure errors lead with the takeaway on the first line (Claude Code may surface only the first stderr line) and tell the user sending another message retries.

## Testing

Verified on **desktop (macOS)** and **headless SSH** — cold start, silent token refresh, dead refresh token, timeout, and send-another-message recovery all work. Timeout path is identical on both (message at 165s, Claude Code clears its panel at 180s).

Regression tests added:
- `canSurfaceVerificationURL` truth table (silent vs login × headless vs browser)
- headless login proceeds past the gate (no longer fails fast)
- `pollForToken` timeout wraps `context.DeadlineExceeded` (the friendly-message contract)

`go test ./... -race -count=1` and `go vet ./...` pass.

## Notes

- Tier-1 credential path (`.claude/rules/review-tiers.md`): regression tests included; backward-compatible (no config/struct changes → no Go↔Python parity impact); unaffected auth types (oidc, none) unchanged.
- A companion docs PR demotes the `claude-bedrock` launcher to optional now that in-session recovery works.

### DIFF
--- a/source/go/cmd/credential-process/idc.go
+++ b/source/go/cmd/credential-process/idc.go
@@ -135,9 +135,15 @@ func (a *credentialApp) resolveIDCSettings(ctx context.Context) (*idcSettings, i
 func (a *credentialApp) runIDCLogin() int {
 	debugPrint("IDC login (sign-in only) for profile '%s'", a.profile)
 
-	// Generous timeout: the user must approve in a browser, possibly on another
-	// device, before this returns.
-	ctx, cancel := context.WithTimeout(context.Background(), 180*time.Second)
+	// Timeout must fire JUST BELOW Claude Code's hardcoded 180s awsAuthRefresh
+	// hook timeout (added in Claude Code v2.1.43, issue anthropics/claude-code
+	// #25457; not configurable). If we exceed 180s, Claude Code kills the hook
+	// with no message and the sign-in prompt "quietly disappears". 165s leaves
+	// ~15s for our own actionable timeout error to print and flush before that
+	// kill. When run directly in a terminal (not via the hook) there is no
+	// external limit, but the shorter bound is still a reasonable ceiling for a
+	// human browser approval.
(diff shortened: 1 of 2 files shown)


wirjo (approved):
LGTM — well-scoped fix with strong test coverage. The `loginMode` gate is the right approach to differentiate interactive (awsAuthRefresh) from silent (awsCredentialExport) paths. 165s timeout under Claude Code's 180s cap is pragmatic.

Minor nits (non-blocking):
- PR body references "300s/600s" but code uses 165s/180s — stale from development?
- Panic-based gate test is creative but unorthodox; a mock returning a known error would be more conventional

Safe to merge. No file overlap with #733.

PR #743, opened 2026-07-03
Fix/otel default department

adiospeds:
Why
When packaging Claude Code (ccwb package), the generated settings.json
falls back to a hardcoded set of default OTEL_RESOURCE_ATTRIBUTES whenever
the operator hasn't configured their own. That fallback shipped with
department=engineering.

engineering is not a safe default — most deployments are not exclusively
engineering teams, and the value silently mislabels OTEL telemetry and cost
attribution for any non-engineering user who never overrides it. Every other
attribute in the same default string is neutral (team.id=default,
cost_center=default, organization=default, project=default), so
department should match. Operators who do want a real department value can
still set OTEL_RESOURCE_ATTRIBUTES and that override continues to win.

What
Change the default department=engineering → department=default in the
fallback OTEL attributes.
Fix applies in both code paths that contain the duplicated default:
source/claude_code_with_bedrock/cli/commands/package.py (PackageCommand)
source/claude_code_with_bedrock/cli/commands/package_cb.py (PackageCbCommand)
Add a regression test (TestPackageCommandOtelDefaults) that:
drives _create_claude_settings with a mocked monitoring endpoint and
asserts the rendered default is department=default (and not
department=engineering);
asserts an explicit OTEL_RESOURCE_ATTRIBUTES value still overrides the
default.
Testing
cd source && poetry run pytest tests/ -q → 919 passed, 22 skipped.
New regression tests fail against the old default and pass with the fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Guidance for Claude Code and Cowork on Amazon Bedrock
 
-[![Stable Release](https://img.shields.io/github/v/release/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock?style=for-the-badge&label=stable)](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/releases/latest)
+[![Stable Release](https://img.shields.io/github/v/release/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock?style=for-the-badge&label=stable&filter=v*.*.*)](https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/releases/latest)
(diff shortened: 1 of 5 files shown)


wirjo:
## Review

**Risk:** Very Low

Changes the default OTEL resource attribute from `department=engineering` to `department=default` — a neutral baseline that avoids assuming every deployment is an engineering team. Also bumps the version badge filter and adds a regression test. Backwards compatible since the attribute was only a default (custom values still override).

**Findings:**
- None — clean change. The test coverage is solid, the version bump is appropriate, and the badge filter fix is a nice touch.

**Verdict:** ✅ Ship it

adiospeds:
This got pushed via #745 hence closing this.

PR #744, opened 2026-07-03
feat: declarative extra_files for package + distribute

adiospeds:
## Why

Admins sometimes need to ship files *on top of* the generated package —
preinstall/postinstall scripts, an OIDC signing certificate, a CA bundle, etc.
Today that means hand-editing `package.py` and the three archive builders in
`distribute.py`, which doesn't survive a rebase onto upstream, drifts across the
four hardcoded file inventories (the exact problem `distribution-manifest-parity.md`
warns about), and can't express "this file only goes to Windows".

This adds a single declarative list — `extra_files` — read by both `package`
and `distribute`. It is **additive**: it never re-declares the binaries,
`config.json`, `install.sh`, or `managed-settings.json`; those keep working
exactly as they do today. The list lives in the admin-side deployment profile,
so it never reaches the runtime `config.json` and is never synced to the Go
credential-process.

## What

- **config:** new `extra_files` field on the `Profile` dataclass (default `[]`,
  so old profiles load unchanged) plus a shared `extra_files.py` module with the
  OS-token matcher (`extra_applies_to`) and validator (`validate_extra_files`).
- **init:** wizard step to add/edit/remove entries, with round-trip save/reload
  wiring so re-running `init` shows the saved list.
- **package:** copies the full superset of extras into the build folder; fails
  fast (non-zero exit) on a missing `from` source or a `name` that collides with
  a generated artifact — a listed file is never silently dropped.
- **distribute:** a shared `_add_extra_files_to_zip` helper filters extras
  per-OS across all three archive builders (all-OS = every extra, per-OS token
  filter, landing-family filter). Distribute reads from the already-built
  package folder, never the `from` source, keeping the two commands decoupled.
- **docs:** Extra Files subsection in the CLI reference.

Each entry has three fields: `name` (path inside the package), `targets`
(`all` / family / arch / list), and `from` (source path on the admin's machine).

## Backward compatibility

- New field defaults to `[]` — old `profile.json` files load without error.
- Not synced to Go config; not written to runtime `config.json`.
- No change to existing package/distribute behavior when `extra_files` is empty.

## Testing

137 tests across `test_extra_files.py` (matcher truth table + validator),
`test_config.py` (round-trip), `test_package.py` (copy, missing source,
validation), and `test_distribute.py` (all-OS / per-OS filtering, arch-in-family).

```
137 passed
```

Path safety (`name`) rejects absolute paths, `..` traversal, Windows drive
letters, and collisions with generated artifacts — the zip-slip surface, since
`name` becomes a zip entry path.

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -257,7 +257,34 @@ poetry run ccwb package [options]
 2. Creates `config.json` with federation config read from the admin profile
 3. Creates `claude-settings/settings.json` with Bedrock model and OTel endpoint (or `managed-settings.json` if `--managed` was used during init)
 4. Creates installer scripts (`install.sh`, `install.bat`, `ccwb-install.ps1`)
-5. Outputs to `dist/{profile}/{timestamp}/`
+5. Copies any admin-defined **extra files** into the build folder (see below)
+6. Outputs to `dist/{profile}/{timestamp}/`
+
+**Extra Files:**
+
+You can ship additional files or folders on top of the generated package —
+preinstall/postinstall scripts, an OIDC signing certificate, a CA bundle, and
+so on. Configure them during `ccwb init` (the "Extra files" step); they are
+stored in your admin deployment profile, never in the runtime `config.json` and
+never delivered to end users' Claude Code config.
+
+Each entry has three fields:
+
+| Field | Meaning |
+|---|---|
+| `name` | path inside the package (relative — no `..` or absolute paths) |
+| `targets` | which machines receive it (see tokens below) |
+| `from` | source path on your machine (a file or a whole folder) |
+
+`targets` accepts `all`, a family (`macos`, `linux`, `windows`), an
+architecture (`macos-arm64`, `macos-intel`, `linux-x64`, `linux-arm64`), or a
+list such as `["macos", "linux"]`.
+
+`ccwb package` copies **all** extras into the build folder. `ccwb distribute`
+then filters them per platform — a `macos` extra lands in the macOS package but
+not the Windows one; an `all` extra lands in every package. `package` fails
+fast (non-zero exit) if a `from` source is missing or a `name` collides with a
+generated artifact, so a listed file is never silently dropped.
 
 **Build Modes:**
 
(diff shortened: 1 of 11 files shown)


adiospeds:
here is what the input looks like during init:
image

The package and distribute commands pick up the settings and copy the files accordingly.

adiospeds:
@wirjo can you pls review and merge this.

wirjo:
## Review

**Risk:** Low

Adds a declarative `extra_files` feature allowing admins to ship additional files (certs, scripts, etc.) on top of generated packages. Introduces a new `extra_files.py` module with platform-token matching and validation, wires it into `init` (interactive wizard), `package` (copy to build folder with fail-fast), and `distribute` (per-OS filtering at zip time). The field defaults to `[]` so existing profiles load unchanged — fully backwards compatible.

**Findings:**
- Excellent defensive design: zip-slip prevention, reserved-name collision checks, fail-fast on missing sources, and per-OS filtering hierarchy are all well-thought-out and thoroughly tested (185-line unit test for the matcher alone).
- Minor: `_add_extra_files_to_zip` returns early when `validate_extra_files(entries)` is truthy (i.e. has errors), which is correct but the comment says "No extras, or an invalid config" — the `not entries` check handles the "no extras" case while the `validate_extra_files(entries)` check handles errors. Logic is fine but the `or` reads slightly confusingly at first glance.
- The `from` import inside `_add_extra_files_to_zip` is fine for keeping it lazy but is repeated in `_copy_extra_files` — consider a top-level import in distribute.py if this stabilises.

**Verdict:** ✅ Ship it

wirjo:
@adiospeds — nice implementation! Could you share the use case that motivated this? Curious what extra files your team needs bundled into the distribution package (e.g., custom scripts, certificates, company-specific configs?).

adiospeds:
Thanks! Our main driver is Netskope, which we use as our corporate proxy. Two things specifically:
Preinstall script for our Netskope combined CA: our tenant's TLS-interception certificate can rotate, so this script generates/configures the combined CA bundle at install time and can fix things up if the cert changes and add correct env vars. Felt out of scope for ccwb itself, so extra_files was a clean way to hook it in without touching package.py/distribute.py.

Reset script: a cleanup script we ship alongside that fully tears down the install for a user (useful for support/reimaging flows).

Testing features ahead of upstream: this setup also lets us bundle and try out additional features with our users that aren't merged or adopted into ccwb repo immediately without forking or maintaining a patched copy of the repo. We can also customise it so it only goes to some of our beta testing users.


Beyond our case, the general motivation is that any admin could manually drop extra scripts/certs into the package folder before zipping, but it's easy to forget or drift across the three archive builders. Hardening this into the package/distribute flow itself makes it a declared, validated, repeatable part of the pipeline rather than a manual step someone has to remember.

adiospeds:
@wirjo .. Can we please merge this

PR #745, opened 2026-07-03
fix(init): let users set max_session_duration and retain it on re-run

adiospeds:
## Why

Admins reported that `max_session_duration` resets on every `ccwb init`. Re-running the wizard silently discarded any customized value, and there was no way to set the duration through the wizard in the first place — it was effectively hardcoded.

Root cause: the OIDC step in `init.py` unconditionally overwrote the value with a federation-type default (`43200` for direct STS, `28800` for Cognito) on every run. `_check_existing_deployment` restored the saved value into the config dict correctly, but this line clobbered it right after.

## What

- Replace the hardcoded assignment with a validated prompt (`3600–43200` seconds) whose default is the **existing value** when present, falling back to the federation-type recommendation only when unset.
- The save/reload round-trip already persisted the field, so retaining the value here is the only change needed.

## Tests

Added `source/tests/cli/commands/test_init_session_duration.py` (4 tests) that drive the OIDC step of `_gather_configuration`:
- a typed value is honored,
- re-running and accepting the default retains the prior value,
- with no prior value the default follows federation type (43200 direct / 28800 cognito).

The two behavioral tests fail against the old code and pass with the fix. Existing init/package suites (58 tests) still pass; `ruff check`/`format` clean.

## Notes

- `max_session_duration` already exists on both the Python `Profile` dataclass and the Go `ProfileConfig` struct — no config-sync change required.
- Tier 2 file (`init.py`); no auth-mode-specific behavior changed (OIDC/IDC/none unaffected — this path is OIDC-only and pre-existing).

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -873,7 +873,27 @@ class InitCommand(Command):
                 return None
 
             config["federation_type"] = federation_type
-            config["max_session_duration"] = 43200 if federation_type == "direct" else 28800
+
+            # Session duration — how long issued AWS credentials remain valid before
+            # re-authentication. Retain any previously configured value across re-runs;
+            # otherwise fall back to the federation-type recommended default.
+            _recommended_duration = 43200 if federation_type == "direct" else 28800
+            _existing_duration = config.get("max_session_duration")
+            _default_duration = _existing_duration if _existing_duration else _recommended_duration
+
+            console.print("\n[cyan]Session Duration[/cyan]")
+            console.print(
+                "How long issued AWS credentials stay valid before re-authentication "
+                f"(3600–43200 seconds; recommended {_recommended_duration})."
+            )
+            duration_str = questionary.text(
+                "Max session duration (seconds):",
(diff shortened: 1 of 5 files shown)


adiospeds:
@wirjo can you pls review and merge this.

wirjo:
## Review

**Risk:** Very Low

Replaces the hardcoded `max_session_duration` assignment with an interactive prompt that lets admins choose a value (3600–43200s) and retains it across re-runs. Previously, every `ccwb init` re-run silently clobbered any custom duration. Also includes the same OTel default fix from PR #743. Backwards compatible — existing configs without the prompt simply get the same federation-type default they had before.

**Findings:**
- The `validate` lambda uses a closure pattern (`lambda x: (x.isdigit() and 3600 <= int(x) <= 43200) or "error msg"`) — this works with questionary's validation but note `isdigit()` would reject negative numbers or whitespace, which is fine here since the bounds are positive.
- Good defensive fallback: `int(duration_str) if duration_str else _default_duration` handles a None/empty return from the prompt gracefully.
- Test suite is comprehensive with parametrized federation-type coverage.

**Verdict:** ✅ Ship it

PR #749, opened 2026-07-06
fix(otel): resolve OIDC provider type before --get-monitoring-token so silent bearer refresh works

adiospeds:
## Why

On **OIDC + central collector**, a user's telemetry silently stops being counted **~50 minutes after each login** while Bedrock inference keeps working. The result is that `ccwb quota usage` (and the CloudWatch `ClaudeCode` metrics) read materially **lower** than Claude Code's own `/usage`, because metric batches are dropped at the collector ALB with **401** and OTLP/HTTP exporters do not retry on 401.

This was reproduced live on a production OIDC profile (Azure, `azure_auth_mode=certificate`, keyring storage) and traced to a concrete ordering bug in `credential-process`.

## Root cause

`otel-helper` runs `credential-process --get-monitoring-token` on every OTEL export to obtain the monitoring bearer (the OIDC `id_token`). That flag handler dispatches in `main()` **before** the OIDC provider type is resolved:

- `--get-monitoring-token` (and `--get-mcp-auth-header`) dispatch early in `main()`.
- `app.providerType` / `app.redirectPort` were assigned only later, in the OIDC `run()` path.

So the silent `refresh_token` exchange on that path ran with `providerType == ""`, which caused a cascade:

1. `provider.TokenEndpointURL("", …)` returns an **empty URL** (`ConfigFor("")` is a zero-value `Config`).
2. `resolveConfidentialAuth()` is gated on `providerType == "azure"`, so with an empty type it skipped the **Azure client-assertion / certificate** entirely.
3. `RefreshTokenExchange` POSTed an unsigned `grant_type=refresh_token` to an empty URL → failed.
4. On that failure, `tryRefreshToken` called `ClearRefreshToken` — **deleting the still-valid refresh_token**.
5. The next cycle had no refresh_token at all, so the only recovery was a full interactive browser login.

Once the bearer crossed its 10-minute pre-expiry buffer (in `storage.GetMonitoringToken`), `otel-helper` emitted attribution headers **without an `authorization` header**, and the collector ALB's `jwt-validation` action returned **401 before forwarding** to the collector. No retry, no buffer → the batch is lost.

**Observed debug trace (before the fix):**

```
Debug: Found cached refresh_token, attempting token exchange...
Debug: Refresh token exchange failed: refresh token request failed: Post "": unsupported protocol scheme ""
Debug: OIDC authentication not available: unknown provider type:
```

Note `Post ""` — the empty token endpoint URL — and `unknown provider type:` (empty string) reaching the auth layer.

## The fix

1. **Resolve provider type + redirect port before the flag dispatch**, gated to real OIDC profiles (`IsSsoEnabled() && !IsIDC()`). This is important for IDC-safety: `resolveProviderType` calls `provider.Detect`, which returns `"oidc"` (→ `os.Exit(1)`) for the empty `provider_domain` typical of IDC/none profiles. Those profiles never use this OIDC refresh path, so they are explicitly skipped and continue to dispatch via their own `IsIDC()` / `!IsSsoEnabled()` branches.

2. **Only clear the stored `refresh_token` on a definitive IdP rejection** (`invalid_grant` / `invalid_token`, RFC 6749 §5.2). Transient failures — network/5xx/timeout, an empty/misconfigured token URL, or a missing client assertion — now **retain** the token so a later cycle can retry. Clearing on a transient failure was what turned a recoverable blip into a permanent silent-refresh outage until the next browser login. Applied to **both** `tryRefreshToken` and `refreshIDTokenOnly` (the MCP `--get-mcp-auth-header` path had the identical bug).

3. `RefreshTokenExchange` now returns a typed `RefreshExchangeError{Definitive}` (with `errors.As` helper `IsDefinitiveRefreshFailure`) and guards against an empty/relative token endpoint URL as defense-in-depth.

4. `resolveRedirectPort()` never returns `0`. Once this path can reach browser auth (previously it always died at `unknown provider type`), an unset port would have opened `http://localhost:0` → `ERR_UNSAFE_PORT`.

## Verification

**Live, after the fix** — the `--get-monitoring-token` refresh now reaches the **real** Azure token endpoint and classifies the response correctly (here with a deliberately-invalid token, hence `invalid_grant`; a valid refresh_token returns HTTP 200 and is rotated):

```
Debug: Found cached refresh_token, attempting token exchange...
Debug: Refresh token exchange failed (HTTP 400): {"error":"invalid_grant","error_description":"AADSTS9002313: ..."}
Debug: Refresh_token rejected by IdP (invalid_grant); clearing stored token
```

No more `Post ""` / `unknown provider type`, and a valid token is no longer wiped by a transient failure.

**Probe evidence** — a local probe ran the real `otel-helper` on a fixed interval, took the bearer it emitted, and POSTed to the real collector endpoint, recording the ALB status. `auth=False` cycles (bearer past the 10-min buffer, silent refresh unavailable) are exactly the dropped batches. `egress_ip` scrubbed:

```
time (UTC)            auth   exp_min  export  reason     drop
---------------------------------------------------------------------
2026-07-06T16:13:19Z  True   10.8     200     ok
2026-07-06T16:15:26Z  False  None     401     auth_401   <<< DROP
2026-07-06T16:17:34Z  False  None     401     auth_401   <<< DROP
   … (18 consecutive auth_401 drops over ~37 min) …
2026-07-06T16:52:46Z  False  None     401     auth_401   <<< DROP
2026-07-06T16:54:53Z  True   59.8     200     ok         (only after a fresh browser login)
---------------------------------------------------------------------
export status counts: {'200': 65, '401': 27}
DROP RATE (auth_401/total): 27/92 = 29.3%
cycles where helper emitted NO bearer: 27
```

The signature is a periodic square wave: ~50 min of `200`s after each login, then a hard wall of `auth_401` that only a new browser login clears. With the fix, silent refresh keeps the bearer alive across that boundary instead of failing and wiping the refresh_token.

## Tests

- `internal/oidc`: `RefreshTokenExchange` error classification — `invalid_grant`/`invalid_token` are definitive; 5xx, unexpected 4xx, non-JSON body, and empty/relative URL are transient. `IsDefinitiveRefreshFailure` safe for `nil` and unrelated errors.
- `cmd/credential-process`: `refresh_token` is cleared on `invalid_grant` but **retained** on a transient failure, for **both** `tryRefreshToken` and `refreshIDTokenOnly`; `resolveRedirectPort` never returns 0. Updated two existing tests that asserted the old clear-on-any-failure behavior.
- Full suite green: `go test ./... -race -count=1`, `go vet ./...`, `gofmt` clean.

## Auth-mode coverage & parity

- **OIDC**: fixed path (provider type resolved early; silent refresh works).
- **IDC / none**: explicitly skipped by the `IsSsoEnabled() && !IsIDC()` gate — no regression, no `os.Exit(1)`.
- **Python parity**: the Python provider already resolves `provider_type` in `__init__` and performs no refresh on the monitoring path (a documented Go/Python asymmetry), so no Python change is needed.

## Scope

Client-side `credential-process` (Go) only — **no CloudFormation / collector / ALB changes**. Rollout is `ccwb package` → `ccwb distribute` → re-install. Tier 1 critical files touched (`credential-process/main.go`, `internal/oidc/token.go`); regression tests included per the review-tiers policy.

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -98,6 +98,25 @@ func main() {
 		runExplain(profile, cfg)
 	}
 
+	// Resolve OIDC provider type + redirect port BEFORE the flag dispatch below.
+	// The --get-monitoring-token and --get-mcp-auth-header handlers perform a
+	// silent refresh_token exchange that depends on app.providerType (to build the
+	// token endpoint URL and pick the Azure confidential-auth path) and, on
+	// fall-through to browser auth, on app.redirectPort. Historically both fields
+	// were set only after the dispatch block (in the OIDC run() path), so those
+	// handlers ran with providerType=="" (empty token URL + no client assertion →
+	// failed exchange that wiped the refresh_token) and redirectPort==0 (browser
+	// fallback opened http://localhost:0 → ERR_UNSAFE_PORT).
+	//
+	// Gated to real OIDC profiles only: resolveProviderType calls provider.Detect,
+	// which returns "oidc" (→ os.Exit(1)) for the empty provider_domain typical of
+	// IDC/none profiles. IDC/none never use this OIDC refresh path, so skip them
+	// here and let their own dispatch branches (IsIDC / !IsSsoEnabled) handle them.
+	if cfg.IsSsoEnabled() && !cfg.IsIDC() {
+		app.providerType = resolveProviderType(cfg)
+		app.redirectPort = resolveRedirectPort(cfg)
+	}
+
 	// Flag dispatch — must run before auth-type branching so IDC users
 	// can use --get-monitoring-token, --show-tags, etc.
 	if *clearCache {
@@ -148,18 +167,15 @@ func main() {
 		exitAfterNotifications(app.runPassthrough())
 	}
 
-	// OIDC path — resolve provider type and redirect port
-	providerType := resolveProviderType(cfg)
-	redirectPort := 8400
-	if envPort := os.Getenv("REDIRECT_PORT"); envPort != "" {
-		if p, err := strconv.Atoi(envPort); err == nil && p > 0 {
-			redirectPort = p
-		}
-	} else if cfg.RedirectPort > 0 {
-		redirectPort = cfg.RedirectPort
+	// OIDC path — provider type + redirect port were already resolved above
+	// (before the flag dispatch) for this OIDC profile; resolve defensively if
+	// somehow unset.
+	if app.providerType == "" {
+		app.providerType = resolveProviderType(cfg)
 	}
-	app.providerType = providerType
-	app.redirectPort = redirectPort
+	if app.redirectPort == 0 {
+		app.redirectPort = resolveRedirectPort(cfg)
+	}
 
 	if *refreshIfNeeded {
 		if cfg.CredentialStorage != "session" {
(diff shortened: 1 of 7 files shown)


adiospeds:
@wirjo can you please merge this.

PR #750, opened 2026-07-06
feat(package): support offline/air-gapped Go and OTEL collector builds

wantmys2000:
## Why

Admins packaging from air-gapped/offline machines cannot complete `ccwb package` in sidecar mode. A `vendor/` directory in `source/go` fixes the credential-process/otel-helper builds, but the OTEL Collector sidecar build cannot use one: `_build_otelcol` downloads the OCB binary from GitHub, and OCB generates a **fresh Go module in a temp directory on every run**, resolving its dependencies from the network (including an internal `go mod tidy`). There is nowhere for vendored deps to live.

## What

Adds `scripts/prepare-offline-go-bundle.sh` with two modes:

- **`prepare`** (on an internet-connected machine): downloads the OCB binary pinned in `package.py` (version parsed from the source so the bundle can't drift), seeds a fresh `GOMODCACHE` covering both `source/go` and the OCB-generated collector module, then **self-verifies with a full `GOPROXY=off` rehearsal build** before producing a tarball (~250 MB).
- **`install`** (on the offline machine): copies the OCB binary to `~/.cache/ocb/` (where `package.py` checks before downloading) and emits an `offline-env.sh` setting `GOMODCACHE`, `GOPROXY=off`, `GOSUMDB=off`, `GOTOOLCHAIN=local`.

Because every `go`/`ocb` subprocess in `package.py` inherits the parent environment, **no changes to ccwb itself are required** — the seeded cache satisfies all module resolution offline.

Also fills the empty **Offline Packaging** stub heading in `assets/docs/CLI_REFERENCE.md`.

## Testing

Verified end-to-end on Linux/amd64 with Go 1.25.5:
- `prepare` produced a bundle whose built-in offline rehearsal (OCB source generation + `go mod download` + compile, all with `GOPROXY=off`) passed.
- With the seeded cache and `GOPROXY=off GOSUMDB=off`, cross-compiled the collector for linux/amd64, darwin/arm64, and windows/amd64, and built `credential-process` + `otel-helper` from `source/go` — all without network access to module proxies.
- `install` correctly seeds `~/.cache/ocb/` and warns on Go version mismatch between prepare/install machines.

No regression test: this is additive admin tooling (Tier 3, no Python/Go code paths changed); the script self-verifies at prepare time.

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -278,6 +278,21 @@ With `--go`, all 5 platforms are always available regardless of the admin's OS. 
 
 **Offline Packaging:**
 
+`ccwb package` normally needs internet on the admin machine for three things: Go module downloads for `source/go`, the OCB (OpenTelemetry Collector Builder) binary from GitHub, and Go module downloads for the OCB-generated collector module (sidecar mode only). A `vendor/` directory cannot cover the collector build — OCB generates a fresh Go module in a temp directory on every run.
+
+For air-gapped environments, use `scripts/prepare-offline-go-bundle.sh` to pre-seed everything:
+
+```bash
+# On an internet-connected machine (same OS/arch and Go version as the offline box):
+./scripts/prepare-offline-go-bundle.sh prepare
+# Transfer ccwb-offline-go-bundle.tar.gz to the offline machine, extract, then:
+./scripts/prepare-offline-go-bundle.sh install <bundle-dir>
+source <bundle-dir>/offline-env.sh   # sets GOMODCACHE, GOPROXY=off, GOSUMDB=off, GOTOOLCHAIN=local
+poetry run ccwb package --go
+```
+
+The bundle contains the pinned OCB binary (installed to `~/.cache/ocb/`, where `package.py` looks before downloading) and a pre-populated Go module cache covering both `source/go` and the collector. Because every `go`/`ocb` subprocess inherits the environment, `GOPROXY=off` plus the seeded `GOMODCACHE` satisfies all module resolution — including the `go mod tidy` OCB runs internally — with no changes to `ccwb` itself. The `prepare` step rehearses a fully offline collector build before archiving, so a bundle that ships is a bundle that works.
+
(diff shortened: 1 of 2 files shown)


wirjo (dismissed):
Solid implementation for offline/air-gapped Go and OTEL collector builds. The approach is clean and well-reasoned:

- Correctly identifies the core problem (OCB generates a fresh Go module in a temp dir, making vendor/ impossible)
- Solves it with pre-seeded GOMODCACHE + GOPROXY=off — leveraging Go's existing offline capabilities
- OCB binary path in the script exactly matches package.py's cache check (Path.home() / '.cache/ocb/ocb_{ver}_{os}_{arch}')
- Self-verifying prepare step (offline rehearsal build) ensures the bundle works before shipping
- Zero changes to ccwb itself — fully additive, backwards compatible
- OCB_VERSION is parsed from package.py source, preventing version drift
- Go version mismatch warning in install mode is a nice safety net
- set -euo pipefail for robust error handling

Minor notes (non-blocking):
- No checksum verification of the downloaded OCB binary (curl -fsSL only), but this is the same trust model as package.py itself
- ~250MB bundle is expected given Go module cache size

All existing tests pass (305 passed). No regression risk.

wirjo:
Nice work — clean approach using Go's native offline mechanisms. Approved.

**Suggestion for discoverability:** Consider exposing this as `ccwb package --prepare-offline` rather than (or in addition to) a standalone script. That way:

1. Admins discover it naturally via `ccwb package --help`
2. When `ccwb package` fails due to a network error, the error message can say: *"For air-gapped environments, run `ccwb package --prepare-offline` first on a connected machine"*
3. The two-step workflow becomes: `ccwb package --prepare-offline` (connected) → transfer bundle → `ccwb package` (air-gapped, auto-detects bundle)

The standalone script is fine as-is, but a flag on the existing command makes the feature more discoverable. Non-blocking — happy to merge either way.

wirjo:
Superseded by #753 which exposes this as `ccwb package --prepare-offline` for discoverability via `--help`. Your script is included unchanged — thank you for the contribution @wantmys2000! 🙏

PR #751, opened 2026-07-07
feat(idc-landing-page): integrate IAM Identity Center Self-Service Portal into ccwb CLI

duttaabh:
## Motivation / Summary

This came out of a customer conversation. Their existing distribution model (Presigned S3 / Authenticated Landing Page, both static installer downloads) has a scaling problem for two reasons:

1. **Contractors and unmanaged devices.** Some of their Claude Code users are contractors on devices that aren't on the corporate network and aren't MDM-managed. Static installer distribution works, but it's a one-time push — there's no channel back to those devices afterward.

2. **Every admin change requires a re-push.** With a static installer/profile, changing a model allowlist, a tool policy, or an MCP server means regenerating the package and getting every user to re-download and reinstall it. That's fine for one-off setup, but for a change that should take effect org-wide (a new MCP server, a tightened tool policy, a model rotation), it doesn't scale past a handful of users — and it scales even worse when contractors/unmanaged devices can't be reached via MDM push in the first place.

The customer pointed out that Anthropic's Enterprise Console gives them this kind of centrally-managed, auto-propagating configuration out of the box, and the lack of an equivalent was becoming a real blocker for adopting Claude via Bedrock over Claude direct.

This repo already had two of the three pieces needed to solve this, just never connected:

- A bootstrap server for dynamic config delivery (`deployment/infrastructure/bootstrap-device-code.yaml` / `bootstrap-oidc-bearer.yaml`, documented in `assets/docs/PLUGINS.md`) — but it was wired only for external OIDC providers (Okta/Azure/Auth0/Google/Cognito).
- IAM Identity Center as a first-class `ccwb` auth type — but with no distribution mechanism of its own.

This PR closes that gap: an **IAM Identity Center Self-Service Portal** — an internet-reachable landing page (optionally CloudFront-fronted, so it works for contractor devices with no VPN) that authenticates via IAM Identity Center and delivers Claude Desktop configs which self-update via the same bootstrap mechanism. Admin changes to models, tool policies, MCP servers, or command permissions propagate to every device automatically (~30 min poll interval) with no reinstall. An optional **admin console** provides the management UI for all of it.

**Relationship to #752:** #752 (add IAM Identity Center auth to the existing landing page) merged into `beta` first. This branch was reconciled onto `beta` and adopts its vocabulary and CloudFormation approach, while contributing the fuller implementation on top: the admin console, bootstrap-driven self-update, CloudFront/internal-ALB support, and command-level permissions.

Key architectural decisions:

[...]

PR #752, opened 2026-07-07
feat(distribution): add IAM Identity Center auth to existing landing page

wirjo:
Extends the existing landing page with optional IDC/SAML auth — no new stack needed.

## What

When `AuthType=idc`, deploys:
- Cognito User Pool with SAML provider (IAM Identity Center)
- ALB uses `authenticate-cognito` action instead of `authenticate-oidc`

Existing OIDC flow unchanged (default).

## Why

Consolidates IDC support into the existing template rather than maintaining a separate landing page (as proposed in #751). One landing page, different auth backends.

## Backwards compatible

- Default `AuthType=oidc` — zero change for existing deployments
- New Cognito resources are conditional (`IsIdcAuth`)
- Existing OIDC listener rules unaffected

Extracted from #751. Credit to @duttaabh for the IDC integration design.

Co-authored-by: duttaabh <duttaabh@users.noreply.github.com>

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -283,7 +283,8 @@
         "filename": "deployment/infrastructure/landing-page-distribution.yaml",
         "hashed_secret": "5ffe533b830f08a0326348a9160afafc8ada44db",
         "is_verified": false,
-        "line_number": 1070
+        "line_number": 1070,
+        "is_secret": false
       }
     ],
     "source/claude_code_with_bedrock/cli/utils/cowork_3p.py": [
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -183,7 +183,7 @@
         "filename": "deployment/infrastructure/bootstrap-device-code.yaml",
         "hashed_secret": "b63bf00edb07af6ffba7f7ceb7ed573a913271f7",
         "is_verified": false,
-        "line_number": 172,
+        "line_number": 177,
         "is_secret": false
       }
     ],
@@ -277,13 +277,22 @@
         "is_secret": false
       }
     ],
+    "deployment/infrastructure/landing-page-distribution.yaml": [
+      {
+        "type": "Secret Keyword",
+        "filename": "deployment/infrastructure/landing-page-distribution.yaml",
+        "hashed_secret": "5ffe533b830f08a0326348a9160afafc8ada44db",
+        "is_verified": false,
+        "line_number": 1070
+      }
+    ],
     "source/claude_code_with_bedrock/cli/utils/cowork_3p.py": [
       {
         "type": "Base64 High Entropy String",
         "filename": "source/claude_code_with_bedrock/cli/utils/cowork_3p.py",
         "hashed_secret": "a0422442dba3fda71eda7b743f0e726ad8af17a5",
         "is_verified": false,
-        "line_number": 345,
+        "line_number": 596,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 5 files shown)


wirjo:
## Review

**Risk:** Low
Adds IAM Identity Center (SAML) auth as an alternative to the existing OIDC flow. Default is `oidc`, so existing deployments are unaffected.

**Findings:**
- ✅ **Backwards compatibility** — `AuthType` defaults to `oidc`; the original `HTTPSListener` is conditioned on `IsOidcAuth` which is true by default, preserving the existing OIDC path exactly.
- ✅ **CFN quality** — All new IDC resources (`IdcUserPool`, `IdcSamlProvider`, `IdcUserPoolClient`, `IdcUserPoolDomain`, `IdcHTTPSListener`, `IdcLoginUrl` output) are properly conditioned on `IsIdcAuth`. The reworked `IsCognito` condition (`IsCognitoProvider AND IsOidcAuth`) correctly prevents the OIDC-Cognito listener config from conflicting with the IDC Cognito pool.
- ✅ **Security** — ALB `authenticate-cognito` action uses TLS 1.3 policy, 12h session timeout, `OnUnauthenticatedRequest: authenticate` (forces login). SAML metadata consumed via HTTPS URL. `GenerateSecret: true` on the app client is correct for ALB-backed Cognito auth.
- ⚠️ **Minor: `SamlMetadataUrl` not validated in CFN** — The parameter has no `AllowedPattern` or constraint; if someone deploys with `AuthType=idc` but forgets the URL, the Cognito IdP resource will fail at create-time with a less-than-helpful error. Consider adding a condition or `MinLength` constraint on the parameter (low severity — deploy will still fail-safe).
- ⚠️ **Minor: Cognito domain collision risk** — `Domain: !Sub '${AWS::StackName}-idc'` could collide if the stack name is long (max 63 chars for Cognito domain prefix). Consider truncation or a hash suffix (cosmetic — CFN will error clearly if too long).
- ℹ️ **`effective_auth_type` referenced in `deploy.py`** — not defined in this diff. Assuming it's defined elsewhere on the Profile model (if not, this would be a runtime error).

**Verdict:** ✅ Ship it — clean conditional architecture, no regressions for existing OIDC users, minor hardening suggestions above are non-blocking.

scouturier:
### Code review

Thanks for splitting this out of #751 — the OIDC/IDC separation reads cleanly. One issue worth addressing before merge:

1. The IDC landing-page deploy path passes `IdPProvider=None` to CloudFormation. The IDC init branch writes the distribution config without an `idp_provider` key, so `profile.distribution_idp_provider` stays at its `None` default:

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/b9336086c4cbf360ee2a90af272d782806dc3fce/source/claude_code_with_bedrock/cli/commands/init.py#L1768-L1774

The landing-page params block is reached for any `distribution_type == "landing-page"` (no auth-type guard) and unconditionally appends the value, producing the literal string `IdPProvider=None`:

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/b9336086c4cbf360ee2a90af272d782806dc3fce/source/claude_code_with_bedrock/cli/commands/deploy.py#L1005-L1007

That parameter has `AllowedValues: [okta, azure, auth0, cognito, generic]` and no `Default`, so `None` fails template validation and the IDC deploy aborts before creating resources:

https://github.com/aws-solutions-library-samples/guidance-for-claude-code-with-amazon-bedrock/blob/b9336086c4cbf360ee2a90af272d782806dc3fce/deployment/infrastructure/landing-page-distribution.yaml#L56-L60

wirjo:
Good catch @scouturier — fixed in #794 (merged to beta):

1. `deploy.py`: `IdPProvider` is now only appended when `distribution_idp_provider` is not None (both real deploy and print-cmd paths)
2. `landing-page-distribution.yaml`: Added `Default: ""` and `""` to `AllowedValues` so IDC deploys pass validation without an IdP

The IDC path no longer hits the CFN `AllowedValues` rejection.

PR #754, opened 2026-07-07
fix(quota): aggregate token usage with sum_over_time, not increase (delta metric)

adiospeds:
## Description

The quota-monitor Lambda (`claude-code-quota-monitor`) aggregates per-user token usage every 15 minutes and writes it to the `UserQuotaMetrics` DynamoDB table that `ccwb quota usage` reads and that monthly/daily quota enforcement is based on. It was querying CloudWatch's PromQL-compatible API with `increase()` over `claude_code.token.usage`. This PR switches the aggregation to `sum_over_time()`.

## Why is this change needed

`claude_code.token.usage` is an OpenTelemetry **Counter** that Claude Code exports with **DELTA temporality by default** (`OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` — see [Claude Code monitoring docs](https://code.claude.com/docs/en/monitoring-usage)). With delta temporality, each exported datapoint is the tokens emitted **since the last export**, so the series steps up **and** down (a sawtooth) — it is *not* a cumulative running total.

`increase()` is only valid for series with **cumulative** temporality (a monotonically-increasing counter that resets to 0 on process restart). Fed a delta series, `increase()` interprets every down-step as a counter reset and applies bogus "reset compensation," so it returns **empty results or wildly understated values** depending on the exact arrangement of samples in the window. CloudWatch's PromQL API even returns an advisory warning for this exact case:

```
PromQL info: metric might not be a counter, name does not end in _total/_sum/_count/_bucket: "claude_code.token.usage"
```

### User-visible impact

When `increase()` returned empty/zero for a user in a given 15-minute cycle, `update_quota_metrics()` hit its `if delta <= 0: continue` guard and **skipped the DynamoDB write entirely**. That had two effects:

1. **`total_tokens` (monthly) froze** and was systematically undercounted — so monthly quota enforcement was too lax.
2. **`daily_date` was never advanced to the current day**, so the read path's stale-day guard (in both `quota.py` and the Lambda's `_build_usage_entry`) correctly zeroed the daily counter — surfacing as **`Daily Tokens: 0`** in `ccwb quota usage` even for an actively-used account.

Because `increase()`'s output depends on the shape of samples in the sliding window, the command's numbers appeared to update **intermittently** (a cycle that happened to land on a rising window would write a small value and refresh `daily_date`; the next cycle would write nothing) — which made the bug look flaky rather than systematic.

Meanwhile the other consumers were **correct the whole time**, because they sum the per-interval delta values directly:
- **Grafana dashboards** (via Athena — the OTEL logs exported to S3, `SUM(...)`)
- **CloudWatch metric** queried with the `Sum` statistic
- Claude Code's own `/usage`

Only the quota Lambda tried to reconstruct a total with `increase()`, and that was the sole broken link.

Fixes # 

## Type of Change

## Changes Made

- `deployment/infrastructure/lambda-functions/quota_monitor/index.py`
  - Replaced `increase(...)` with `sum_over_time(...)` in all three `claude_code.token.usage` queries (per-user total; type+model breakdown for cost; type-only breakdown for input/output/cache).
  - Replaced `increase(...)` with `sum_over_time(...)` in both `ClaudeCoWork` `token.usage.input/output` queries (also delta, MetricFilter-derived per-event counts).
  - Added a docstring/comment explaining the delta-vs-cumulative temporality reasoning and the coupling caveat (see below).
- `source/tests/test_quota_monitor_lambda.py`
  - Added `TestPromQLAggregationFunction` with a regression test that monkeypatches `_promql_query` to capture the generated query strings and asserts `sum_over_time(` (never `increase(`) for both the Claude Code and CoWork metrics, plus a test that a non-empty vector flows through the aggregation.

## Additional Notes

**Temporality coupling (documented in-code):** `sum_over_time()` is correct **as long as** the metric is exported with delta temporality (Claude Code's default). If a deployment sets `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=cumulative`, then `increase()` would become the correct function instead. This is noted in the function docstring so the two never drift.

**Window-boundary tiling:** `sum_over_time([900s])` sums every datapoint in the trailing 900s window, and the EventBridge schedule is also `rate(15 minutes)`. Scheduler jitter (~sub-second) could in theory double-count or miss a datapoint at a window boundary; this is negligible (well under 1%) compared to the ~10x undercount being fixed. A follow-up could pin the query to a fixed grid via the already-present-but-unused `time_param` argument on `_promql_query()` if exact tiling is ever required.

**Existing data:** month-to-date `total_tokens` for existing deployments was undercounted while `increase()` was in effect. This PR fixes forward (correct going forward). Operators who want the historical month corrected can backfill the `MONTH#<month>` rows from Athena/CloudWatch (`SUM`/`sum_over_time` over the month) — out of scope for this PR.

**Diagnosis method (reproducible):** against a live account, the same 900s window returned an empty/tiny vector for `increase()` while `sum_over_time()` returned the real total that matched Grafana/CloudWatch. Anyone can reproduce with a SigV4-signed POST to `https://monitoring.<region>.amazonaws.com/api/v1/query`:

```bash
# understated / empty (old behavior):
query=sum by ("user.email")(increase({"claude_code.token.usage"}[900s]))
# correct (new behavior):
query=sum by ("user.email")(sum_over_time({"claude_code.token.usage"}[900s]))
```

### What was tested

- [x] Unit/integration tests pass

### Special cases

- [x] n/a

## Testing Evidence

**Test Environment:** Python 3.12, macOS; PromQL diagnosis run against `us-east-1`.

**Test Results:**

Red/green verified — the new regression tests fail against the old `increase()` code and pass with the fix:

```text
# Against the ORIGINAL (increase()) code — new tests correctly FAIL:
FAILED tests/test_quota_monitor_lambda.py::TestPromQLAggregationFunction::test_claude_code_queries_use_sum_over_time_not_increase
FAILED tests/test_quota_monitor_lambda.py::TestPromQLAggregationFunction::test_cowork_queries_use_sum_over_time_not_increase
AssertionError: CoWork delta metric must use sum_over_time:
  sum by ("user_email", "model")(increase({"ClaudeCoWork","token.usage.input"}[900s]))

# With the fix applied — all pass:
$ poetry run pytest tests/test_quota_monitor_lambda.py -q
.......                                                                  [100%]
7 passed in 0.27s

$ ruff check tests/test_quota_monitor_lambda.py
All checks passed!
$ ruff format --check tests/test_quota_monitor_lambda.py
1 file already formatted
```

### DIFF
--- a/deployment/infrastructure/lambda-functions/quota_monitor/index.py
+++ b/deployment/infrastructure/lambda-functions/quota_monitor/index.py
@@ -81,22 +81,38 @@ TOKEN_TYPE_TO_RATE_KEY = {
 
 
 def fetch_usage_from_promql():
-    """Query PromQL for per-user token usage in the last aggregation window only."""
+    """Query PromQL for per-user token usage in the last aggregation window only.
+
+    Aggregation MUST use sum_over_time(), NOT increase(). Claude Code exports
+    ``claude_code.token.usage`` as an OpenTelemetry Counter with DELTA temporality
+    by default (OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta): each
+    datapoint is the tokens emitted since the last export, so the series steps up
+    AND down (a sawtooth). increase() assumes CUMULATIVE temporality (a monotonic
+    running total) and reads every down-step as a counter reset — it returns empty
+    or wildly understated results, which froze DynamoDB and surfaced as
+    "Daily Tokens: 0" in `ccwb quota usage`. sum_over_time() sums the per-interval
+    deltas in the window = tokens used in the last 15 minutes, matching how the
+    Athena/CloudWatch consumers compute usage.
+
+    Coupling: this is correct only while the metric is exported with delta
+    temporality. If a deployment sets the temporality preference to `cumulative`,
+    increase() would become the correct function instead.
+    """
     window = AGGREGATION_WINDOW
 
     # Delta tokens per user in the last window
     results = _promql_query(
-        f'sum by ("user.email")(increase({{"claude_code.token.usage"}}[{window}s]))'
+        f'sum by ("user.email")(sum_over_time({{"claude_code.token.usage"}}[{window}s]))'
     )
 
     # Delta token type AND model breakdown per user (for cost calculation)
     type_model_results = _promql_query(
-        f'sum by ("user.email", type, model)(increase({{"claude_code.token.usage"}}[{window}s]))'
+        f'sum by ("user.email", type, model)(sum_over_time({{"claude_code.token.usage"}}[{window}s]))'
     )
 
     # Delta token type breakdown per user (without model, for backward compat)
     type_results = _promql_query(
-        f'sum by ("user.email", type)(increase({{"claude_code.token.usage"}}[{window}s]))'
+        f'sum by ("user.email", type)(sum_over_time({{"claude_code.token.usage"}}[{window}s]))'
     )
 
     users = {}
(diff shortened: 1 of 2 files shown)


adiospeds:
@wirjo can you review and merge this - it seems to be a major bug for all token quota based users.

wirjo (approved):
✅ **Approved — valid bug, correct fix**

**Code path traced:**
1. EventBridge fires every 15min → `fetch_usage_from_promql()` queries CloudWatch Prometheus API
2. `claude_code.token.usage` is an OTel Counter with **DELTA temporality** (each point = tokens-since-last-export, sawtooth pattern)
3. `increase()` assumes CUMULATIVE (monotonically-increasing) — interprets the sawtooth down-steps as counter resets → returns near-zero or empty
4. `fetch_usage_from_promql()` returns ~0 → `update_quota_metrics()` hits `if delta <= 0: continue` → skips DDB write → total_tokens freezes
5. `sum_over_time()` simply sums all datapoints in the window — correct for delta-temporality series

Fix is mechanical (5 string replacements), well-documented (docstring explains temporality coupling), and regression-tested (3 new tests verify query construction). Backward compatible — return shape unchanged, only the values are now correct instead of understated.

PR #757, opened 2026-07-07
feat(auth): attach existing IAM managed policies to the federated role

adiospeds:
## Why

Enterprises deploying this guidance over the internet (no PrivateLink/VPN requirement) frequently need network-level guardrails on the Bedrock federated role. The canonical example: a customer-managed IP-restriction policy that denies all actions unless `aws:SourceIp` falls within corporate CIDR ranges, so leaked or replayed federated credentials are useless outside the corporate network.

Today the auth stacks hardcode the federated role's `ManagedPolicyArns` to the single stack-created `BedrockAccessPolicy`, so the only way to add such a guardrail is attaching the policy manually in the IAM console. That has real operational costs:

- **Configuration drift** — the attachment lives outside CloudFormation, invisible to stack updates and drift detection.
- **Not repeatable** — every account/environment needs the same manual step, and nothing enforces it during rollout.
- **Fragile** — if the role is ever replaced (rename, stack recreation), the attachment is silently lost and Bedrock access is un-guarded until someone notices.
- **Blocks clean teardown** — out-of-band attachments can interfere with stack deletion.

## What

Admins can now supply existing customer-managed policy ARNs during `ccwb init`; `ccwb deploy` forwards them to the auth stack, which attaches them alongside the stack-created Bedrock access policy — fully managed by CloudFormation.

- **Templates** (all 7 `bedrock-auth-*.yaml`): new `AdditionalManagedPolicyArns` parameter (`Type: String`, `Default: ''`, ARN-list `AllowedPattern`) and a positive `HasAdditionalManagedPolicies` condition. Each federated role (`DirectIAMRole`, `CognitoAuthenticatedRole`, `BedrockIDCRole`) builds `ManagedPolicyArns` via `!If` — base policy only when empty, base + supplied ARNs when set. `CognitoUnauthenticatedRole` is deliberately untouched.
- **Init wizard**: optional prompt in both the OIDC and IDC flows (comma-separated ARNs, validated per entry, cancel/non-interactive safe). Field is persisted to the profile and restored on re-run (`_save_configuration` ↔ `_check_existing_deployment` parity).
- **Deploy**: parameter is appended only when configured — omitted otherwise so the template default keeps existing behavior. Covered in the real deploy path (IDC + OIDC) and the `--show-commands` dry-run path.
- **Config**: `additional_managed_policy_arns: list[str] = []` on the Python `Profile`, mirrored as `AdditionalManagedPolicyARNs` in the Go `ProfileConfig` (deploy-time only; not added to `--explain` since it never affects resolved runtime config).

Example usage during init:

```
Additional managed policy ARNs (comma-separated, optional): arn:aws:iam::123456789012:policy/corp-ip-restriction
```

## Auth mode coverage

| Mode | Behavior |
|------|----------|
| OIDC (all providers) | Parameter wired into `DirectIAMRole` and `CognitoAuthenticatedRole` |
| IDC | Parameter wired into `BedrockIDCRole` |
| none | No federated-role stack deployed — naturally a no-op |

## Backwards compatibility

- Parameter defaults to `''` — existing stacks update in place with no change.
- Profiles saved before this field existed load with an empty list (regression-tested).
- No new required CLI arguments; non-interactive runs are unaffected.

## Testing

- `poetry run pytest tests/ -q` — 1706 passed, 28 skipped
- `cfn-lint deployment/infrastructure/bedrock-auth-*.yaml` — clean
- `ruff check` / `ruff format --check` — clean
- New tests:
  - `test_cloudformation_additional_policies.py` — parameter, condition, and `ManagedPolicyArns` wiring asserted across all 7 templates, including that the unauthenticated Cognito role is not extended
  - `test_deploy_additional_policies.py` — exercises the real `_deploy_stack` auth path (OIDC + IDC): parameter present when configured, absent when empty
  - `test_init_additional_policies_roundtrip.py` — init re-run round-trip, old-config backward compat, serialization round-trip, and ARN-list validator cases (GovCloud partition, AWS-managed policies, path-prefixed policies, invalid entries)
- Dry-run verified manually: `--show-commands` output includes `AdditionalManagedPolicyArns=...` only when the profile sets it

## Notes for reviewers

- IAM allows 10 managed policies per role by default (quota-raisable to 20); documented in `CLI_REFERENCE.md`.
- The supplied policies must already exist in the deployment account — CloudFormation fails the stack operation with a clear error if an ARN doesn't resolve, rather than silently ignoring it.

### DIFF
--- a/assets/docs/CLI_REFERENCE.md
+++ b/assets/docs/CLI_REFERENCE.md
@@ -85,6 +85,10 @@ poetry run ccwb init [options]
 - Prompts for authentication method selection:
   - Direct IAM: Uses IAM OIDC Provider for federation
   - Cognito: Uses Cognito Identity Pool for federation
+- Prompts for additional IAM managed policy ARNs to attach to the federated role (optional):
+  - Attach existing customer-managed guardrail policies (for example, an IP-restriction policy that denies requests unless `aws:SourceIp` matches your corporate CIDR ranges)
+  - Policies must already exist in the deployment account; they are attached alongside the stack-created Bedrock access policy
+  - IAM allows at most 10 managed policies per role by default (quota increase possible up to 20)
 - Configures AWS settings (region, stack names)
 - Prompts for Claude model selection (Opus, Sonnet, Haiku)
 - Configures cross-region inference profiles (US, Europe, APAC)
--- a/assets/docs/DEPLOYMENT.md
+++ b/assets/docs/DEPLOYMENT.md
@@ -36,6 +36,8 @@ The `ccwb` (Claude Code with Bedrock) CLI tool guides you through deployment wit
 
 The wizard asks you to choose an authentication method. You can select either Direct IAM federation or Cognito Identity Pool based on your organization's requirements. Both methods provide secure OIDC federation to AWS credentials.
 
+The wizard also lets you attach additional IAM managed policies to the federated role. If your organization requires extra guardrails — for example, an IP-restriction policy that denies access unless `aws:SourceIp` matches your corporate network ranges — create that policy in the deployment account first, then supply its ARN (comma-separate multiple ARNs) when prompted. The deploy step attaches the policies alongside the generated Bedrock access policy, so they stay managed by CloudFormation instead of manual console edits.
+
 Next, you'll select your Claude model and configure regional access. Choose from available Claude models (Opus, Sonnet, Haiku) and select a cross-region inference profile (US, Europe, or APAC) for optimal performance. The wizard will then prompt you to select a source region within your chosen profile for model inference. Finally, choose where to deploy the authentication infrastructure (typically your primary AWS region) and configure optional monitoring setup, which provides usage analytics and cost tracking through OpenTelemetry.
 
 Once configuration is complete, deploy the infrastructure with:
--- a/deployment/infrastructure/bedrock-auth-auth0.yaml
+++ b/deployment/infrastructure/bedrock-auth-auth0.yaml
@@ -57,7 +57,15 @@ Parameters:
       - 'false'
     Description: Enable CloudWatch monitoring for Bedrock usage
 
+  AdditionalManagedPolicyArns:
+    Type: String
+    Default: ''
+    Description: Comma-separated ARNs of existing IAM managed policies to attach to the federated role (e.g. an IP-restriction policy). Leave empty to skip.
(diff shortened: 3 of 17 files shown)


wirjo:
Thanks @adiospeds — this is a really thorough implementation and the use case is clearly valid. The code quality is excellent (well-tested, covers all auth modes, backward-compatible defaults).

However, after discussing internally we'd prefer a **docs-only approach** for this use case. Here's our reasoning:

### Why docs-only

1. **Maintenance cost** — touching all 7 auth templates + init wizard + deploy + config + Go struct creates significant surface area for a feature that's relevant to a small subset of enterprises.

2. **Right layer for enforcement** — enterprises that need IP-restriction policies already have governance tooling (AWS Config rules, SCPs, or their own IaC layer wrapping CCWB). They wouldn't rely on `ccwb init` to enforce security policy — that's what their compliance automation does.

3. **Drift is a governance problem, not a deploy problem** — if an admin is sophisticated enough to need IP-restriction policies, they're sophisticated enough to use AWS Config (`iam-policy-in-use`) or an SCP to enforce the attachment persists.

### What we'd accept instead

A ~10-line addition to `DEPLOYMENT.md` documenting:

```bash
# Attach additional managed policies to the federated role post-deploy:
aws iam attach-role-policy \
  --role-name $(aws cloudformation describe-stacks \
      --stack-name <profile>-auth \
      --query 'Stacks[0].Outputs[?OutputKey==`FederatedRoleName`].OutputValue' \
      --output text) \
  --policy-arn arn:aws:iam::123456789012:policy/corp-ip-restriction
```

Plus a note that:
- IAM allows up to 10 managed policies per role (quota-raisable to 20)
- Enterprises should enforce attachment via AWS Config or SCPs for drift protection
- If the auth stack is recreated, re-attach after deploy

Would you be open to converting this to a docs-only PR? We'd love to merge that quickly. The implementation work you did here is genuinely impressive — it's just a case where the simpler solution better fits the project's complexity budget.

adiospeds:
Hi @wirjo - can you review and merge this please

adiospeds:
fair say on the cost of change - and I can create a doc and for my co admins so they dont miss the changes. Thanks for the review. Can you also check #758 - I think that is a critical one for us to change manually - If it looks good I will start working on it - Closing this one - wlll create another PR with doc change.

PR #762, opened 2026-07-08
fix(quota): enforce quota before STS exchange on silent refresh paths

### LINKED ISSUE #761: Bug: Go credential-process can return AWS credentials before quota enforcement completes on silent refresh paths
## Summary

The Go `credential-process` can issue AWS credentials on the silent refresh and refresh-token paths without a successful quota check.

There are two related problems:

1. `trySilentRefresh()` and `tryRefreshToken()` exchange OIDC tokens for AWS credentials and save those credentials before `run()` performs quota enforcement.
2. `run()` only calls the quota API if `storage.GetMonitoringToken(...)` returns a token. If no monitoring token is available, the quota check is skipped and credentials are still returned.

This can allow over-quota users to continue receiving Direct STS credentials even when a quota API endpoint is configured.

## Environment

- Auth: OIDC provider
- Federation: Direct STS (`AssumeRoleWithWebIdentity`)
- Credential helper: Go `credential-process`
- Credential storage: keyring or session storage
- Quota monitoring: enabled
- Enforcement: `block`
- Silent refresh / refresh-token flow available

This is not provider-specific and does not require account IDs or user identifiers to describe.

## Steps to reproduce

One way to exercise the issue:

1. Configure quota enforcement with a quota API endpoint.
2. Exceed the user's quota so the quota API would return `allowed: false`.
3. Ensure there is no valid cached monitoring token for the profile, while a refresh path can still obtain AWS credentials.
4. Force the cached AWS credentials to expire or reduce `max_session_duration` for testing.
5. Run:

   ```bash
   credential-process --profile <profile>
   ```

6. Observe whether credentials are returned without the quota API being called or without a blocking response being enforced.

## Expected behavior

When `quota_api_endpoint` is configured, `credential-process` should not print AWS credential JSON unless a quota check has completed successfully and returned `allowed: true`.

If the helper cannot obtain a token that the quota API accepts, it should respect `quota_fail_mode` or fail closed for blocking enforcement configurations. It should not silently skip quota enforcement and return credentials.

## Actual behavior

On the silent refresh path:

```go
if creds := a.trySilentRefresh(); creds != nil {
    if a.cfg.QuotaAPIEndpoint != "" {
        token, _ := storage.GetMonitoringToken(a.profile, a.cfg.CredentialStorage)
        if token != "" {
            qr := quota.Check(...)
            ...
        }
    }
    outputJSON(creds)
    return 0
}
```

On the refresh-token path:

```go
if creds := a.tryRefreshToken(); creds != nil {
    if a.cfg.QuotaAPIEndpoint != "" {
        token, _ := storage.GetMonitoringToken(a.profile, a.cfg.CredentialStorage)
        if token != "" {
            qr := quota.Check(...)
            ...
        }
    }
    outputJSON(creds)
    return 0
}
```

If `token == ""`, the quota check block is skipped and `outputJSON(creds)` still runs.

Separately, `trySilentRefresh()` and `tryRefreshToken()` already call `getAWSCredentials()` and save the resulting AWS credentials before the quota check in `run()` can block the response.

## Root cause

Quota enforcement is performed after credential exchange in the silent refresh paths, and only if a cached monitoring token can be loaded.

This makes the enforcement dependency inverted:

```text
current:
  refresh token/id token -> STS credentials -> save credentials -> maybe quota check -> output credentials

expected:
  refresh token/id token -> quota check -> STS credentials -> save/output credentials
```

## Impact

- Cost-based or token-based `block` enforcement can be bypassed on refresh paths.
- A failed or missing monitoring-token cache can become an allow path.
- If the quota check later blocks, already-saved AWS credentials may remain available to subsequent SDK calls.
- This is especially visible with long Direct STS sessions, but still matters with shorter test sessions such as 900 seconds.

## Proposed fix

Restructure the refresh paths so quota enforcement happens before AWS credentials are saved or returned.

Possible implementation approach:

- Make `trySilentRefresh()` and `tryRefreshToken()` return an auth result or fresh ID token plus claims, rather than already-exchanged AWS credentials.
- Run `quota.Check(...)` with that token before `getAWSCredentials(...)`.
- Only call `saveCredentials(...)` and `outputJSON(...)` after `allowed: true`.
- If `quota_api_endpoint` is set and no usable quota token can be obtained, do not silently skip the check.
- Ensure `quota_fail_mode` semantics are applied consistently.
- If a quota check returns `allowed: false`, clear any newly-created credentials before returning an error.

## Testing

Suggested regression tests:

- Silent refresh path with quota endpoint and `allowed: false` must not print credentials.
- Refresh-token path with quota endpoint and `allowed: false` must not print credentials.
- Missing monitoring token with quota endpoint must not silently skip enforcement.
- If a quota check blocks, no fresh AWS credentials should remain in keyring or session credentials storage.
- stdout remains valid credential JSON only on success; warnings/errors must not corrupt stdout.

## Related

- #606 tracked the broad symptom that `block` did not deny access.
- #655 covered warning visibility, not this credential issuance ordering problem.
- #664 covered refresh-token UX for CoWork, not quota enforcement ordering.

adiospeds:
## Why

Fixes #761. With `block` enforcement configured, over-quota users could still receive fresh AWS credentials:

1. **Ordering inversion:** on the silent-refresh and refresh_token paths, `trySilentRefresh()`/`tryRefreshToken()` exchanged the id_token for STS credentials and **saved them to cache** before `run()` performed the quota check. A block arrived after the credentials already existed, and the saved credentials could still be served from cache until the next recheck interval.
2. **Silent skip:** the quota check on those paths re-read the monitoring token from storage and was skipped entirely when that read came back empty (`if token != ""`), ignoring `quota_fail_mode`. IdPs issuing id_tokens with lifetimes inside the 10-minute storage buffer made this a systematic bypass.
3. **Same class, other entrypoints:** the desktop helper (`--desktop`) silent refresh minted a Bedrock bearer token with no quota check at all, and the `--get-monitoring-token` browser fallback (Go and Python) minted and cached credentials as a side effect with no quota check.

## What

- `run()` refresh paths now obtain the id_token first (`cachedTokenAuthResult()` / `tryRefreshToken()`), enforce quota with that in-scope token, and only then exchange for and persist credentials (`exchangeAndSaveCredentials()`). Blocked → exit 1 with nothing minted or cached.
- New `enforceQuota()` is the single pre-issuance check used by the browser, silent-refresh, refresh-token, and desktop paths. A configured endpoint with no usable token resolves via `quota_fail_mode` (closed → block; open → allow with a debug note) instead of being silently skipped. It never falls through to the SigV4 path with an empty token (credential recursion).
- `tryRefreshToken()` now performs only the token exchange (absorbing the former `refreshIDTokenOnly()` twin, which was line-for-line identical) — no caller can mint credentials without passing quota enforcement first. `performQuotaRecheck()` uses the fresh in-scope id_token instead of re-reading storage, closing the same silent fail-open on the cache-hit fast path.
- `--get-monitoring-token` (Go + Python) persists only the monitoring token after browser auth, so blocked users keep OTEL attribution without gaining credentials. Its consumers (otel-helper, `ccwb test`, install scripts) read only stdout, verified unaffected.
- Python `_try_silent_refresh` split into token-load + exchange phases with the quota check between them — Go/Python parity preserved (Python has no refresh_token store, so only the silent path applies).
- Docs: `credential-flow.md`, `token-endpoint-single-builder.md`, and the root `CLAUDE.md` flow updated to the quota-before-STS ordering.

## Behavior notes

- Browser-auth path ordering is unchanged (it already checked quota before STS).
- A quota block on a refresh path exits 1 **without** clearing the refresh_token, so users resume silently once quota resets (matches browser-path behavior).
- `--get-monitoring-token` no longer fails when the STS exchange would fail — more correct for a telemetry entrypoint, and matches the pre-existing Python behavior.
- In degraded STS conditions one invocation may check quota more than once as it falls through silent-refresh → refresh_token → browser; each path independently enforces before minting.

## Regression tests

Go (`quota_enforcement_order_test.go`, mock IdP + quota API + STS via `AWS_ENDPOINT_URL_STS`):
- Silent-refresh path, quota blocked → exit 1, **0 STS calls**, nothing cached (fails pre-fix: STS was called first and credentials saved)
- refresh_token path, quota blocked → same assertions
- Short-lived id_token (inside the 10-min buffer) → quota API still called and block enforced (fails pre-fix: check silently skipped, credentials issued)
- Quota allowed → STS called once, credentials issued and cached (happy path preserved)
- `enforceQuota("")` honors fail mode closed/open; no-endpoint no-op
- Desktop helper: blocked → exit 1, 0 STS calls, no token emitted; allowed → bearer token emitted
- `performQuotaRecheck` with short-lived fresh token → blocked and cached credentials wiped (fails pre-fix: fail-open)

Python (`test_silent_refresh.py`): quota blocked on silent refresh → blocked handler invoked, `get_aws_credentials`/`save_credentials` never called; quota allowed → exchange proceeds. Existing tests updated to the two-phase API.

Consolidating `refreshIDTokenOnly` into `tryRefreshToken` removed its two twin tests; the surviving `TestTryRefreshToken_*` pair now covers the single clear-vs-retain site for all callers, and `TestGetMCPAuthHeader_RefreshSucceedsWithoutAWSCreds` still pins the decoupling contract (fresh id_token survives STS failure).

## Verification

- `cd source/go && go build ./... && go vet ./... && go test ./... -race -count=1` — all pass
- `GOOS=windows GOARCH=amd64 go build ./...` and `GOOS=linux GOARCH=amd64 go build ./...` — clean
- `cd source && poetry run pytest tests/ -q` — 1669 passed, 27 skipped
- `ruff check` / `ruff format --check` / `gofmt` — clean

## Out of scope (pre-existing, noted during review)

- The desktop helper's cached-credentials fast path has no periodic quota recheck (unlike `run()`); a user crossing quota keeps working until cached credentials expire.
- Credential Tier 1 files touched: tested with OIDC flows end-to-end via mocks; IDC and passthrough entrypoints are untouched by this change (separate code paths, verified by their existing tests).

### DIFF
--- a/.claude/rules/credential-flow.md
+++ b/.claude/rules/credential-flow.md
@@ -11,8 +11,8 @@ AWS SDK → credential-process [Go or Python]
   2. If valid + not expired → return immediately (fast path, ~5ms)
   3. If expired → try silent refresh (refresh_token exchange, no browser)
   4. If no refresh_token → OIDC browser flow (or IDC passthrough)
-  5. Exchange token → STS AssumeRoleWithWebIdentity → temporary credentials
-  6. Check quota (if configured) → blocked? exit non-zero
+  5. Check quota with the id_token (if configured) → blocked? exit non-zero
+  6. Exchange token → STS AssumeRoleWithWebIdentity → temporary credentials
   7. Save credentials to cache
   8. Emit OTEL attribution headers to cache file
   9. Print JSON to stdout → AWS SDK uses credentials
@@ -39,10 +39,14 @@ print(json.dumps(credentials))  # only JSON on stdout
 
 ## Quota Integration
 
-Quota check happens AFTER auth but BEFORE outputting credentials:
+Quota check happens AFTER auth (id_token in hand) but BEFORE the STS
+exchange — an over-quota user must never mint or cache fresh AWS
+credentials (#761):
 - If quota blocks → exit non-zero (credential-process contract: no JSON = failure)
 - If quota warns → print warning to stderr, still output credentials
 - Quota check uses the same token (OIDC: Bearer, IDC: SigV4)
+- Check with the in-scope id_token; never re-read it from storage and
+  silently skip when the read comes back empty (fail per quota_fail_mode)
 
 ## OTEL Attribution
 
--- a/.claude/rules/token-endpoint-single-builder.md
+++ b/.claude/rules/token-endpoint-single-builder.md
@@ -31,8 +31,9 @@ tokenURL = provider.TokenEndpointURL(providerType, oktaAuthServerID, providerDom
 
 ## Callers (must all use the shared builder)
 - `oidc/flow.go` — browser authorization-code exchange
-- `tryRefreshToken()` in main.go — silent refresh
-- `refreshIDTokenOnly()` — monitoring token refresh (when added)
+- `tryRefreshToken()` in main.go — the single silent refresh_token exchange
+  (serves credential refresh, monitoring token refresh, and the MCP auth
+  header path; the former `refreshIDTokenOnly()` twin was merged into it)
 - Any future token endpoint caller
 
 ## Related
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,8 +46,8 @@ AWS SDK calls credential-process
   → read cache → [valid?] → return cached credentials
   → [expired?] → try silent refresh (refresh_token)
   → [no token?] → OIDC browser flow or IDC passthrough
+  → check quota with id_token → [blocked?] → exit non-zero
   → STS AssumeRoleWithWebIdentity → credentials
-  → check quota → [blocked?] → exit non-zero
   → emit OTEL attribution headers → cache → stdout JSON
 ```
 
--- a/source/credential_provider/__main__.py
+++ b/source/credential_provider/__main__.py
@@ -1679,14 +1679,12 @@ class MultiProviderAuth:
                 except Exception:
                     pass  # nosec B110
 
-            # Get AWS credentials (we need them but won't output them)
-            self._debug_print("Exchanging token for AWS credentials...")
-            credentials = self.get_aws_credentials(id_token, token_claims)
-
-            # Cache credentials for future use
-            self.save_credentials(credentials)
-
-            # Save monitoring token
+            # Persist the monitoring token only. Deliberately NO AWS credential
+            # exchange here: this entrypoint serves OTEL attribution, and
+            # minting credentials on it would bypass quota enforcement (#761) —
+            # every credential-minting path must enforce quota first. A blocked
+            # user still gets a monitoring token so their telemetry stays
+            # attributed.
             self.save_monitoring_token(id_token, token_claims)
 
             # Return just the monitoring token
@@ -2268,29 +2266,43 @@ class MultiProviderAuth:
     # End Quota Check Methods
     # ===========================================
 
-    def _try_silent_refresh(self):
-        """Attempt to refresh AWS credentials using a cached, still-valid OIDC id_token.
+    def _get_cached_token_for_silent_refresh(self):
+        """Return (id_token, token_claims) from the cached monitoring token, or (None, None).
 
-        Returns:
-            Tuple of (credentials, id_token, token_claims) if successful, (None, None, None) otherwise.
+        Read-only: no STS exchange happens here so the caller can enforce
+        quota with the token BEFORE any credentials are minted.
         """
         try:
             id_token = self.get_monitoring_token()
             if not id_token:
                 self._debug_print("No valid cached id_token for silent refresh")
-                return None, None, None
-
-            self._debug_print("Found valid cached id_token, attempting silent credential refresh...")
+                return None, None
             token_claims = jwt.decode(id_token, options={"verify_signature": False})
+            return id_token, token_claims
+        except Exception as e:
+            self._debug_print(f"Could not load cached id_token for silent refresh: {e}")
+            return None, None
 
+    def _try_silent_refresh(self, id_token, token_claims):
+        """Exchange a quota-cleared cached id_token for AWS credentials and persist them.
+
+        Callers MUST run the quota check first — this method mints and caches
+        credentials, so calling it before quota clearance would hand an
+        over-quota user a full STS session.
+
+        Returns:
+            Credentials dict if successful, None otherwise (caller falls back to browser auth).
+        """
+        try:
+            self._debug_print("Found valid cached id_token, attempting silent credential refresh...")
             credentials = self.get_aws_credentials(id_token, token_claims)
             self.save_credentials(credentials)
             self.save_monitoring_token(id_token, token_claims)
             self._debug_print("Silent credential refresh succeeded")
-            return credentials, id_token, token_claims
+            return credentials
         except Exception as e:
             self._debug_print(f"Silent refresh failed, will require browser auth: {e}")
-            return None, None, None
+            return None
 
     def _run_passthrough(self):
         """Emit credentials from the ambient AWS credential chain (Identity Center, env vars, instance profile).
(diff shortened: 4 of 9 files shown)


adiospeds:
@wirjo - can you review and merge this please.

wirjo:
## Review — Code Path Trace + Implications

### What this PR does (plain English)

When Claude Code or Claude Desktop needs fresh AWS credentials, there is a "silent refresh" path that renews them in the background without popping up a browser. **Before this PR**, the code first minted the credentials (via AWS STS) and *then* checked whether the user was over-quota. A blocked user already had usable, cached credentials by the time the check ran — so the block was ineffective.

This PR reorders the logic so the quota check happens **before** the STS exchange on every credential-minting path. No new AWS credentials can be issued unless the quota API says "allowed."

---

### Code path trace

#### Go credential-process (main.go)

| Path | Before | After |
|------|--------|-------|
| **Silent refresh** (run → trySilentRefresh) | trySilentRefresh = read token → STS exchange → save creds → then read token from storage again → quota check (skipped if storage read empty) | cachedTokenAuthResult() = read token only → enforceQuota(token) → exchangeAndSaveCredentials(auth) |
| **Refresh-token** (run → tryRefreshToken) | tryRefreshToken = token exchange + STS exchange + save creds → re-read monitoring token from storage → quota check (same skip risk) | tryRefreshToken() = token exchange only (returns AuthResult) → enforceQuota(auth.IDToken) → exchangeAndSaveCredentials(auth) |
| **Browser auth** (run → authenticate) | Quota checked with authResult.IDToken before STS ✓ | Same — now via shared enforceQuota() |
| **Desktop helper** (runDesktopHelper) | tryRefreshToken() returned credentials directly — **no quota check at all** | tryRefreshToken() → enforceQuota(auth.IDToken) → exchangeAndSaveCredentials(auth) |
| **--get-monitoring-token** | Browser/refresh path minted + saved AWS credentials as a side effect | Now persists *only* the monitoring token — no STS call, no credential minting |
| **--get-mcp-auth-header** | Used refreshIDTokenOnly() (a line-for-line duplicate of tryRefreshToken) | Uses the unified tryRefreshToken() (which no longer does STS) |

#### Key new functions

- **enforceQuota(idToken string) bool** — Single pre-issuance gate. If endpoint configured but token empty: honours quota_fail_mode (closed → block, open → allow). Never falls through to SigV4 (would recurse).
- **cachedTokenAuthResult()** — Read-only extraction of the cached monitoring token into an AuthResult. No network calls.
- **exchangeAndSaveCredentials(auth)** — The only place silent/refresh paths mint credentials. Comment mandates: "callers MUST run enforceQuota first."
- **refreshIDTokenOnly() is deleted** — its logic was a verbatim duplicate. tryRefreshToken() now returns an AuthResult (token + claims), not credentials.

#### Python (__main__.py)

Same structural split:
- _get_cached_token_for_silent_refresh() → read-only token extraction
- _try_silent_refresh(id_token, token_claims) → now takes pre-validated params, mints creds
- run() calls quota check between the two phases

---

### Implications

#### Performance
- **Latency on every refresh:** Yes, a quota API HTTP call is now on the critical path for *every* credential refresh (silent or browser). The existing QuotaCheckTimeout (configurable, default 2s) bounds this. For orgs with in-region quota endpoints this is likely 10-50ms, tolerable for a credential process that already does an STS roundtrip (~50-200ms).
- **Potential double-check in degraded flows:** If silent refresh fails and falls through to refresh_token and then browser, quota may be checked up to 3 times in one invocation. Each is independent and fast, so not a concern in practice.

#### Reliability (fail-open vs fail-closed)
- **Quota API down, quota_fail_mode=open:** enforceQuota returns true with a debug note — user proceeds. This is the safe default.
- **Quota API down, quota_fail_mode=closed:** If the API is truly unreachable and timeout fires, the quota package's Check() respects QuotaFailMode. The new enforceQuota handles the *no-token-available* case separately (prints error to stderr, returns false for closed mode). This is correct but operators should be aware: a closed-mode outage = **all users blocked**.
- **Token empty + fail-closed:** This is a real edge case. If the cached token expires and refresh fails, enforceQuota("") blocks under closed mode. The next attempt will go through browser auth which produces a token, so it's a single-failure transient.

#### Auth type scope
- **OIDC:** Fully affected — all paths reordered.
- **IDC / passthrough:** Explicitly out of scope, untouched code paths. These use SigV4 for quota, not Bearer tokens.
- **Desktop helper (Bedrock bearer):** Now covered — previously had zero quota enforcement on refresh.

#### UX: Could this break users mid-session?
- A quota block on refresh returns exit code 1, which the AWS SDK interprets as "credential-process failed." The SDK will retry (its own backoff), and if the user is still over-quota, the tool reports an error.
- **Crucially:** the refresh_token is NOT cleared on a quota block (matching browser-path behavior). Once quota resets, the next invocation silently succeeds without a browser popup. Good UX decision.

---

### Backwards compatibility

| Concern | Status |
|---------|--------|
| --get-monitoring-token no longer saves AWS credentials | ✅ Safe. Its consumers (otel-helper, ccwb test, install scripts) only read stdout, verified in the PR description. |
| tryRefreshToken() return type changed | Internal only (not exported). All callers updated in the same PR. getMCPAuthHeader and performQuotaRecheck adapted. |
| refreshIDTokenOnly() removed | Was internal. Its sole caller (getMCPAuthHeader) now uses tryRefreshToken() which returns the same IDToken field. |
| Test API change in Python _try_silent_refresh | Takes (id_token, claims) args now. All internal tests updated. Not a public API. |
| quota_fail_mode semantics for empty-token | **New behavior**: previously silently skipped → now respects the mode. Operators with closed mode may see blocks where they previously saw silent passes. This is *correct* but is a behavioral change worth noting in release notes. |

---

### Risk assessment

| Risk | Severity | Mitigation |
|------|----------|------------|
| Added latency on refresh | Low | Bounded by QuotaCheckTimeout (2s default); STS already adds 50-200ms |
| Quota API outage with closed mode blocks everyone | Medium | This is the *intended* semantics of closed mode; operators opt into it. Document in release notes. |
| Regression: silent refresh could fail if exchangeAndSaveCredentials returns nil after quota passes | Low | Falls through to browser auth (same as before); tested in the happy-path test. |
| Code deduplication (refreshIDTokenOnly removed) leaves MCP header path depending on the same function | Low | Actually reduces risk — one code path instead of two identical ones that could drift. |
| performQuotaRecheck now uses auth.IDToken from a fresh refresh instead of re-reading storage | Low-positive | Fixes the storage buffer skip bug; tested explicitly (TestPerformQuotaRecheck_ShortLivedTokenStillChecked) |

**Test coverage is excellent:** 466 new lines of Go tests with mock STS/IdP/quota servers, plus Python regression tests. The "would fail pre-fix" assertions add real confidence.

---

### Verdict

**Ship it.** 🚢

This is a well-structured security fix with correct ordering semantics, comprehensive regression tests, and thoughtful handling of edge cases (fail-mode, short-lived tokens, refresh_token preservation). The code deduplication (refreshIDTokenOnly → tryRefreshToken) is a welcome simplification.

Minor suggestions (non-blocking):
1. Consider adding a brief note to operator-facing docs about the behavioral change for quota_fail_mode=closed + empty-token scenarios.
2. The desktop helper cached-credentials fast path still has no periodic recheck (noted as out-of-scope) — worth a follow-up issue to track.

Great work @adiospeds.

PR #763, opened 2026-07-08
fix(package): filter extra_files by --target-platform

adiospeds:
## Why

`ccwb package --target-platform macos` copied **every** `extra_files` entry into the build folder, including windows-targeted ones — so a single-platform package shipped files intended only for other platforms. The superset design from #744 assumed `distribute` always filters later at zip time, but the package folder itself is the deliverable when admins share it directly.

## What

- **extra_files.py**: new `extra_applies_to_any()` matcher that resolves an entry's `targets` against the `platforms_to_build` vocabulary. Generic family tokens (`macos`/`linux`, produced by the non-Go build path) are expanded to their arches so an arch-targeted extra (`macos-arm64`) still ships in a generic `macos` build.
- **package.py**: `_copy_extra_files` now receives `platforms_to_build` and skips entries that don't apply to any platform being built (with a dim "skipped (not targeted for this build)" note). A skipped entry's missing `from` source no longer fails the build; applicable entries still fail fast with a non-zero exit.
- No change to `distribute`: per-OS zip filtering already works correctly. Since non-target extras now never reach the package folder, distribute archives can't include them either (missing files are skipped at zip time).

## Backward compatibility

- `--target-platform all` (and the non-Go "all" expansion) behaves exactly as before — every extra applies to some built platform.
- Users without `extra_files` configured are untouched — the early-return on an empty list is unchanged.
- `platforms_to_build=None` preserves the unfiltered superset behavior for any direct caller.

## Testing

- Regression test: windows-targeted extra + `platforms_to_build=["macos-arm64"]` → not copied; `all`/family-targeted entries still copied.
- Skipped entry with a missing `from` source no longer fails; an applicable entry with a missing source still returns a fatal error.
- Matcher truth table for `extra_applies_to_any` including family-token expansion and sibling-arch exclusion.
- All tests are self-contained (`tmp_path` fixtures, in-memory profiles) — no dependence on host files.

```
1680 passed, 27 skipped, 1 xfailed, 1 xpassed
```

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -811,11 +811,11 @@ class PackageCommand(Command):
             console.print("\n[cyan]Generating CoWork 3P MDM configuration...[/cyan]")
             self._generate_cowork_3p_mdm_config(output_dir, profile, profile_name)
 
-        # Copy admin-defined extra files into the build folder (superset; per-OS
-        # filtering happens in distribute at zip time). Fail fast on a missing
-        # source or a validation error — a listed cert the admin expects shipped
-        # must not be silently skipped.
-        copied_extra_files = self._copy_extra_files(profile, output_dir, console)
+        # Copy admin-defined extra files into the build folder, filtered to the
+        # platforms actually being built (distribute filters again per-OS at zip
+        # time). Fail fast on a missing source or a validation error — a listed
+        # cert the admin expects shipped must not be silently skipped.
+        copied_extra_files = self._copy_extra_files(profile, output_dir, console, platforms_to_build)
         if copied_extra_files is None:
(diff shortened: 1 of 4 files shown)


adiospeds:
@wirjo can you review and merge this minor fix.

wirjo (approved):
Reviewed the diff and logic — this is a clean, well-scoped bug fix. The new `extra_applies_to_any()` matcher correctly reuses the existing `extra_applies_to()` with platform expansion, the filtering in `_copy_extra_files` is opt-in via the `platforms_to_build` parameter (None preserves superset behavior), and the test coverage is thorough including regression, edge cases, and backward compatibility. LGTM.

wirjo:
Merged — thanks @adiospeds! Clean fix with solid test coverage. 🎉

PR #764, opened 2026-07-08
feat(package): stamp settings_version into OTEL_RESOURCE_ATTRIBUTES

adiospeds:
## Why

Admins distribute packages from timestamped dist folders (`dist/<profile>/<timestamp>`), but once installed there is no way to tell which distribution each user is actually running. When settings change (new model defaults, telemetry config, quota wiring), there's no signal for who has upgraded and who is still on an old package.

## What

`ccwb package` now stamps the dist-folder timestamp into the generated `settings.json` as a `settings_version` entry in `OTEL_RESOURCE_ATTRIBUTES`, so every user's telemetry carries the version of the package they installed.

- Applied in both the standard package command and the CodeBuild variant, including the regenerate-installers path (which stamps its new output-dir timestamp, matching the regenerated `config.json`).
- Admin-customized telemetry attributes from the wizard still get the version appended.
- Packages built before this change simply report no `settings_version` — itself an outdated-package signal.
- Works identically across OIDC, IDC, and none auth modes and both collector modes — it's a plain env var attached by Claude Code's OTLP export, not part of the auth/attribution chain.

With the analytics pipeline enabled, the `awsemf` exporter's `resource_to_telemetry_conversion` surfaces the attribute as a field on every metric event in `/aws/claude-code/metrics`, so adoption is queryable in Logs Insights with no collector change:

```
fields @timestamp
| stats count_distinct(`user.email`) as users by settings_version
```

Deployments that want `settings_version` as a first-class CloudWatch dimension can add a scoped `metric_declarations` entry in `otel-collector.yaml`; the docs note the cardinality/cost caveat (each release adds a new dimension value).

## Notes

- No new `Profile` field — the version derives from the timestamp already in scope, so no Go `ProfileConfig` mirror or `--explain` change is needed.
- Docs updated in `assets/docs/MONITORING.md` (new section under per-user attribution).

## Testing

- Four new tests in `tests/cli/commands/test_package.py`: version appended to default attributes, appended to custom attributes, omitted when absent (backward compat), and a parity test for the CodeBuild variant.
- Full suite: 1673 passed, 27 skipped.
- Verified locally: generated `claude-settings/settings.json` ends with `,settings_version=<dist folder timestamp>`.

### DIFF
--- a/assets/docs/MONITORING.md
+++ b/assets/docs/MONITORING.md
@@ -130,6 +130,31 @@ are only populated for OIDC. IDC attributes `user.email` only. See
 [Cost Attribution](COST_ATTRIBUTION.md#idc-limitation) for the CUR/ABAC path to
 per-user team/department breakdowns under IDC.
 
+### Tracking deployed package versions (`settings_version`)
+
+`ccwb package` stamps the dist-folder timestamp (e.g. `2026-07-08-120000`) into
+the generated `settings.json` as a `settings_version` entry in
+`OTEL_RESOURCE_ATTRIBUTES`. Every user's telemetry then carries the version of
+the distribution package they installed, so you can see who is running an
+outdated package. Installs from packages built before this attribute existed
+simply report no `settings_version` — itself a signal that the user is on an
+old package.
+
+When the analytics pipeline is enabled, the `awsemf` exporter converts resource
+attributes to fields on every metric event in `/aws/claude-code/metrics`, so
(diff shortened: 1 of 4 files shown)


adiospeds:
@wirjo can you review and merge this one - this will also be useful for pinpointing issues to specific builds and tracking the settings version for multiple users have across an org.

wirjo:
Merged — thanks @adiospeds! Nice addition for tracking package versions in telemetry. 👍

PR #765, opened 2026-07-09
fix(package): make install.bat actually merge existing Claude settings

wantmys2000:
## Why
On Windows, running `install.bat` with an existing `~/.claude/settings.json` silently applies **nothing**. The PowerShell merge one-liner is generated from an f-string with over-escaped braces (`{{{{` -> `{{`), so PowerShell parses the `foreach` body as a scriptblock *literal* that is never invoked — the user's file is rewritten unchanged, and the Bedrock env, credential process, and telemetry settings are never configured. The failure is also undetectable: `%errorlevel%` inside a parenthesized batch block expands at parse time, so the error branch can never fire.

## What
- Emit single braces so the merge loop actually executes
- Deep-merge the `env` object (incoming keys win; user-added env vars survive) instead of replacing it wholesale
- Run the merge under `$ErrorActionPreference = 'Stop'` with `catch { exit 1 }`, and check `!errorlevel!` / `!OVERWRITE!` / `!WRITE_SETTINGS!` with delayed expansion so failures surface
- Fix the dead overwrite path: answering `y` to "Merge failed. Overwrite?" previously did nothing because the write branch also required the settings file to not exist

## Testing
- New `TestWindowsInstallerSettingsMerge` regression tests: static assertions (no doubled braces, deep env merge, delayed-expansion checks) run on every platform; a functional test extracts the generated merge command and executes it under real PowerShell on Windows CI, asserting new keys land, placeholders resolve, and user customizations (top-level keys and custom env vars) survive
- Full suite passes on Linux (1688 passed)

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -3658,7 +3658,7 @@ REM external command" and silently drops all metrics, so we fail the install
 REM loudly rather than leave a broken telemetry config.
 if exist "otel-helper.cmd" (
     copy /Y "otel-helper.cmd" "%USERPROFILE%\\claude-code-with-bedrock\\otel-helper.cmd" >nul
-    if %errorlevel% neq 0 (
+    if !errorlevel! neq 0 (
         echo ERROR: Failed to copy otel-helper.cmd
         pause
         exit /b 1
@@ -3676,7 +3676,7 @@ if exist "otel-helper.cmd" (
 )
 if exist "otel-helper.ps1" (
     copy /Y "otel-helper.ps1" "%USERPROFILE%\\claude-code-with-bedrock\\otel-helper.ps1" >nul
-    if %errorlevel% neq 0 (
+    if !errorlevel! neq 0 (
         echo ERROR: Failed to copy otel-helper.ps1
         pause
         exit /b 1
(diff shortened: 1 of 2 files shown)


sg-nitd:
**Confirmed in a production enterprise deployment (v2.5.x, Windows + Entra ID OIDC, central monitoring mode).**

Every Windows user was missing from the CloudWatch dashboards, the OTLP metric store, and DynamoDB quota tracking, while macOS users (install.sh path) appeared normally. Diagnosis matched this PR's root cause exactly:

- `install.bat` printed `OK Claude Code settings merged [user settings preserved]` but also echoed the literal loop body three times (once per incoming property) — the scriptblock-literal no-op described here.
- The user's `~/.claude/settings.json` was left with only their pre-existing keys; `otelHeadersHelper`, `CLAUDE_CODE_ENABLE_TELEMETRY`, and all `OTEL_*` env vars were silently dropped, so Claude Code never exported telemetry.
- PromQL against the OTLP store showed the machine's usage either absent or in the empty `{}` label bucket.
- Hand-fixing the doubled braces in the generated install.bat and re-running it resolved the issue immediately; the user's attributed usage appeared in `sum by ("user.email")(...)` within minutes.

The silent-success aspect (`%errorlevel%` parse-time expansion) made this hard to catch in the field — the installer output looks healthy. +1 for merging; happy to provide more environment details if useful.

wirjo:
Merged — thanks @wantmys2000! Great catch on the doubled braces and errorlevel expansion. The deep env merge is nicely done. 🎉

PR #766, opened 2026-07-09
fix(otel): honor supplied --profile across all otel-helper variants

wantmys2000:
## Why
The otel-helper resolves its profile only from the `AWS_PROFILE` env var, and the packaged Claude Code settings give `otelHeadersHelper` no profile argument at all — unlike `AWS_CREDENTIAL_PROCESS` and `awsAuthRefresh`, which already pass `--profile`. Users on a non-default profile get attribution headers from the wrong cache and a collector sidecar running under the wrong AWS profile, with no way to pin the helper to the right one.

## What
One resolution rule across all four helper variants, aligned with credential-process's existing `CCWB_PROFILE` convention:

`--profile` flag > `CCWB_PROFILE` env > `AWS_PROFILE` env > `"ClaudeCode"`

An explicit override is exported to `AWS_PROFILE` so the header cache path, credential-process subprocess, `<profile>-collector` sidecar profile, and the AWS SDK in proxy mode all resolve the same profile.

- `package.py` / `package_cb.py` write `otelHeadersHelper` as `__OTEL_HELPER_PATH__ --profile <name>`
- Go: new `--profile` flag + `resolveProfile()`; `--status` reports the resolved profile instead of re-reading `AWS_PROFILE` inline
- Python: new `--profile` argparse option; `otel-helper.sh` parses `--profile`; `otel-helper.ps1` hand-parses it from `$args` (PowerShell `-File` can't bind Go-style `--profile`, so the old `param([string]$Profile)` errored when `otel-helper.cmd` forwarded args — and shadowed the automatic `$PROFILE` variable)

## Testing
- Go `TestResolveProfile_Precedence` pins the four-tier order and the `AWS_PROFILE` export; Python `tests/test_otel_helper_profile.py` mirrors it and asserts the cache path follows the flag; packaging tests updated to expect the `--profile` suffix
- Full Python suite (1688) and `go test ./...` pass; cross-compiles for windows/darwin verified

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/package.py
+++ b/source/claude_code_with_bedrock/cli/commands/package.py
@@ -4390,7 +4390,10 @@ Available metrics include:
                     _is_idc = getattr(profile, "effective_auth_type", profile.auth_type) == "idc"
                     _idc_zero_binary = _is_idc and not bool(getattr(profile, "quota_api_endpoint", None))
                     if not _idc_zero_binary:
-                        settings["otelHeadersHelper"] = "__OTEL_HELPER_PATH__"
+                        # Pass the profile explicitly (same as AWS_CREDENTIAL_PROCESS /
+                        # awsAuthRefresh above) so the helper serves THIS profile even
+                        # when AWS_PROFILE in the helper's environment points elsewhere.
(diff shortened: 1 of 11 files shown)


wirjo:
Merged — thanks @wantmys2000! Great work threading the profile consistently across all variants. 👏

PR #768, opened 2026-07-10
fix(credential-process): quota-blocked users recover automatically — remove ~/.aws/credentials stanza instead of writing EXPIRED placeholder

### LINKED ISSUE #767: Quota-blocked users never recover after quota reset/unblock: clearCache() writes EXPIRED placeholder to ~/.aws/credentials, permanently shadowing credential_process
## Summary

When quota enforcement blocks a user, `clearCache()` in the Go `credential-process` writes an `EXPIRED` placeholder credential stanza into `~/.aws/credentials`. Because that file outranks `credential_process` in the AWS SDK's credential resolution chain, the placeholder permanently wedges every SDK consumer (Claude Code, Claude Desktop): the SDK keeps reading the static `EXPIRED` keys and **never invokes credential-process again** — so when the quota resets (new day/month) or an admin runs `ccwb quota unblock`, the user does not recover. They stay on "credentials expired / refresh failed" until someone manually deletes the stanza.

## Environment

- Auth: OIDC (Entra ID), Direct STS federation
- Credential helper: Go `credential-process`
- Credential storage: **keyring** (also reproduces in session mode)
- Quota monitoring enabled, enforcement `block`
- Platforms: macOS and Windows (mechanism is OS-independent)

## Steps to reproduce

1. Configure quota enforcement (`block`) with a quota API endpoint.
2. Exceed the user's quota; on the next credential invocation the quota check returns `allowed: false` and `clearCache()` runs. Observe `~/.aws/credentials` now contains:
   ```ini
   [<profile>]
   aws_access_key_id     = EXPIRED
   aws_secret_access_key = EXPIRED
   aws_session_token     = EXPIRED
   x-expiration          = 2000-01-01T00:00:00Z
   ```
3. Lift the block (wait for the daily/monthly reset, or `ccwb quota unblock <user>`).
4. Launch Claude Code (or run `aws sts get-caller-identity --profile <profile>`).

**Expected:** the SDK falls through to `credential_process`, which re-checks quota (now allowed) and mints fresh credentials — automatic recovery.

**Actual:** the SDK resolves the static `EXPIRED` keys from `~/.aws/credentials` and fails with expired/invalid credentials forever. `credential-process` is never invoked, so the quota re-check that would unblock the user never runs. Running the binary manually works (it bypasses the SDK chain and reads the keyring), which makes the failure confusing to diagnose.

## Root cause

`clearCache()` (`source/go/cmd/credential-process/main.go`):

```go
func (a *credentialApp) clearCache() {
	if a.cfg.CredentialStorage == "keyring" {
		_ = storage.ClearKeyring(a.profile)
	}
	// Also clear session file
	expired := &federation.AWSCredentials{
		Version: 1, AccessKeyID: "EXPIRED", SecretAccessKey: "EXPIRED",
		SessionToken: "EXPIRED", Expiration: "2000-01-01T00:00:00Z",
	}
	_ = storage.SaveToCredentialsFile(expired, a.profile)
	...
}
```

The placeholder write is unconditional — even in keyring mode, where `~/.aws/credentials` is otherwise never touched. The intent (invalidating a stale session-file entry, consumed by `IsExpiredDummy()`) is legitimate, but **deleting the profile section achieves the same invalidation without poisoning the SDK chain**: the binary already treats "no entry" and "expired dummy" identically (`getCachedCredentials` returns nil for both), while the SDK treats them very differently — "no entry" falls through to `credential_process`; a static entry is terminal.

## Field impact

Observed in a production enterprise deployment: a user quota-blocked overnight remained hard-down the following day after the daily reset, reporting "token expired and refresh failed" on every Claude Code launch while the same binary run manually minted credentials fine. Deleting the `[profile]` stanza from `~/.aws/credentials` restored service instantly. In keyring mode the stanza is written **only** at block time, so every blocked-then-unblocked keyring user hits this.

## Proposed fix

Replace the placeholder write with removal of the profile section (atomic rewrite, other profiles preserved). PR to follow.

## Workarounds until fixed

- Delete the `[<profile>]` block from `~/.aws/credentials` after a quota block is lifted (keyring deletion alone is NOT sufficient — the file stanza is the wedge).

Related: #606 / PR #609 introduced the clear-cache-on-block behavior (correctly, to make `block` mode effective); this issue is the recovery half of that lifecycle. Distinct from #761 (enforcement gap on refresh paths).

sg-nitd:
## Why

Fixes #767.

When quota enforcement blocks a user, `clearCache()` writes an `EXPIRED` placeholder credential stanza into `~/.aws/credentials`. That file outranks `credential_process` in the AWS SDK resolution chain, so SDK consumers (Claude Code / Claude Desktop) keep reading the dead static keys and **never invoke credential-process again**. When the quota resets or an admin runs `ccwb quota unblock`, the user does not recover — they stay on "credentials expired / refresh failed" until the stanza is deleted by hand. In keyring mode the stanza is written *only* at block time, so every blocked-then-unblocked keyring user hits this.

Confirmed in a production enterprise deployment (Entra ID OIDC, Direct STS, keyring storage): a user blocked overnight remained hard-down after the daily reset; the same binary run manually minted credentials fine (it bypasses the SDK chain), and deleting the `[profile]` stanza restored service instantly.

## What

- **storage**: add `RemoveFromCredentialsFile(profile)` — deletes just that profile's section from `~/.aws/credentials` via the same atomic tmp-write/chmod/rename pattern as `SaveToCredentialsFile`; preserves other profiles; no-op when the file or section is absent.
- **clearCache()**: call it instead of writing the `EXPIRED` placeholder.

Deleting the section achieves the same invalidation the placeholder was for: `getCachedCredentials()` already treats "no entry" and "expired dummy" identically, while the SDK treats them very differently — "no entry" falls through to `credential_process` (automatic recovery); a static entry is terminal.

`IsExpiredDummy()` and its callers are intentionally kept so binaries remain tolerant of credentials files poisoned by older builds.

## Testing

- `go build ./...`, `go vet`, and the full `internal/storage` + `cmd/credential-process` suites pass.
- End-to-end: built the binary, seeded a sandbox `$HOME/.aws/credentials` with the incident state (EXPIRED stanza for the CCWB profile + an unrelated static profile), ran `--clear-cache`: the CCWB profile section is removed entirely, the unrelated profile survives byte-for-byte.

## Behavior change

Before: after a quota block, `~/.aws/credentials` contains a poisoned stanza that requires manual deletion once the block is lifted.
After: after a quota block, the profile has no entry in `~/.aws/credentials`; the SDK falls through to `credential_process` on the next invocation, the quota re-check runs, and recovery is automatic.

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -274,12 +274,13 @@ func (a *credentialApp) clearCache() {
 	if a.cfg.CredentialStorage == "keyring" {
 		_ = storage.ClearKeyring(a.profile)
 	}
-	// Also clear session file
-	expired := &federation.AWSCredentials{
-		Version: 1, AccessKeyID: "EXPIRED", SecretAccessKey: "EXPIRED",
-		SessionToken: "EXPIRED", Expiration: "2000-01-01T00:00:00Z",
-	}
-	_ = storage.SaveToCredentialsFile(expired, a.profile)
+	// Remove the profile section from ~/.aws/credentials rather than writing an
+	// EXPIRED placeholder. That file outranks credential_process in the AWS SDK
+	// resolution chain, so a placeholder entry would permanently wedge SDK
+	// consumers (Claude Code): they'd keep reading the static EXPIRED keys and
+	// never invoke this binary again. With the section removed, the SDK falls
+	// through to credential_process and recovery is automatic.
+	_ = storage.RemoveFromCredentialsFile(a.profile)
 	// Clear refresh token
 	storage.ClearRefreshToken(a.profile)
 	fmt.Fprintf(os.Stderr, "Cleared cached credentials for profile '%s'\n", a.profile)
--- a/source/go/internal/storage/session.go
+++ b/source/go/internal/storage/session.go
@@ -105,6 +105,46 @@ func credentialsFilePath() string {
 	return filepath.Join(home, ".aws", "credentials")
 }
 
+// RemoveFromCredentialsFile deletes a profile's section from ~/.aws/credentials.
+// Leaving a stale or placeholder section behind is harmful: ~/.aws/credentials
+// outranks credential_process in the AWS SDK resolution chain, so any static
+// entry for the profile prevents the SDK from ever invoking this binary again.
+// Removing the section lets the SDK fall through to credential_process and
+// recover automatically. Other profiles in the file are preserved.
+func RemoveFromCredentialsFile(profile string) error {
+	credPath := credentialsFilePath()
+	if _, err := os.Stat(credPath); os.IsNotExist(err) {
+		return nil
+	}
+
+	cfg, err := ini.LoadSources(ini.LoadOptions{
+		IgnoreInlineComment: true,
+	}, credPath)
+	if err != nil {
+		return fmt.Errorf("reading credentials file: %w", err)
+	}
+
+	if _, err := cfg.GetSection(profile); err != nil {
+		return nil // Section doesn't exist — nothing to remove
+	}
+	cfg.DeleteSection(profile)
+
+	// Atomic write, same pattern as SaveToCredentialsFile
+	tmpPath := credPath + ".tmp"
+	if err := cfg.SaveTo(tmpPath); err != nil {
+		return fmt.Errorf("writing credentials: %w", err)
+	}
+	if err := os.Chmod(tmpPath, 0600); err != nil {
+		os.Remove(tmpPath)
+		return err
+	}
+	if err := os.Rename(tmpPath, credPath); err != nil {
+		os.Remove(tmpPath)
+		return err
+	}
+	return nil
+}
+
 // IsExpiredDummy checks if credentials are the "EXPIRED" placeholder.
 func IsExpiredDummy(creds *federation.AWSCredentials) bool {
 	return creds != nil && creds.AccessKeyID == "EXPIRED"


wirjo (approved):
## Risk Assessment — Approved ✅

### (1) Unintended side effects on other profiles
**Risk: None.** `RemoveFromCredentialsFile` loads the INI, calls `cfg.DeleteSection(profile)` targeting only the named section, then rewrites atomically. Other profiles are preserved byte-for-byte — same pattern as `SaveToCredentialsFile`. The PR author confirmed this with an end-to-end test (sandbox credentials with an unrelated profile surviving intact).

### (2) Race conditions with concurrent invocations
**Risk: Very Low — pre-existing, not introduced by this PR.** The atomic-write pattern (`.tmp` → `Rename`) is identical to `SaveToCredentialsFile`, so the concurrency posture is unchanged. Two concurrent writers to `.tmp` could race (last writer wins), but this is the existing behavior for all credential writes in this codebase and is benign: both are removing the same section. A lost rename would simply leave the section present one extra cycle until the next invocation cleans it up.

### (3) Backward compatibility on downgrade
**Risk: Very Low.** `IsExpiredDummy()` and its callers are intentionally preserved, so older binaries that encounter a missing section simply get `nil` from `ReadFromCredentialsFile` (same as an expired dummy — both return nil in `getCachedCredentials`). A downgraded binary that later writes the EXPIRED placeholder would re-introduce the old behavior, but no crash or data corruption occurs.

### Summary
All 24 CI checks pass (including Go tests on Linux, macOS, Windows). The change is minimal, well-scoped, and addresses a confirmed production regression. The new function follows established patterns exactly. **Very low risk — approved.**

wirjo:
Merged — thanks @sg-nitd! Clean fix, atomic removal with no side effects on other profiles. Nicely done. 🙌

PR #776, opened 2026-07-13
feat(models): Claude Opus 4.8 in GovCloud + partition-aware model wizard

wantmys2000:
## Why
Claude Opus 4.8 is [available in AWS GovCloud (US) as of May 2026](https://aws.amazon.com/about-aws/whats-new/2026/05/claude-opus-4.8-aws-govcloud-us/) — in-region in us-gov-west-1 and geo-routed from us-gov-east-1 via the `us-gov.` CRIS prefix ([regional availability](https://docs.aws.amazon.com/bedrock/latest/userguide/models-region-compatibility.html)) — but the model registry has no entry for it, so GovCloud admins can't select it.

Fixing that surfaced deeper GovCloud gaps in model selection:
- The wizard offers **every commercial model to GovCloud admins** (and the `-govcloud` entries to commercial admins). Picking a cross-partition model deploys fine and then fails on every invoke.
- GovCloud entries render with an **empty region label** (the label builder doesn't know the `us-gov` profile key), and "Europe" never displays (checks the legacy `"europe"` key; profiles use `"eu"`).
- The `-govcloud` keys are absent from `MODEL_TIER_PREFERENCES`, so `DEFAULT_*_MODEL` aliases **never resolve for GovCloud deployments** (this already affected GovCloud Sonnet).

## What
- New `opus-4-8-govcloud` registry entry: `us-gov.anthropic.claude-opus-4-8`, source regions us-gov-west-1/east-1 (mirrors the existing `sonnet-4-5-govcloud` shape and the dateless ID convention of the commercial 4.8 entries).
- Wizard model choices filtered by the selected region's partition, with a partition-appropriate default, a "GovCloud" region label, and the eu/europe label fix. Filtering lives in a testable `_model_keys_for_region()` helper.
- Tier resolution learns `us-gov`: `-govcloud` keys added to the tier preference lists (safe for commercial — CRIS prefixes only match their own partition's profiles), and `us-gov` joins `DATA_RESIDENCY_PREFIXES` since it's a partition boundary — resolution yields a `us-gov.*` ID or nothing, never a commercial fallback that can't work. GovCloud haiku falls through to GovCloud Sonnet, mirroring jp/au.

## Testing
`tests/test_govcloud_model_selection.py` pins the registry entry, wizard filtering in both directions, the us-gov tier chain including the no-commercial-fallback invariant, alias resolution for `us-gov.*` IDs, and that commercial tier resolution is unchanged. Existing model-structure guard test updated. Full suite passes (1703 tests).

### DIFF
--- a/source/claude_code_with_bedrock/cli/commands/init.py
+++ b/source/claude_code_with_bedrock/cli/commands/init.py
@@ -32,6 +32,26 @@ from claude_code_with_bedrock.cli.utils.validators import (
 from claude_code_with_bedrock.config import WEBSEARCH_SUPPORTED_REGIONS, Config, Profile
 
 
+def _model_keys_for_region(region: str | None) -> list[str]:
+    """Model registry keys the wizard should offer for a target AWS region.
+
+    GovCloud is a separate partition: only models with a "us-gov" CRIS profile
+    are invokable there, and those profiles are not invokable from commercial
+    regions. Filtering keeps the wizard from offering models the selected
+    region can never serve (previously a GovCloud admin saw every commercial
+    model, and picking one produced a deployment that 400s on invoke).
+    """
+    from claude_code_with_bedrock.models import CLAUDE_MODELS, get_available_profiles_for_model
+    from claude_code_with_bedrock.utils.partition import aws_partition_for_region
+
+    in_govcloud = aws_partition_for_region(region or "") == "aws-us-gov"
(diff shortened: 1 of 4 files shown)


wirjo:
Merged — @scouturier FYI, partition-aware model wizard + Opus 4.8 GovCloud support. 🎉

PR #780, opened 2026-07-14
feat(credential-process): add --show-claims IdP claim diagnostic

wantmys2000:
## Why
Admins wiring group-based quota policies, cost attribution, or OTEL identity mapping need to answer "what claims is my IdP actually sending — is `groups` present, and under which claim name?" The Go credential-process only exposes slices of the ID token (`--show-tags` for the principal-tags claim, `--get-tag` for one tag), so the current workaround is hand-decoding JWTs. The Python provider already dumps the full claim set under `COGNITO_AUTH_DEBUG=1` — a credential-helper-parity gap.

## What
New `--show-claims` flag: prints **all** ID-token claims as pretty JSON on stdout (pipe-friendly for `jq`). Token acquisition mirrors `--show-tags` exactly — cached monitoring token first, fresh OIDC sign-in only when nothing is cached. IDC profiles get a clear "IDC has no JWT" explanation instead of an empty dump. Follows the cmd conventions: flag not subcommand, dispatched in the early-exit section, no change to the `--explain` contract.

## Testing
Tests cover the cached-token path (full claim set surfaces including `groups` and `custom:*` claims, valid JSON, no network/browser) and pin the flag registration. Full Go suite and Python suite (1692) pass; windows/darwin cross-compiles verified.

### DIFF
--- a/source/go/cmd/credential-process/main.go
+++ b/source/go/cmd/credential-process/main.go
@@ -50,6 +50,7 @@ func main() {
 	checkExpiration := flag.Bool("check-expiration", false, "Check if credentials are expired")
 	refreshIfNeeded := flag.Bool("refresh-if-needed", false, "Refresh credentials if expired")
 	showTags := flag.Bool("show-tags", false, "Print the https://aws.amazon.com/tags claim from the cached ID token (debug)")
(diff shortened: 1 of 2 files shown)


wirjo:
Merged — @scouturier FYI, new --show-claims diagnostic for IdP debugging. 👍

PR #783, opened 2026-07-14
fix(cowork): use inferenceBedrockProfile; remove non-functional helper mode

### LINKED ISSUE #782: CoWork 3P: default credential mode (helper) is non-functional — Claude Desktop rejects credential-process output
## Problem

Since #681, CoWork 3P defaults to `cowork_credential_mode = "helper"`, which sets Claude Desktop's `inferenceCredentialHelper` to run `credential-process` directly.

This never works. Claude Desktop's credential-helper contract requires stdout to be a bare token or `{"token": "..."}`, but `credential-process` emits AWS SigV4 `credential_process` JSON (`AccessKeyId`/`SecretAccessKey`/`SessionToken`). Claude Desktop rejects it:

> Credential helper output was not a usable credential (Output ... isn't valid JSON with a 'token' string)

Result: **every CoWork session fails at authentication** on a default deployment. Reproduced live against a real Cognito-backed OIDC setup on Windows.

## Fix

Use `inferenceBedrockProfile` (the AWS SDK `credential_process` path Claude Code CLI already uses), and remove the non-functional helper mode. Validated end-to-end: CoWork authenticates and Bedrock responds.

## Scope

CoWork 3P only. Claude Code CLI is unaffected (its `~/.aws/config` credential flow is unchanged).

scouturier:
## Why

CoWork 3P defaults to `cowork_credential_mode = "helper"` (#681), which points Claude Desktop's `inferenceCredentialHelper` at `credential-process`. That contract requires stdout to be a bare token or `{"token": "..."}`, but `credential-process` emits AWS SigV4 `credential_process` JSON — so Claude Desktop rejects it and **every CoWork session fails at auth** on a default deployment. Reproduced live against a real Cognito OIDC + Windows Claude Desktop setup.

## What

- Default to `inferenceBedrockProfile` — the AWS SDK `credential_process` path Claude Code CLI already uses (validated end-to-end: CoWork authenticates, Bedrock responds).
- Remove the non-functional helper mode: the `cowork_credential_mode` / `cowork_credential_helper_ttl_sec` fields, the `build_mdm_config` helper branch, `_credential_process_path()`, and the `.reg` helper-path rewrite.
- Update docs and tests to the profile-only flow.

Net −167 lines; no new features (repo is in maintenance mode).

## Scope / safety

- **CoWork 3P only.** Claude Code CLI is unaffected — its `~/.aws/config` credential flow and the credential-process binary are unchanged.
- **Backward compatible:** old `profile.json` files that set `cowork_credential_mode` still load (`Profile.from_dict` filters unknown keys).

## Testing

- `pytest tests/` passes (one pre-existing unrelated failure: `test_installer_launcher`).
- ruff check + format clean.
- Verified live on Windows: profile-mode config authenticates and returns Bedrock responses.

Fixes #782

### DIFF
--- a/assets/docs/COWORK_3P.md
+++ b/assets/docs/COWORK_3P.md
@@ -166,40 +166,17 @@ Claude Desktop supports object entries in `inferenceModels` with `anthropicFamil
 
 ### How Credentials Flow
 
-This solution supports two credential modes for CoWork. The **credential helper** mode (default since v2.6.0) is recommended because it gives Claude Desktop direct control over credential lifecycle, eliminating the stale-credential bug that required manual app restarts.
-
-#### Credential Helper Mode (Recommended)
-
-When Claude Cowork starts a session:
-
-1. Claude Desktop reads `inferenceCredentialHelper` from the MDM policy — this points directly at the credential-process binary
-2. Claude Desktop runs the binary and reads temporary AWS credentials from stdout
-3. The output is cached for `inferenceCredentialHelperTtlSec` seconds (default: 3500, just under the 1h STS token lifetime)
-4. When the cache expires, Claude Desktop automatically re-runs the helper — **no restart required**
-5. If credentials are rejected mid-session, Claude Desktop re-runs with `CLAUDE_HELPER_CONTEXT=mid-session-refresh` for seamless recovery (20s timeout)
-
-The credential-process binary handles the `CLAUDE_HELPER_CONTEXT` environment variable:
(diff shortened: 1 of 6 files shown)


scouturier:
Closing as obsolete. This was built on a stale local `beta` (45 commits behind). Current `origin/beta` already merged #733, which fixes CoWork 3P via a Bedrock bearer token (`credential-process --desktop`) — the seamless approach. Reverting to profile mode would regress that. Validating #733 directly instead.

PR #785, opened 2026-07-14
fix(auth): grant bedrock:CallWithBearerToken on Google federated role

### LINKED ISSUE #784: CoWork 3P on Google auth: missing bedrock:CallWithBearerToken → 403 Invalid API Key format
## Problem

CoWork 3P uses `credential-process --desktop` (#733) to emit a short-term Bedrock **bearer token** for Claude Desktop's `inferenceCredentialHelper`. Bedrock bearer-token calls require the IAM action `bedrock:CallWithBearerToken` on the federated role.

Commit `b44bd1d` added that action to the auth templates, but **`bedrock-auth-google.yaml` was missed**. Every other provider template (Okta, Azure, Auth0, Cognito, generic, IDC) grants it; Google does not.

Result: on a **Google-auth** deployment, CoWork fails with:

> 403 AccessDeniedException — Invalid API Key format: Must start with pre-defined prefix

(The token authenticates as the correct assumed-role principal, but the role isn't authorized to call `bedrock:CallWithBearerToken`.) Reproduced live this pattern on a Cognito-backed setup; Google deployments hit it because the permission is absent.

## Fix

Add `bedrock:CallWithBearerToken` to both Bedrock invoke statements (Regional + Global) in `bedrock-auth-google.yaml`, matching the other 7 templates.

## Scope

CFN template only, Google auth path. No code changes. cfn-lint clean.

scouturier:
## Why

CoWork 3P uses `credential-process --desktop` (#733) to emit a short-term Bedrock bearer token for Claude Desktop's `inferenceCredentialHelper`. Bedrock bearer-token calls require `bedrock:CallWithBearerToken` on the federated role.

`b44bd1d` added that action to the auth templates but missed `bedrock-auth-google.yaml`. Every other provider (Okta, Azure, Auth0, Cognito, generic, IDC) grants it; Google does not — so **Google-auth CoWork deployments fail** with `403 AccessDeniedException — Invalid API Key format`.

## What

Add `bedrock:CallWithBearerToken` to both Bedrock invoke statements (Regional + Global) in `bedrock-auth-google.yaml`, matching the other 7 templates. 2 lines, CFN-only.

## Testing

- `cfn-lint deployment/infrastructure/bedrock-auth-google.yaml` — clean.
- Diff matches the exact placement used in the Okta/Azure/etc. templates.

Fixes #784

### DIFF
--- a/deployment/infrastructure/bedrock-auth-google.yaml
+++ b/deployment/infrastructure/bedrock-auth-google.yaml
@@ -100,6 +100,7 @@ Resources:
             Action:
               - 'bedrock:InvokeModel'
               - 'bedrock:InvokeModelWithResponseStream'
+              - 'bedrock:CallWithBearerToken'
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:*::foundation-model/*'
               - !Sub 'arn:${AWS::Partition}:bedrock:*:*:inference-profile/*'
@@ -111,6 +112,7 @@ Resources:
             Action:
               - 'bedrock:InvokeModel'
               - 'bedrock:InvokeModelWithResponseStream'
+              - 'bedrock:CallWithBearerToken'
             Resource:
               - !Sub 'arn:${AWS::Partition}:bedrock:::foundation-model/*'
           - Sid: AllowBedrockListRegional


scouturier:
Companion fix: #787 (issue #786) fixes the `inferenceCredentialHelper` ENOENT spawn failure (helper value had inline args; Claude Desktop runs the path with no arguments). For a Google CoWork deployment **both** are required end-to-end — #787 lets the helper spawn, this PR grants `bedrock:CallWithBearerToken` so the minted token isn't rejected with a 403.

wirjo (approved):
✅ Clean fix — adds `bedrock:CallWithBearerToken` to the Google federated role IAM policy, matching all other auth templates (Okta, Azure, Auth0, Cognito, Generic, IDC). +2 lines, single template, backward compatible (additive IAM permission). cfn-lint passes. Ship it.

PR #796, opened 2026-07-20
fix(cowork): sign Bedrock bearer token with empty-payload SHA-256, not UNSIGNED-PAYLOAD

scouturier:
## Why

CoWork 3P's `credential-process --desktop` (#733) mints a Bedrock bearer
token for Claude Desktop's `inferenceCredentialHelper` by hand-rolling a
SigV4 presigned request. The canonical request used the `UNSIGNED-PAYLOAD`
sentinel, which is only valid for services (S3, S3 Express) whose presigned-URL
auth classes explicitly override `payload()` to return it. Bedrock's
`CallWithBearerToken` signer does not use that override — AWS's own reference
generator (`aws-bedrock-token-generator-python`, via botocore's
`SigV4QueryAuth`) signs the SHA-256 hash of the empty POST body instead.

Because a SigV4 signature is a keyed hash over the exact canonical-request
bytes, this mismatch means every bearer token minted via `--desktop` carries
an invalid signature — Bedrock would reject it.

This was flagged by a user via an AI assistant transcript; verified independently
against botocore's `auth.py` source before fixing (see `EMPTY_SHA256_HASH` /
`SigV4Auth.payload()`).

## What

- Replace the `UNSIGNED-PAYLOAD` literal with the well-known SHA-256 digest of
  an empty payload (`e3b0c442...b855`) in the canonical request.
- Add a regression test (`bedrock_token_test.go`) that independently
  recomputes the expected SigV4 signature from the token's own presigned URL
  and fails if the canonical request reverts to `UNSIGNED-PAYLOAD`. Verified
  it fails on the pre-fix code and passes on the fix.

## Scope

Go only, `--desktop` flag path (CoWork 3P / Claude Desktop helper-script mode).
Claude Code CLI is unaffected — it uses the default (non-`--desktop`)
credential-process output, which is AWS SDK–signed, not this hand-rolled path.

## Testing

- `go test ./... -race -count=1` — all packages pass
- `go vet ./...` — clean
- New test verified red (with `UNSIGNED-PAYLOAD`) → green (with the fix)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -307,6 +307,26 @@
         "is_secret": false
       }
     ],
+    "source/go/cmd/credential-process/bedrock_token.go": [
+      {
+        "type": "Hex High Entropy String",
+        "filename": "source/go/cmd/credential-process/bedrock_token.go",
+        "hashed_secret": "244f421f896bdcdd2784dccf4eaf7c8dfd5189b5",
+        "is_verified": false,
+        "line_number": 27,
+        "is_secret": false
+      }
+    ],
+    "source/go/cmd/credential-process/bedrock_token_test.go": [
+      {
+        "type": "Secret Keyword",
+        "filename": "source/go/cmd/credential-process/bedrock_token_test.go",
+        "hashed_secret": "92e2c153a730c3d1eeffce95c4e480ec200edad9",
+        "is_verified": false,
+        "line_number": 19,
+        "is_secret": false
+      }
+    ],
     "source/go/cmd/credential-process/config_fixtures_test.go": [
       {
         "type": "Base64 High Entropy String",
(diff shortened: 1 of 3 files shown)


scouturier:
@wirjo could you review this when you have a chance? This fixes a SigV4 signing bug in `generateBedrockToken()` (the `--desktop` bearer-token path from #733) — the canonical request was hashing the payload as `UNSIGNED-PAYLOAD` instead of the empty-body SHA-256 that Bedrock's signer expects, so tokens minted via this path would fail signature verification.

All CI checks are green. Verified via an independent regression test that recomputes the expected signature against AWS's reference signer (botocore's `SigV4Auth.payload()`). Would appreciate your eyes given you wrote the original `--desktop` implementation and have more context on how it's been validated in practice.

Issue #797, opened 2026-07-24
Go credential-process ignores AWS_SHARED_CREDENTIALS_FILE, and session-mode static block shadows credential_process on natural expiry (auth breaks after ~12h)


scouturier:
## Why (user impact)

A customer running the **Go** credential-process (v2.4.0, `credential_storage: session`, Windows) reports that authentication **breaks once the cached AWS credentials expire** (their STS session lifetime is 12h) whenever `AWS_SHARED_CREDENTIALS_FILE` points somewhere other than the user home, e.g.:

```json
"AWS_SHARED_CREDENTIALS_FILE": "C:\\ProgramData\\.aws\\credentials"
```

Reported behavior:
- **env set + creds valid** → works, binary not invoked (fast).
- **env set + creds expired** → `credential-process` is never re-invoked to refresh; Claude Code cannot complete auth (403).
- **env unset** → auth works, but the binary is invoked on ~every request, adding ~5–7s latency per conversation.

Their stopgap is a wrapper that reads the cached file directly and only invokes the binary near expiry (getting them to 0–2s).

## Root cause — a combination of two real defects

The break requires **both**. Defect 2 is why the SDK serves stale creds instead of refreshing; Defect 1 is why the eventual refresh lands in a file the SDK isn't reading, turning transient staleness into a *permanent* wedge.

### Defect 1 — the binary hardcodes its cache file and ignores `AWS_SHARED_CREDENTIALS_FILE`

```go
// source/go/internal/storage/session.go:103-106
func credentialsFilePath() string {
	home, _ := os.UserHomeDir()
	return filepath.Join(home, ".aws", "credentials")
}
```

This is the single source of truth for all three session-file ops — `ReadFromCredentialsFile` (`:16`), `SaveToCredentialsFile` (`:54`), and `RemoveFromCredentialsFile` (`:114`). There are **no non-test reads** of `AWS_SHARED_CREDENTIALS_FILE` anywhere in `source/go`. Python has the same blind spot at three sites: `source/credential_provider/__main__.py` ~L620 (clear/EXPIRED path), ~L963 (save), ~L1014 (read).

So the binary's write file (`%USERPROFILE%\.aws\credentials`, env-blind) and the SDK's consumer file (`C:\ProgramData\.aws\credentials`, env-aware) are the **same file only when the env var is unset**. This divergence is the crux.

[...]

PR #798, opened 2026-07-24
fix(credential-process): honor AWS_SHARED_CREDENTIALS_FILE in session mode (#797)

scouturier:
## Why
With a relocated AWS shared credentials file (e.g. MDM sets `AWS_SHARED_CREDENTIALS_FILE` to `C:\ProgramData\.aws\credentials`), the session-mode credential-process read/wrote/cleared `~/.aws/credentials` while the AWS SDK resolved credentials from the relocated file. A stale static block in the SDK's file — which outranks `credential_process` in the profile chain — could never be refreshed or cleared, permanently shadowing `credential_process` (403 `InvalidClientTokenId`).

Reported by a customer on v2.4.0 (Windows, EntraID/OIDC, Go credential-process).

## What
- **Go + Python** session storage now resolve `AWS_SHARED_CREDENTIALS_FILE` (raw, matching the Go/JS SDK loaders) with byte-identical fallback to `~/.aws/credentials` when unset.
- **Python clear path** now DELETES the profile section instead of writing an `EXPIRED` static placeholder (parity with the Go binary, #767/#768) so the SDK falls through and re-invokes `credential_process`.
- `ccwb destroy` clear path honors the env var too.

## Compatibility
Additive and env-gated: no existing deployment sets the var, so the resolved path is byte-identical to today's `~/.aws/credentials`.

## Tests (fail without fix)
- Go `session_test.go` (5): env-var honored, unset fallback, custom-path round-trip, home-file-untouched divergence guard, missing-file → nil.
- Python `test_shared_credentials_file_env.py` (8): incl. Go↔Python path parity and clear-deletes-section (no `EXPIRED` placeholder left behind).
- All Go storage + credential-process and Python related suites pass; ruff clean.

### DIFF
--- a/source/claude_code_with_bedrock/cli/utils/helpers.py
+++ b/source/claude_code_with_bedrock/cli/utils/helpers.py
@@ -5,6 +5,7 @@
 
 import configparser
 import logging
+import os
 import platform
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


wirjo (approved):
Reviewed: additive env-gated fix, zero risk to existing deployments, 13 regression tests, clean code. LGTM.

PR #805, opened 2026-07-28
feat(admin-console): IAM Identity Center admin console for config management (#751 part 2/4)

wirjo:
## What this does for users

Gives IDC landing page administrators a **web-based management UI** for controlling Claude Desktop configuration across their organization — no CLI or JSON editing needed.

**User story:** An enterprise admin deploys the IDC landing page, then navigates to `/admin` to configure which models each IAM Identity Center group can access, what tools are enabled, and what command-level permissions apply. Changes propagate automatically to all users via bootstrap polling (~30 min).

## Capabilities

- **Model allowlists per group** — assign different model tiers to different teams
- **Tool policies** — disable or restrict built-in tools (Bash, Editor, etc.)
- **Command permissions** — allow/ask/deny rules for specific commands (e.g. `Bash(git push:*)`)
- **MCP server configuration** — manage managed MCP servers per-group
- **One-click deploy** — push configuration to all groups from the admin UI

## Architecture

- Separate CFN stack (`admin-console.yaml`) — doesn't modify the landing page stack
- Attaches to the same ALB via a `/admin*` listener rule
- Access gated by IDC group membership (admin group only)
- Security: HMAC-SHA256 session cookies, CSRF origin checks, HTML escaping, Secrets Manager-backed signing key
- Config stored in S3 (same bucket as landing page assets)

## Files

- `deployment/infrastructure/admin-console.yaml` — CFN template (498 lines)
- `deployment/infrastructure/lambda-functions/admin_console/index.py` — Lambda handler with embedded admin UI (2404 lines)
- `source/tests/cli/commands/test_admin_console.py` — unit tests

## Lint notes

- cfn-lint: 2 warnings (W1030 Ref pattern, W3002 package hint) — both pre-existing patterns in this repo
- ruff: BLE001 warnings for broad exception catches in Lambda — deliberate resilience pattern for graceful degradation

## Context

Part 2 of decomposed series from PR #751 (@duttaabh's IAM Identity Center Self-Service Portal).

Co-authored-by: duttaabh <duttaabh@users.noreply.github.com>

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -177,6 +177,16 @@
         "is_secret": false
       }
     ],
+    "deployment/infrastructure/admin-console.yaml": [
+      {
+        "type": "Secret Keyword",
+        "filename": "deployment/infrastructure/admin-console.yaml",
+        "hashed_secret": "b63bf00edb07af6ffba7f7ceb7ed573a913271f7",
+        "is_verified": false,
+        "line_number": 331,
+        "is_secret": false
+      }
+    ],
     "deployment/infrastructure/bootstrap-device-code.yaml": [
       {
         "type": "Secret Keyword",
--- a/deployment/infrastructure/admin-console.yaml
+++ b/deployment/infrastructure/admin-console.yaml
@@ -1,0 +1,498 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: >-
+  Optional admin console for the IAM Identity Center landing page. Adds a
+  /admin* listener rule + Lambda target group to the ALB provisioned by
+  landing-page-distribution.yaml (AuthType=idc). Deployed as a separate
+  stack (`ccwb deploy admin-console`) so the landing page itself has no
+  dependency on IDC group/permission-set management — only sites that need
+  the admin UI deploy this stack.
+#
+# Why a separate stack instead of bundling into landing-page-distribution.yaml:
+#   - The landing page (self-service downloads + auth) is useful on its own;
+#     the admin console (group->model->permission-set provisioning) is a much
+#     more powerful, more privileged capability that not every deployment
+#     wants or needs.
+#   - Reuses the SAME ALB/listener/Cognito-IDC-bridge the distribution stack
+#     already created — no new ALB, no new Cognito app client, no new
+#     Secrets Manager secret. This stack only adds a listener rule + target
+#     group + Lambda.
+#   - Auth is 100% ALB-native `authenticate-oidc` (the same mechanism the
+#     landing page's default listener action uses), not a custom Lambda
+#     session-cookie scheme. ALB validates the JWT and passes the decoded
+#     claims via the x-amzn-oidc-data header; this Lambda only adds a second,
+#     independent authorization check (live IAM Identity Center group lookup)
+#     on top of that authentication, gating everything under /admin.
+
+Parameters:
+  IdentityPoolName:
+    Type: String
+    Description: Name prefix for resources (should match the distribution stack's IdentityPoolName)
+
+  CustomDomainName:
+    Type: String
+    Description: >-
+      Custom domain of the distribution stack's ALB (same value passed to
+      landing-page-distribution.yaml's CustomDomainName). Used to build the
+      admin console URL and to validate the Origin header on admin POST
+      requests (CSRF defense-in-depth).
+    AllowedPattern: '.+'
+
+  HTTPSListenerArn:
+    Type: String
+    Description: >-
+      ARN of the distribution stack's ALB HTTPS listener (from that stack's
+      HTTPSListenerArn output). This stack attaches a new listener rule here
+      rather than creating its own ALB.
+    AllowedPattern: '.+'
+    ConstraintDescription: 'HTTPSListenerArn is required — deploy the distribution stack first.'
+
+  ListenerRulePriority:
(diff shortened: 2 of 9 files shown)


wirjo:
### Hardening: XSS prevention in admin console JavaScript

Fixed stored XSS vectors — all user-controlled values in `renderMcpServers()` and `renderMappings()` are now wrapped with `escapeHtml()` before `innerHTML` insertion.

**Affected values:**
- MCP servers: `s.name`, `s.url`, `s.command`, `s.description`, `s.args.join()`
- Model mappings: `m.modelName` in both `<span>` tags and `<option>` elements

Low practical risk (admin-only surface, an attacker would need admin access to write malicious values and another admin would need to view the page), but good defense-in-depth for code review confidence.

wirjo (commented):
## Review: IAM Identity Center admin console for config management (part 2/4)

**Security/regression risk: Low-Medium.** This is a large, self-contained new feature (+4059/-2, 9 files) that adds a privileged admin surface for IDC group→model→permission-set management. It's not yet wired into any existing deploy orchestration or CLI command (only `deployment/infrastructure/lambda-functions/admin_console/`, `shared/`, and its own CFN stack + tests are touched), so it carries no regression risk to existing functionality. Risk is scoped entirely to the new admin console's own auth/authz correctness.

### What I verified

**Auth/authz model** (the part I scrutinized hardest, given this touches IDC):
- Authentication is delegated entirely to the ALB's native `authenticate-oidc` listener action — the Lambda never handles raw credentials or mints sessions in ALB mode. In CloudFront mode, identity comes from an HMAC-signed session cookie shared with the landing page, validated with `hmac.compare_digest` (constant-time) — correct, not a naive `==` comparison.
- Authorization is a **separate, live** IAM Identity Center group-membership lookup on every request (`get_user_idc_groups`), never trusted from a token claim. This matches the pattern described in the module docstring and is the right call for an admin surface — a stale/replayed JWT claim can't grant stale admin access.
- CSRF defense-in-depth via Origin/Referer header checking (`verify_request_origin`) on all state-changing (POST) requests, fails closed if neither header is present. Reasonable given this sits behind SSO, not a public form.
- IAM policy for the Lambda's execution role is tightly scoped where the SSO Admin/Identity Store APIs allow it (Instance-scoped, PermissionSet-scoped by instance ID), with `Resource: *` only where the AWS API genuinely requires it (documented inline with references to the service-authorization reference) — no over-broad wildcards that could've been scoped.
- S3 access is scoped to `admin/*` and `config/*` prefixes only, not full bucket access.
- Internal errors are never leaked to callers (`internal_error_response()` returns a generic message; full detail goes to CloudWatch only via `print()`), and user-controlled strings are newline-stripped before logging (`log_safe()`) to prevent log injection.

**XSS (flagged in a prior local commit, `fcd0356`):** verified the fix is correct — `escapeHtml()` uses the safe DOM `textContent`/`innerHTML` round-trip pattern (not a manual regex blocklist), and `attrArg()` correctly composes `JSON.stringify` + `escapeHtml` for safe injection into `onclick="..."` attribute contexts. However, I found the original fix was incomplete: `renderModels()` (modelName/modelId/type) and `renderPolicies()`'s allowed-folders/egress-hosts lists still built `innerHTML` from unescaped array values. I fixed these in a follow-up commit for consistency. Practical severity remains low either way — this is an admin-only surface (IDC-gated), and the affected values are Bedrock API responses or admin-entered policy config rather than untrusted third-party input — but defense-in-depth is worth completing since the pattern is already established elsewhere in the same file.

**Other things checked, no issues found:**
- `group_name_to_config_key()` (derives S3 key prefixes from IDC group display names) is not path-traversal-hardened (a group named e.g. `claude-code-../../evil` would produce a `../../evil` config key) — but the group names come from a live IDC `list_groups` lookup filtered to names containing "claude", not raw user input, and only IDC admins can create/rename IDC groups. Low risk, but could be worth a stricter sanitizer (strip `/`, `.`) in a follow-up if this becomes reachable from lower-trust input later.
- No secrets logged or leaked; `.secrets.baseline` diff only adds one detected-and-cleared "Secret Keyword" false positive (`!Ref SessionSigningSecretArn`, a parameter reference, not a literal secret).
- `cfn-lint` on `admin-console.yaml`: only 2 pre-existing-pattern warnings (W1030 ARN-pattern mismatch on a Ref, W3002 packaging note), no errors.

### Validation performed locally
- `ruff check .` / `ruff format --check .`: clean.
- `poetry run pytest tests/cli/commands/test_admin_console.py`: 16/16 passed (auth header extraction, admin/non-admin group gating, CSRF origin checks, config-key derivation, MDM/bootstrap S3 key generation).
- `poetry run pytest tests/` (full unit suite, excl. e2e/integration which need live AWS): 1448 passed, 28 skipped, 1 xpassed — no regressions.
- Pushed a follow-up commit completing the XSS-escaping cleanup described above.

### Verdict
No blocking security issues found beyond the (now fully fixed) XSS gaps, which were low-severity given the admin-only, IDC-gated surface. Auth/authz design is sound and appropriately paranoid for an identity-admin surface (live group re-resolution, no trust in token claims, tight IAM scoping). Merge-ready from a security/correctness standpoint — leaving as COMMENT rather than APPROVE per review-agent convention, but I'd be comfortable seeing this merged once a maintainer has eyeballed the diff.

PR #827, opened 2026-09-16
fix(config): strip whitespace from profile fields used in headers and paths

TahiriAbdou:
A trailing newline in the profile name reaches an S3 request header and aborts `ccwb distribute` at the upload step:

```
Ready to distribute for: macos-arm64
⠇ Preparing upload...
Uploading to S3 ━━━━━━━━━━━━━━━━━━━━ 0.0% • 0.0/66.2 MB • -:--:--

An HTTP Client raised an unhandled exception: Invalid header value b'regily-prod-eu-west-1\n'
```

`Profile.from_dict` stores string fields verbatim, so a stray newline or trailing
space in `~/.ccwb/profiles/<name>.json` (hand-edited, or shell-interpolated during
setup) propagates into `botocore` request headers and S3 key paths.

This strips whitespace on load for the fields that end up in HTTP headers or S3
keys: `name`, `provider_domain`, `client_id`, `aws_region`, `identity_pool_name`.

Repro:

```python
from claude_code_with_bedrock.config import Profile

p = Profile.from_dict({
    "name": "regily-prod-eu-west-1\n",
    "aws_region": " eu-west-1\n",
    "identity_pool_name": "claude-code-auth\n",
})
assert p.name == "regily-prod-eu-west-1"
```

Before: values keep their whitespace, and `ccwb distribute` fails as above.
After: values are stripped and the upload proceeds.

Tests: 1779 passed, 28 skipped. The one failure on this branch
(`test_cfn_lint_passes[cognito-user-pool-setup.yaml]`) also fails on a clean
`main` and is unrelated to this change.


kimnamu (commented):
Thanks for tracing this to the exact header and for putting the `from_dict` repro in the description; both made this quick to verify. Stripping at the load boundary is the right layer: `get_profile` and `load_profile` both route through `Profile.from_dict`, so every command picks up the fix at once.

**Verified** on main `014d7ab` vs PR head `2e2dcbc`, with S3 pointed at a closed local port so header validation is the only thing that can fail before the socket opens:

| | main | PR head |
|---|---|---|
| `Profile.from_dict({"name": "regily-prod-eu-west-1\n", ...}).name` | `'regily-prod-eu-west-1\n'` | `'regily-prod-eu-west-1'` |
| `put_object(Metadata={"profile": profile.name})` (distribute.py L1086) | `HTTPClientError: ... Invalid header value b'regily-prod-eu-west-1\n'` | `EndpointConnectionError` (header accepted, socket refused as expected) |
| `pytest tests/` on PR head | | `1674 passed, 29 skipped, 1 xpassed` (integration deselected) |

**One thing to add: a regression test.** The repo's CLAUDE.md asks that every bug fix ship with a test that fails without it. This one drops into `tests/test_config.py::TestProfileModel` next to `test_from_dict_with_cross_region`, and I ran it on both branches:

```python
    def test_from_dict_strips_whitespace_from_header_and_path_fields(self):
        """Stray whitespace in a hand-edited profile JSON must not reach HTTP
        headers (S3 Metadata) or file paths. Fails without the strip in from_dict."""
        profile = Profile.from_dict(
            {
                "name": "test\n",
                "provider_domain": " test.okta.com\n",
                "client_id": "test-client ",
                "credential_storage": "session",
                "aws_region": " us-east-1\n",
                "identity_pool_name": "test-pool\n",
            }
        )

        assert profile.name == "test"
        assert profile.provider_domain == "test.okta.com"
        assert profile.client_id == "test-client"
        assert profile.aws_region == "us-east-1"
        assert profile.identity_pool_name == "test-pool"
```

```
main    (014d7ab): 1 failed    AssertionError: assert 'test\n' == 'test'
PR head (2e2dcbc): 28 passed   (this test + test_config.py + test_questionary_none_choice.py)
```

<details>
<summary>Two related observations, not blocking</summary>

1. `Config._is_valid_profile_name` uses `re.match(r"^[a-zA-Z0-9\-]+$", name)`, and `$` matches before a trailing newline, so the save side accepts the same value this PR strips on load. I confirmed `save_profile` then writes the newline into the filename:

```
_is_valid_profile_name('regily-prod-eu-west-1\n') -> True
save_profile(...) accepted; files: ['regily-prod-eu-west-1\n.json']
re.fullmatch(r"[a-zA-Z0-9\-]+", 'regily-prod-eu-west-1\n') -> False
```

Switching that `re.match(...$)` to `re.fullmatch` would close the door on the save side too. Happy to see it left for a follow-up if you'd rather keep this PR to the load path.

2. Three more string fields end up as `region_name` and are not in the tuple: `codebuild_region` (via `get_codebuild_region`), `sso_region` (packaged as `idc_region`) and `selected_source_region`. botocore rejects the whitespace variants as well:

```
boto3.client("s3", region_name="eu-west-1\n") -> ValueError: Invalid endpoint: https://s3.eu-west-1
boto3.client("s3", region_name=" eu-west-1")  -> InvalidRegionError: Provided region_name ' eu-west-1' doesn't match a supported format.
```

If widening is acceptable, one comprehension in the same spot covers every string field and removes the tuple to maintain. I ran it against the test above plus `test_config.py`, `test_config_migration.py` and `test_questionary_none_choice.py` (35 passed); lists, dicts and bools pass through untouched:

```python
        data = {k: v.strip() if isinstance(v, str) else v for k, v in data.items()}
```

No profile field carries meaningful leading or trailing whitespace as far as I can tell, but that is your call; the explicit tuple is also fine.

</details>

